Issue #182, opened 2012-06-27
node 0.8.0 and toDataURL()


lpinca:
toDataURL() method seems to fail with node 0.8.0
Tried with live-clock example and got this error:

```
TypeError: Cannot call method 'toString' of undefined
    at Canvas.toDataURL (/home/luigi/Desktop/node-canvas/lib/canvas.js:174:37)
    at Server.<anonymous> (/home/luigi/Desktop/node-canvas/examples/live-clock.js:111:29)
    at Server.EventEmitter.emit (events.js:91:17)
    at HTTPParser.parser.onIncoming (http.js:1785:12)
    at HTTPParser.parserOnHeadersComplete [as onHeadersComplete] (http.js:111:23)
    at Socket.socket.ondata (http.js:1682:22)
    at TCP.onread (net.js:404:27)
```

It seems that toBuffer() return undefined.

PR #194, opened 2012-08-08
Fixes an issue related to a wrong variable type (#182)

lpinca:
The issue was generated by passing the negative value of an unsigned variable as parameter of the function `AdjustAmountOfExternalAllocatedMemory`.

The version of V8 included in node v0.8.x changed the param type of the function `AdjustAmountOfExternalAllocatedMemory` from `int` to `intptr_t`.

The result of this change is that if `foo` is an unsigned variable, passing `-foo` to `AdjustAmountOfExternalAllocatedMemory` is the same of passing the value `(UINT_MAX - foo + 1)` and not the expected `- foo`.

Changing the type of the passed variable to `int16_t` fixed the issue.

### DIFF
--- a/src/closure.h
+++ b/src/closure.h
@@ -16,7 +16,7 @@ typedef struct {
   Persistent<Function> pfn;
   Handle<Function> fn;
   unsigned len;
-  unsigned max_len;
+  int16_t max_len;
   uint8_t *data;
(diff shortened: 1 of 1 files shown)


Dman9296:
Nothing but the best

PR #230, opened 2012-11-08
Make sure node-canvas doesn't terminate node on jpeg errors

dfoody:
Previously, when jpeg processing failed, the library would just print an error and terminate the process.  These changes detect the jpeg error and return an error to the caller instead of quitting the process.


TooTallNate:
Do you have an image that will cause jpeg processing to fail, and node quits?

dfoody:
@TooTallNate I've added a test (corruption.test.js) to the suite that sees how canvas responds to corrupt jpeg data.  On my branch the test passes (and process doesn't terminate).  If you run the same test on master branch you'll see the process terminates mid-test.

We have seen images "in the wild" that caused our node processes to terminate - but we didn't capture them at the time, so this test simulates the same case.

When writing this test case, I found there's another issue with corruption (but haven't addressed it): if the header bytes of an image are corrupted then no image can later be loaded in the same process (you have to quit the process and restart).

src/Image.cc
```diff
@@ -754,6 +765,21 @@ Image::decodeJPEGIntoSurface(jpeg_decompress_struct *args) {
...
 /*
+ * Callback to recover from jpeg errors
+ */
+
+METHODDEF(void) canvas_jpeg_error_exit (j_common_ptr cinfo) {
+  // cinfo->err really points to a canvas_jpeg_error_mgr struct, so coerce pointer
+  canvas_jpeg_error_ptr cjerr = (canvas_jpeg_error_ptr) cinfo->err;
+
+  // Always display the message.
+  (*cinfo->err->output_message) (cinfo);
```
LinusU: I would like for this message to be returned in the `Error` object instead of always printed.

test/corruption.test.js
```diff
@@ -0,0 +1,166 @@
...
+"HMsV5FpteJ6F39BJZVgP5F5J+BaZd3Sb13rQKDQBqAXAqzU9EKHTjessbALMlyBtWzCwDy1t7u/m7m3Y" +
+"XH3ne62v1j+zWprL8puX3oVI7/bWYL+IjbWnXuN86f6cgx8XM715nU55CNJp9AGwft0rvcHg7eHhwwXz" +
+"IO9elcdWTq2jEEAQBAEAQBAEAIBFDqQED1D0fiM1bmO7ga/XyO1OaTta4aQrZRUlRlly3GapJVOWZ7wk" +
+"zFm50mLlFzFrEMlGSDsd7DvwrUnhNxGXcteuD6zTb2xzGOdy31pLb0957SG+R3snyFa0rElrRoTsTjrT" +
+"McXhI0aa6BRYuAxG9YOD5Cxaxw/Pk9eY8Tqb5FyGYXPPuV+1aF7cp6BlWB/j2Un45aX3dBnPvOUVceUb" +
+"zoHpWpHD11Ekyu3gv7sj5eB8gOpwFG/G6jfSpPD5Nfuao0XLoNeeKtx1vqNgxfRN1MQ+9fyt/wBOMn0v" +
+"ND5gO1dDhPT9uGm4+J7tnzNC7mEnojoNzx+JtLKJscMYaG6gBQKfjFRVEqIj229LM1XFAgCAIAgCAIAg" +
+"CAIAgPHNa4UIqgMabG2soPMwadaAibjofp6d/PJYwOfWoeYmc1d9aVVsop60U4VWtCkdDYMGvy0fwhYf" +
+"4ln9EepGXzp731mVbdK4i3dzRW8bHfWaxrT5wFmjCMdSoWOTeskYrK3j9lgrvVxQvgAakAQBAEAQBAEA" +
+"QBAEAQBAEAQBAEAQBAEAQBAEAQBAEAQBAEAQBAEAQBAEAQBAEAQBAEAQBAEAQBAf/9k=";
```
LinusU: Please add this file as `test/fixtures/corrupt.jpg` instead of embedding it as base64

test/corruption.test.js
```diff
@@ -0,0 +1,166 @@
...
+
+var imageData = new Buffer(imageBase64, 'base64');
+
+function loadImage(data)
+{
```
LinusU: Keep the braces on the above line please, this occurs multiple times...

test/corruption.test.js
```diff
@@ -0,0 +1,166 @@
...
+    try
+    {
+		var img = new Canvas.Image();
```
LinusU: Bad indentation

test/corruption.test.js
```diff
@@ -0,0 +1,166 @@
...
+		ctx.fillRect(0, 0, img.width, img.height);
+		ctx.drawImage(img, 0, 0, img.width, img.height);
+
+		var rtrn = cv.toBuffer();
+        return true;
+    }
+    catch(ex)
+    {
+        return false;
+    }
```
LinusU: Actually, please remove the `try`/`catch` here and use `assert.throws` when calling this.

test/corruption.test.js
LinusU: Actually, reading again it seems like this file isn't corrupt in of its self. Could you use the already existing `checkers.png` or `clock.png`?

LinusU: Of course you can't, because they are png files :facepalm:

Oh well, could you add the file as `test/fixtures/chrome.jpg` then?

test/corruption.test.js
```diff
@@ -0,0 +1,166 @@
...
+    bad0[0]++;
+    bad1[1]++;
+
+    assert.equal(loadImage(imageData), true);
+
+    assert.equal(loadImage(bad0), false);
+    assert.equal(loadImage(bad1), false);
+
+    // If the original image can be loaded after corrupting header data
+	// on a different image, then this assertion should pass
```
LinusU: bad indentation

LinusU:
#1124 got merged, thanks for the initial push @dfoody 😅

PR #292, opened 2013-06-06
Add support for multiple page size

victorgoya:
Hello !

The patch add two optional parameters `witdh` and `height` to the `.addPage()` method to use a different size for the new page. Both are in pixels.

The patch use the [cairo_pdf_surface_set_size](http://cairographics.org/manual/cairo-PDF-Surfaces.html#cairo-pdf-surface-set-size) method.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -222,6 +222,16 @@ ctx.fillText('Hello World 3', 50, 80);
 ctx.addPage();
 ```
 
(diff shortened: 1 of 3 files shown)


LinusU:
I'm sorry that this has gone by without any comments, I think it would be nice to have 👍 

If you would like to update your pull request against the latest master I would be happy to merge this :)

PR #293, opened 2013-06-14
Add a way to write a PDF directly into a file.

galiminus:
The patch add a way to write the resulting PDF directly into a file without having to use the toBuffer method.

``` js
new Canvas(witdh, height, "pdf:/tmp/output.pdf");
```

The internal buffer will not be filled anymore.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -228,6 +228,18 @@ ctx.fillText('Hello World 3', 50, 80);
 ctx.addPage();
 ```
 
+ The `.end()` method flush the underlying document after the last
+ page.
+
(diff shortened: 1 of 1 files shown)


piranna:
I think it should be better to write a stream to a file. I would close this.

phorque:
God my code is so 2013 x)

LinusU:
To me, this doesn't seem to follow standard Node conventions, I would prefer to only expose the streaming methods.

@chearon @zbjornson thoughts? Feel free to close if you agree

PR #316, opened 2013-08-13
Fixed context.restore when there is no stack of saves

GothAck:
Context.restore on a freshly resized canvas broke drawing to the canvas, it seems cairo was always saving and restoring, even when there was no need. This has fixed issues using vanilla flot 0.8.1 :)

### DIFF
--- a/src/CanvasRenderingContext2d.cc
+++ b/src/CanvasRenderingContext2d.cc
@@ -1,4 +1,3 @@
-
 //
 // CanvasRenderingContext2d.cc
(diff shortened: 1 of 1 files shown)


olivierrr:
Is this still an issue? it seems to be happening on 1.2.1

LinusU:
@olivierrr What exactly is the problem that is occurring, would it be possible to write a test case for this?

LinusU:
Fixed by #810

PR #349, opened 2013-10-27
add deps for static compile

magicode:
I added the deps for static compile, 
It works well on Linux 64 , except jpeglib made ​​trouble.
It's not perfect
No I tried on Windows Or Mac


kangax:
Is this still needed?

magicode:
no

PR #357, opened 2013-11-21
Simple svg support

Tharit:
This patch adds support for creating SVGs, similar to the existing PDF implementation.

Just pass "svg" to the Canvas-Constructors third parameter; calling `toBuffer` will produce a SVG document.


kangax:
This is now implemented in 1.2

PR #360, opened 2013-11-27
Fixed include path to cairo.h for OSX build

wizcorp-olivier:
This was not affecting Linux env because gcc seems nicer there (e.g. /usr/local/include is added by default to gcc include path).
On OSX, build error was:
../src/Canvas.h:19:10: fatal error: 'cairo/cairo.h' file not found
#include &lt;cairo/cairo.h&gt;
Other way to fix this is to remove “cairo/“ from both #include in ./src/Canvas.h and ./src/PNG.h

### DIFF
--- a/binding.gyp
+++ b/binding.gyp
@@ -53,7 +53,7 @@
             '<!@(pkg-config libpng --libs)'
           ],
           'include_dirs': [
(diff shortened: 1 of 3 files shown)


LinusU:
This is now fixed by #574

PR #410, opened 2014-05-06
Load font file in FontConfig when using Pango

balsamiqMarco:
We need the text rendering capabilities of Pango (i.e. kerning), so we built node-canvas with 'with_pango' set to true. Problem is, that we use a custom TrueType font and it seems that node-canvas doesn't load it properly when using Pango.

The proposed change fixes the font loading step by calling FontConfig API.

Sample code:

    var canvas = new Canvas(400,40);
    var Font = Canvas.Font;
    
    var myFont = new Font('MyFont', '../fonts/pip.ttf');
    myFont.addFace('../fonts/pipBold.ttf', 'bold');
    
    // Do not call addFont when using Pango
    //ctx.addFont(myFont);
    
    ctx.font = "30px MyFont";
    ctx.fillText("kerning enabled AVAVAVA", 2, 32);

### DIFF
--- a/src/FontFace.cc
+++ b/src/FontFace.cc
@@ -8,6 +8,8 @@
 
 #include <nan.h>
 
+#include <fontconfig/fontconfig.h>
+
 Persistent<FunctionTemplate> FontFace::constructor;
 
 /*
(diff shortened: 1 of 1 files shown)


LinusU:
Seems to be landed on master already :+1:

Issue #427, opened 2014-05-22
How to proceed with libjpeg support on Windows


vweevers:
As GTK no longer bundles `libjpeg`, jpeg support needs to be fixed. There are multiple solutions to this:
1. Use the *.lib files (and includes) from libjpeg-turbo. 
   - User must download [libjpeg-turbo-1.3.1-vc.exe](http://sourceforge.net/projects/libjpeg-turbo/files/1.3.1/libjpeg-turbo-1.3.1-vc.exe/download) (or the 64bit version), which by defaults installs to `C:\libjpeg-turbo`. See [my fork](https://github.com/vweevers/node-canvas/tree/win-jpeg) for a working example. This is similar to how GTK is included, except that I did a static include, because it's a small thing. _Note_, I forked from 1.1.3 because master [fails to build on windows](https://github.com/LearnBoost/node-canvas/pull/373#discussion_r12950488).
2. Use sources of [original libjpeg](http://www.ijg.org/files/)
   - Because it's a zip file (1 MB), we could automatically download it as part of the build process, unzip with a Python script (no extra deps) and then build it to `jpeg.lib`.
   - Building it (outside of node-canvas and gyp) is simple (`nmake /f makefile.vc setup-v10 && nmake /f makefile.vc nodebug=1 libjpeg.lib`). I think it's feasible to rewrite the makefile stuff as a gyp target.. well, worth a try, at least. Make that target a dependency if `OS=="win"` and everything is awesome.

What's standard practice for dependencies like these? Except ignoring windows. Thoughts?

PR #458, opened 2014-08-19
libjpeg-turbo for windows

vweevers:
This PR adds JPEG support to canvas under Windows. See also #427. Instructions for user are simple: install `libjpeg-turbo` to it's default location.

A batch script tests if the directory exists, so nothing breaks when libjpeg-turbo is not installed.

### DIFF
--- a/binding.gyp
+++ b/binding.gyp
@@ -3,10 +3,20 @@
     ['OS=="win"', {
       'variables': {
         'GTK_Root%': 'C:/GTK', # Set the location of GTK all-in-one bundle
-        'with_jpeg%': 'false',
         'with_gif%': 'false',
         'with_pango%': 'false',
-        'with_freetype%': 'false'
+        'with_freetype%': 'false',
+        'variables': { # Nest jpeg_root to evaluate it before with_jpeg
+          'jpeg_root%': '<!(node ./util/win_jpeg_lookup)'
+        },
+        'jpeg_root%': '<(jpeg_root)', # Take value of nested variable
+        'conditions': [
+          ['jpeg_root==""', {
+            'with_jpeg%': 'false'
+          }, {
+            'with_jpeg%': 'true'
+          }]
+        ]
       }
     }, { # 'OS!="win"'
       'variables': {
@@ -139,7 +149,10 @@
           'conditions': [
             ['OS=="win"', {
               'libraries': [
-                '-l<(GTK_Root)/lib/jpeg.lib'
+                '-l<(jpeg_root)/lib/jpeg-static.lib'
+              ],
+              'include_dirs': [
+                '<(jpeg_root)/include'
               ]
             }, {
               'libraries': [
--- a/util/win_jpeg_lookup.js
+++ b/util/win_jpeg_lookup.js
@@ -1,0 +1,21 @@
+var fs = require('fs')
+var paths = ['C:/libjpeg-turbo']
+
+if (process.arch === 'x64') {
+  paths.unshift('C:/libjpeg-turbo64')
+}
+
+paths.forEach(function(path){
+  if (exists(path)) {
+    process.stdout.write(path)
+    process.exit()
+  }
+})
+
+function exists(path) {
+  try {
+    return fs.lstatSync(path).isDirectory()
+  } catch(e) {
+    return false
+  }
+}


kangax:
@vweevers could you please resubmit?

vweevers:
Rebased the PR.

rvagg:
Soooo .. libjpeg-turbo isn't just a Windows thing, does it make any sense in enabling support for this on non-Windows too or does gtk make that difficult?

This also needs docs before merging IMO because it introduces another awkward yak to shave on Windows to get this running (already painful enough!).

Also @kangax fwiw I'd take @kkoopa's +1 as meaning it's worthy of merging, he should probably even be a collaborator here, moreso than me at least.

vweevers:
@rvagg this isn't about libjpeg*-turbo* per se. Only about JPEG support, which other systems already have. I'd rather see a quick merge and then let turbo support for non-windows be a separate issue.

Couple of thoughts about the [docs](https://github.com/Automattic/node-canvas/wiki/Installation---Windows), which I would be happy to update:

- It's [no longer necessary](https://github.com/Automattic/node-canvas/pull/442) to add `C:\GTK\bin` to `PATH`, those instructions can be removed?
- Most of the other stuff is about node-gyp, isn't it easier to refer to the node-gyp docs for that?

rvagg:
I agree it wouldn't be worth holding this up but if turbo really is that much faster then it night be worth supporting for other platforms too. Perhaps someone could ticket this?

Re docs: I'm of two minds on the node-gyp overlap, on one hand it's easier to have a single source of docs on how to do compiles on Windows so pointing to nodr-gyp makes sense but on the other hand a large number of bug reports on native addons I maintain are exactly because people didn't follow the links and install as they are supposed to. People don't read if you give them too much to read (me included). I might leave that for others to judge, perhaps you have some thoughts @kangax?

kangax:
I'm indifferent on this

vweevers:
I have cleaned up and updated the [docs](https://github.com/Automattic/node-canvas/wiki/Installation---Windows).

zbjornson:
@vweevers This works for me, but I'd add a note to the libjpeg-turbo installation section saying (again) that architectures must match (32/64 bit), and that if you install 64 bit, either `libjpeg_root` in binding.gyp needs to be changed or the installation directory needs to be renamed from `C:/libjpeg-turbo64` to `C:/libjpeg-turbo`.

vweevers:
@zbjornson Good point. I've since learned one can also do `node-gyp rebuild --GTK_Root=C:/CustomGTKLocation --libjpeg_root=..`. This seems cleaner than modifying the `binding.gyp`. Also, I just noticed the download pages for Windows are dead. Stated reason:

> It is expected that people who build installers for GTK+ applications for Windows bundle GTK+ with them.

Hakuna matata, the wiki can directly link to [the binaries](http://ftp.gnome.org/pub/GNOME/binaries/) instead.

I will update the wiki asap, and maybe see if I can add a util script that checks if either C:\libjpeg-turbo or C:\libjpeg-turbo64 exists, based on node's bitness.

vweevers:
Rebased and updated PR to support 64 bit libjpeg-turbo. Can this be merged? It's been a year ;)

LinusU:
Hi @vweevers, sorry for the delay on this, lets make it happen!

Could you please squash the commits into one and rebase on the latest master? I'll promise to review it when it's done.

Am I correct in assuming that the wiki has already been updated with the instructions? It looks like it.

zbjornson:
I would love to see this merged...
@LinusU if @vweevers doesn't have time to squash, would you be up for using `git rebase -i` instead? ([info](https://robots.thoughtbot.com/git-interactive-rebase-squash-amend-rewriting-history))

vweevers:
Thanks for the ping @zbjornson, I forgot about this. I'll rebase in a minute

fnlctrl:
so.... why isn't this pr merged?

chearon:
I think it's scheduled for version 2, doesn't seem like it would break backwards compatibility though

LinusU:
I would be open to merging this now, as long as it doesn't require anything that wasn't previously required

zbjornson:
@LinusU yep, this proceeds quietly without JPEG support if libjpeg is not present.

However, I just learned that this doesn't appear to compile with VS2015 because [reason](https://github.com/libjpeg-turbo/libjpeg-turbo/issues/45). That means if someone has VS2015 and libjpeg installed then they will no longer be able to build.

The fixes provided there:

> If you're using the TurboJPEG API, you can link against turbojpeg.dll instead of turbojpeg-static.lib.

No

> If you're using the libjpeg API, you can link against jpeg62.dll instead of jpeg-static.lib, provided that your application isn't calling jpeg_stdio_src() or jpeg_stdio_dest().

Those APIs are used

> If you need to use jpeg-static.lib or turbojpeg-static.lib, you'll need to either build your application with an older version of Visual C++ or build libjpeg-turbo using Visual C++ 2015.

Boo

Can look at using gnuwin's jpeg instead...

LinusU:
Ugh, that's a shame...

This is probably a really dump question but doesn't GTK include a JPEG library on windows? I have a vague memory but it's probably wrong...

> Can look at using gnuwin's jpeg instead, (edit) or using libjpegturbo for all platforms.

Searching around I think that I'm leaning towards gnuwin's jpeg, as I would prefer to now introduce another dependency on the other platforms...

chearon:
Since I really wanna get an MSYS build going, I think I just got gyp/VS2015 to compile against the MSYS/MinGW version of libjpeg-turbo. You just have to use `pacman -S mingw-w64-x86_64-libjpeg-turbo-1.4.2-2` to get the DLL, but then you have to generate your own `.lib` for VS by doing [all this](https://adrianhenke.wordpress.com/2008/12/05/create-lib-file-from-dll/). The MSYS version of it has its own dependencies too so you have to also copy over those DLLs into the release folder.

You can get all of node-canvas's features (pango, gif, free type, etc) working that way.  You end up with more DLLs in the release folder but it's probably better than the GTK.zip people are using now

[libjpeg-turbo.zip](https://github.com/Automattic/node-canvas/files/404561/libjpeg-turbo.zip)

LinusU:
Ehh, I just merged #841, is this the same thing or is this using another library? Sorry, I'm not using Windows at all these days so I'm not in the loop...

ping @zbjornson and @chearon, some thoughts on how to proceed would be appreciated :)

zbjornson:
Merging #841 was correct, it fixes the issue that came up in this PR :)

LinusU:
Awesome, closing this then 👍

PR #488, opened 2014-11-27
Fix image loading for non-POSIX systems.

mdehaussy:
fix fopen mode for non-POSIX platform : on POSIX platform, fopen open file in binary mode by default. On non-POSIX platform (like WIndows), fopen open file in text mode by default. Using 'rb' mode fix this issue.


TooTallNate:
Thanks!

PR #489, opened 2014-11-29
change install script url for libpng to one that exists

alum:
libpng 1.16.10 moved to 'older-releases'...


LinusU:
Just saw that all of out build on Travis have failed to download libpng :(

@alum Could you please rebase on master and push again so that Travis can run with the correct version?

LinusU:
Hmm, it seems like something weird happen during the merge/rebase? Could you try to fetch the latest upstream and then `rebase -i upstream/master` picking only your original commit.

alum:
I tried getting this script to work on my new installation of OS X 10.11.1 but it failed in numerous places. How about abandoning the whole install script? It's either that or building something much more robust. The current script just keeps disappointing... :disappointed:

LinusU:
I agree, it's not working properly, we are no longer using it with travis and we don't recommend it as a way to install dependencies anymore. I'm in favour of dropping this.

Issue #266, opened 2013-03-21
Font Encoding or Secondary font


rcyrus:
Greetings. Thank you for making node-canvas as awesome as it is! 

I'm using node-canvas to rasterize content that I had previously created in a web browser. 
Almost everything works great, the only exception is non-standard text with custom fontFace. 

'normal' text works fine but when I try something extended I get a rectangle instead of the symbol.
Examples:
Text on my browser.
![browser](https://f.cloud.github.com/assets/276867/286506/63bc9ad6-923e-11e2-8e04-90f30b6c40a7.png)

Same text rendered from node-canvas.
![node-canvas](https://f.cloud.github.com/assets/276867/286514/7f7e5dfe-923e-11e2-928d-6990e41a5428.png)

The font in question is [Quattrocentro](https://www.google.com/webfonts/specimen/Quattrocento).

Quattrocentro does not have that symbol but I was hoping it would fallback to a secondary font to fill in like a browser would. 

Is there a way for node-canvas to do this or am I SOL when it comes to extended characters?

ryuukibeat:
Sorry to bump the old issue, but I have this same issue when mostly render Japanese, arabic, and other Unicode fonts.

node-canvas version: v2.0.0-alpha.1

I use that version because I want custom fonts with `Canvas.registerFont(path_to_font, { family: 'MyFont' })`
But when the font doesn't have the glyphs (Japanese, etc), it just shows square and no fallback.

How do I achieve custom fonts (TTF) and fallback fonts using that version? Or is there a better solution?

chearon:
Are you using Windows? Pango font substitution [doesn't work](https://bugzilla.gnome.org/show_bug.cgi?id=162681) there. But you can specify a comma separated list when you set `ctx.font` and it should use them in order.

ryuukibeat:
My dev OS is windows. So I need to specify multiple fonts that support unicode? For example `ctx.font = "15px MyFont, Arial, Sans";`?

And my end product OS is Ubuntu LTS 16.04, still print square txt. The only attempt in ubuntu I take is only `sudo apt-get install fonts-arphic-uming fonts-ipafont-mincho ttf-wqy-zenhei ttf-unfonts-core fonts-takao fonts-dejavu ttf-mscorefonts-installer`.

chearon:
Closed via #715

PR #490, opened 2014-12-09
re-enable pango, use fontconfig and floats for drawImage()

jakeg:
- Pango seems to work fine for me (what's the issue its supposed to cause with freetype?) and means I can have fallback fonts, better kerning etc.

- Fontconfig enables finding fonts when using Pango (see also https://github.com/Automattic/node-canvas/issues/266#issuecomment-66109167).

- I have a need for floats rather than ints for the drawImage() arguments.

### DIFF
--- a/binding.gyp
+++ b/binding.gyp
@@ -13,7 +13,7 @@
         'with_jpeg%': '<!(./util/has_lib.sh jpeg)',
         'with_gif%': '<!(./util/has_lib.sh gif)',
         # disable pango as it causes issues with freetype.
-        'with_pango%': 'false',
+        'with_pango%': '<!(./util/has_lib.sh pangocairo)',
         'with_freetype%': '<!(./util/has_cairo_freetype.sh)'
       }
     }]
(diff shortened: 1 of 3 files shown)


kangax:
> I have a need for floats rather than ints for the drawImage() arguments.
Looks like we should change this indeed. I see `drawImage` arguments are defined as "double" (http://www.w3.org/TR/2dcontext/#conformance-requirements) which is _"a floating point numeric type that corresponds to the set of finite double-precision 64 bit IEEE 754 floating point numbers"_ (http://www.w3.org/TR/WebIDL/#idl-double)

I don't know about points 1 and 2.

@rvagg @TooTallNate could you folks take a look?

rvagg:
lgtm, would be nice to have someone confirm that it compiles and works properly _without_ pango with this one applied; @kangax if you don't get anyone doing that just merge and release and see who complains! That ultimately may be the best way to make this project move forward, i.e. upset a few people and make a bit of noise.

kangax:
@rvagg yep, that's one way to do it :)

PR #493, opened 2014-12-15
Usee double instead of Integer in drawImage

Jake0oo0:
drawImage currently uses an integer, rather than a double. I believe that this is causing the method to be less accurate when run, but correct me if I'm wrong.


TooTallNate:
Do you have like a before/after comparison image?

kangax:
Oh, this is already fixed in a recently pulled PR — https://github.com/Automattic/node-canvas/pull/490

Issue #403, opened 2014-04-15
ImageData.data should return a Uint8ClampedArray instead of a CanvasPixelArray


zbjornson:
`CanvasPixelArray` is deprecated per https://www.w3.org/Bugs/Public/show_bug.cgi?id=12586.

I don't care, except `CanvasPixelArray` doesn't have a `set` method that typed arrays have.

Also encountered here by another user: http://stackoverflow.com/questions/21435027/node-js-canvas-implementation-with-uint8clampedarray

(Seeing if I understand this library well enough to submit a pull request...)

LinusU:
Fixed by #604

PR #498, opened 2015-01-06
ImageData.data return type Uint8ClampedArray

allenjared:
ImageData.data return type Uint8ClampedArray instead of CanvasPixelArray

related to
https://github.com/Automattic/node-canvas/issues/403


LinusU:
Fixed by #604

PR #501, opened 2015-01-16
hsl() and hsla() color parsing

### LINKED ISSUE #437: hsl/hsla support
I'm trying to use `hsl/hsla` in both `fillStyle` and `strokeStyle` but [it seems](https://github.com/LearnBoost/node-canvas/blob/master/test/canvas.test.js) not supported yet.

I think it is related with #28 but I didn't find the original commits by the author (his fork doesn't exist anymore).

Any plans on that?

motiz88:
Fixes #437, see also #28.
Also adds support for [rebeccapurple](http://dev.w3.org/csswg/css-color-4/#valdef-color-rebeccapurple).

### DIFF
--- a/src/color.cc
+++ b/src/color.cc
@@ -11,30 +11,18 @@
 #include <limits>
 
 /*
- * Consume whitespace.
- */
-
-#define WHITESPACE \
-  while (' ' == *str) ++str;
-
-#define WHITESPACE_OR_COMMA \
-  while (' ' == *str || ',' == *str) ++str;
-
-
-/*
  * Parse integer value
  */
 
 template <typename parsed_t>
-static bool parseInteger(const char** pStr, parsed_t *pParsed)
-{
+static bool
+parse_integer(const char** pStr, parsed_t *pParsed) {
   parsed_t& c = *pParsed;
   const char*& str = *pStr;
   int8_t sign=1;
 
   c = 0;
-  if (*str == '-')
-  {
+  if (*str == '-') {
     sign=-1;
     ++str;
   }
@@ -57,15 +45,15 @@ static bool parseInteger(const char** pStr, parsed_t *pParsed)
 
 /*
  * Parse CSS <number> value
+ * Adapted from http://crackprogramming.blogspot.co.il/2012/10/implement-atof.html
  */
 
 template <typename parsed_t>
-static bool parseNumber(const char** pStr, parsed_t *pParsed)
-{
+static bool
+parse_css_number(const char** pStr, parsed_t *pParsed) {
    parsed_t &parsed = *pParsed;
    const char*& str = *pStr;
    const char* startStr = str;
-
    if (!str || !*str)
        return false; 
    parsed_t integerPart = 0;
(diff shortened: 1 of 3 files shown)


kangax:
To my C++-uneducated this LGTM (code, tests, etc.) 

/cc @rvagg @TooTallNate

src/color.cc
```diff
@@ -15,23 +17,249 @@
...
+template <typename parsed_t>
+static bool parseInteger(const char** pStr, parsed_t *pParsed)
+{
+  parsed_t& c = *pParsed;
+  const char*& str = *pStr;
+  int8_t sign=1;
+
+  c = 0;
+  if (*str == '-')
+  {
```
rvagg: lots of inconsistencies here with brace style, some of them are with a newline and some are not; have a look through the rest of the project to find a happy place with consistency

src/color.cc
```diff
@@ -15,23 +17,249 @@
...
+#define WHITESPACE_OR_COMMA \
+  while (' ' == *str || ',' == *str) ++str;
+
+
+/*
+ * Parse integer value
+ */
+
+template <typename parsed_t>
+static bool parseInteger(const char** pStr, parsed_t *pParsed)
```
rvagg: why are we manually doing this integer parsing? I'm looking for a good reason in here but it's escaping me so far.

src/color.cc
```diff
@@ -276,6 +505,49 @@ rgba_from_rgba(uint8_t r, uint8_t g, uint8_t b, uint8_t a) {
 }
 
 /*
+ * Helper function used in rgba_from_hsla().
+ */
+
+static float hue_to_rgb(float t1, float t2, float hue) {
+  if(hue < 0) hue += 6;
```
rvagg: style: the paren backing up against the `if` here is inconsistent, newlines after the `if` may also be required, look at the rest of  the project for prevailing code style (note: I can't tell you off the top of my head what the style should be, I'm just noticing inconsistencies in _this_ contribution and would like it to be made internally consistent and consistent with the rest of the existing code)

src/color.cc
```diff
@@ -276,6 +505,49 @@ rgba_from_rgba(uint8_t r, uint8_t g, uint8_t b, uint8_t a) {
...
+
+/*
+ * Return rgba from (h,s,l,a).
+ */
+
+static inline int32_t
+rgba_from_hsla(float h_deg, float s, float l, float a) {
+  float h = (6 * h_deg) / 360.0f;
+  float m1,m2;
+  if (l<=0.5)
```
rvagg: whitespace needed on the next 5 lines, don't bunch things up so much like a classic C programmer, we're JavaScripters here!

src/color.cc
rvagg: ok .. I think I get it .. we're doing _lots_ of parsing and need this to be just a step in that, ignore my comment then

rvagg:
This is fine by me minus all the code style inconsistencies

As an explanation for that I'm guessing that there might be some copy/paste going on from another project somewhere and perhaps we should include attribution. @motiz88 is this all your own code or can we attribute (simply in a comment) parts of it elsewhere?

motiz88:
Sorry about code style, I was honestly not paying enough attention to that. I think I did a better job now :)
Yes, there are small bits I adapted from elsewhere - see the updated comments.

rvagg:
if it works then :+1: from me!

Issue #467, opened 2014-09-06
ShadowBlur on text


Mathyn:
Is ShadowBlur on text supported? I tried the following:

context.shadowColor = "red";
context.shadowOffsetX = 10;
context.shadowOffsetY = 10;
context.shadowBlur = 25;

context.fillText("Hello World", 100, 100);

And all I saw was the shadow placed at an offset of 10, 10 but with no blur effect applied. Can anyone verify if it should or should not work? On a related note: is there a list of all supported and unsupported canvas features working with node-canvas somewhere?

woodcoder:
This should work fine when using image-based output.  It won't work with vector based output currently (e.g. PDF).  The pull request referenced above adds support for that.

zbjornson:
Fixed in #505 🎉

Issue #133, opened 2011-12-02
shadowBlur for Image not working


bteitelb:
Testing node-canvas with this test script https://gist.github.com/1423054 on a machine with Cairo 1.10.0 installed, I am seeing a Gaussian blur shadow for the squares, but not for the arrangement of leaves.  Specifically, I am seeing this:

 ![shade_test](http://ben.teitelbaum.us/tmp/shade_test.png)

Whereas in Chrome, both render with shadowBlur:

 ![shade_test](http://ben.teitelbaum.us/tmp/shade_test_chrome.png)

inssein:
Did anybody find a solution? @woodcoder mentioned what it would take to fix this, just wondering if there has been any movement.

PromInc:
I haven't pursued this any further my self, but no I haven't heard of any progress.

I had toyed with the idea of a kind of 'ugly' workaround of adding a shape layer directly beneath the image in the same size, thus allowing the shadow to be on the 'hidden' shape and creating the illusion that it's on the image.

PR #505, opened 2015-01-24
Shadow blur

woodcoder:
This pull request
   1. adds a test showing how the shadowOffset{X,Y} values shouldn't be affected by the current transform
   2. fixes the shadowOffset{X,Y} translate to not be affected by the current transform matrix
   3. adds a 'PDF' button to the test-server output for easier comparison with PDF output
   4. creates an image buffer for blurred shadows so that they work with the vector-based PDF output
   5. adjusts the blur radius to more closely match the browser's implementation
   6. adjusts the shadowOffset slightly to more closely match the browser's output

It introduces minor regressions in the following tests:
   * *shadow globalAlpha* -- this shadow is not as definite as before (possibly because of multiple applications of the globalAlpha?)
   * *shadow strokeText()* -- this shadow is bolder than before (possibly due to the resolution of the shadow image buffer not matching the destination image?)

However, the other shadow tests all look much better (especially *shadowBlur values*) when compared against Firefox.


kangax:
Thanks @woodcoder, this looks nice. It will probably fix https://github.com/Automattic/node-canvas/issues/467 Any chance to also take care of https://github.com/Automattic/node-canvas/issues/133 with this PR?

woodcoder:
Just had a really quick look at #133 and written a test for it.  Fixing it looks like we'd need to do something like fix up `DrawImage` to do the following if it has a shadow:
   1. get the image as alpha-channel only (using a CAIRO_CONTENT_ALPHA surface?)
   2. render that to a temporary image surface (as in this PR)
   3. pass that to blur (if blurring) and
   4. then paint it.

woodcoder:
This provides a non-blurred image shadow (but respecting any alpha in the image) by using the image as a mask for the shadow colour.  Hopefully that'll get you closer!

kangax:
@woodcoder nice, thanks!

/cc @rvagg @TooTallNate

rvagg:
lgtm I think, my only question would be whether this is a feature addition (minor version bump) or is there stuff in here that would constitute a breaking change (major version bump)?

kangax:
I would consider this is a bug fix. The fact that shadow didn't work for images was not documented and was just a limitation of previous implementation.

woodcoder:
This last commit fixes the two regressions I noticed in the _shadow globalAlpha_ and _shadow strokeText()_ tests -- it was because the line (stroke) width was not being respected when drawing the shadow.  There may be more drawing settings like this that need to be copied between the contexts, but I'm not familiar enough with cairo to know if there's a quick way of doing this.

kangax:
I suppose I'm gonna pull this in.

woodcoder:
Thank you!

One thing I've noticed since, is that I use `cairo_surface_create_similar_image` which was introduced in cairo 1.12.  This may cause issues on Windows installs because the version of cairo in GTK+ is 1.10.

I'm not sure if that'll be a problem going forward, but it's possible to use `cairo_image_surface_create` instead, as shown here:
https://github.com/woodcoder/node-canvas/commit/c2de36ac986e03fa7f8810bda487cae42d50f39c

PR #507, opened 2015-01-29
io.js compatibility

salzhrani:
Tested on node stable and iojs, seems to work fine

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "test": "make test"
   },
   "dependencies": {
-    "nan": "~1.2.0"
+    "nan": "~1.5.1"
   },
   "devDependencies": {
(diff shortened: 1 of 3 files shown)


salzhrani:
0.6 is failing but newer versions are green

salzhrani:
@kangax is dropping support for 0.6 an option? It seems many libraries upgrading nan are dropping support for 0.6. If not could an iojs branch possible ?

TooTallNate:
Thanks!

salzhrani:
Would you like another PR with Travis update and a version bump ?

TooTallNate:
v1.2.0 published to npm

salzhrani:
Wow. Thanks.

PR #511, opened 2015-02-02
Use non-cairo 1.12 API for shadow blur

### LINKED ISSUE #509: 1.2.0 Build Fails
I attempted to update our project's dependencies, along with node-canvas to 1.2.0, but the build failed locally on Windows, and thinking that it may have been because of Windows, I pushed it to travis, and it continued to fail there. I would pastebin the entire Travis log, but it doesn't keep new lines, so this is the best I can do, with the compile error:

http://pastebin.com/wzE5eSK9

Thanks!

woodcoder:
Just use cairo_image_surface_create for shadow blur surface, rather than cairo_surface_create_similar_image used in #505; this call was introduced in cairo 1.12.  Unfortunately the cairo distributed with GTK+, used in the Windows build, is only 1.10.

Using the similar_image version offers backends the chance to be more efficient, but the alternative used here is used as a fallback in the cairo code anyway.

This should fix #509.

### DIFF
--- a/src/CanvasRenderingContext2d.cc
+++ b/src/CanvasRenderingContext2d.cc
@@ -372,7 +372,7 @@ Context2d::shadow(void (fn)(cairo_t *cr)) {
     cairo_user_to_device_distance(_context, &dx, &dy);
     int pad = state->shadowBlur * 2;
     cairo_surface_t *surface = cairo_get_group_target(_context);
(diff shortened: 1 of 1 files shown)


ekynoxe:
Hi guys, what's needed to merge this one in?
Cheers

kangax:
Published 1.2.1

Issue #512, opened 2015-02-04
drawImage bug in 1.2.0


bmustata:
Currently in version 1.2.0 (Linux) there is a drawImage bug where the image is not draw with it's full size when using the ctx.drawImage() command. This command is working correctly in version 1.1.6.

```javascript
var Canvas = require('canvas')
, Image = Canvas.Image
, canvas = new Canvas(640, 471)
, ctx = canvas.getContext('2d')
, fs = require('fs');

fs.readFile(__dirname + '/images/capture.png', function(err, squid){
    if (err) throw err;
    img = new Image;
    img.src = squid;
    ctx.drawImage(img, 243, 95, 153, 273);

    fs.writeFile('output/out.png', canvas.toBuffer());
});
```

woodcoder:
Is capture.png a different size to the width/height you're providing to drawImage?
And if so, is it getting clipped incorrectly?

mattbasta:
I can confirm this issue.

@woodcoder: in the case of the images that I've seen the issue with, it happens for all calls to `drawImage` where the destination height and width are not the same as the source height and width. See the following examples:

Good:
![good](https://cloud.githubusercontent.com/assets/279498/6257619/8ee182c6-b776-11e4-9c4a-5d9c23ae504f.png)

Bad:
![bad](https://cloud.githubusercontent.com/assets/279498/6257621/926f63f4-b776-11e4-8d07-6c0b5efb09d1.png)

The original images being drawn are 32x32px. The destination is 16x16px. What's interesting in this case is that the images are being clipped at the ratio of the source to the destination, but on the destination size. That is, it's clipping at 8x8 at 16x16 (2x), when the image is 16x16 scaled from 32x32.

woodcoder:
Are you able to try pull request #513 (or #518 which is the same fix but has additional tests)?  I think that should resolve it.

mattbasta:
I haven't, but I can when I have a chance.

mahnunchik:
Hi, any news?

kangax:
The (supposed) fix was already merged, so I'm just keeping it open until we release it. I just don't have time to do a release now.

selaux:
Although the fix was merged, I still experience the issue on current master.

bennlich:
I was having a similar problem with the current release when using drawImage with source width and source height set to values greater than the destination width and destination height. I no longer see the clipping when using the master branch--it seems fixed.

cambrozie:
[...]

PR #513, opened 2015-02-04
Fix drawImage clipping bug

woodcoder:
Fix regression introduced by reordering of image shadow code in 39766d1 -- the clip width and height wasn't being adjusted for the current scale.  Affected images that were not being drawn at 1:1 scale.

Candidate fix for #512.


TooTallNate:
LGTM, I think.

We should probably start getting in the habit of adding test cases for regressions like this.

PR #515, opened 2015-02-05
Accept and ignore toDataURL encoderOptions argument

robertjustjones:
Added the encoderOptions argument in toDataURL per [spec!](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement.toDataURL) so that node-canvas can be used with other canvas libraries.

The argument is ignored since it is not used with image/png.  Tests included.

Thanks, Team!

### DIFF
--- a/lib/canvas.js
+++ b/lib/canvas.js
@@ -198,12 +198,13 @@ Canvas.prototype.createSyncJPEGStream = function(options){
  * @api public
  */
 
-Canvas.prototype.toDataURL = function(type, fn){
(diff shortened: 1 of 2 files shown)


robertjustjones:
This PR goes along with the PR https://github.com/tmpvar/jsdom/pull/1025 that fixes the jsdom integration of node-canvas.

LinusU:
Should be fixed by #696 and released as `1.3.6`.

Sorry for the lack of feedback earlier.

PR #517, opened 2015-02-06
Add toDataURL tests inspired by a failure seen with jsdom.

robertjustjones:
I'm getting the following error when using  with jsdom and it's integration with node-canvas for a `<canvas></canvas>` parse. The canvas init looks like `new Canvas(0,0)`. 
```
Error: the surface type is not appropriate for the operation
```

I've written two simple tests here that duplicate the behavior. I believe jsdom intends to be using the behavior of the first test where width/height are set later (that test passes). However, it's inadvertently mimicking the behavior of the second test, which fails. Is that expected? or should the error be caught?

Thanks! Awesome project.

### DIFF
--- a/test/canvas.test.js
+++ b/test/canvas.test.js
@@ -407,6 +407,24 @@ module.exports = {
     });
   },
 
+  'test Canvas#toDataURL() {width,height}=': function(){
+    var canvas = new Canvas;
+    canvas.width = 50;
+    canvas.height = 50;
+    assert.equal(50, canvas.width);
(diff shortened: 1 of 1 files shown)


robertjustjones:
I've submitted a PR https://github.com/tmpvar/jsdom/pull/1025 to address the error above, but I'm still curious whether the failed test in the PR is "intended behavior".

LinusU:
Should be fixed by #696 and released as `1.3.6`.

Sorry for the lack of feedback earlier.

PR #518, opened 2015-02-06
Fix for 'drawImage' scaling

joefreeman:
The scaling of 'drawImage' wasn't working correctly. I *think* I've fixed this. The exaggerated failing test should highlight the problem (the test results were inconsistent before, but only subtly).


joefreeman:
D'oh - just noticed this is the same fix as #513 :(

PR #539, opened 2015-03-19
fixed bug in StreamPNGSync

smremde:


kangax:
Could you please add a test case as well?

smremde:
Not really sure how to add a test case - would have to compare binary outputs.

It's just a typo (and it had been fixed/not made in other places in the code):
https://github.com/Automattic/node-canvas/blob/master/src/Canvas.cc#L283

kangax:
Well, I thought you probably ran into it when something didn't work. Hence the test case. In this case it probably doesn't matter much; we're just trying to add as many tests as possibly since recently.

Issue #525, opened 2015-02-19
Can't get the font example to work on Ubuntu


joefreeman:
The font example (`examples/font.js`) isn't working for me as expected when I run it on an AWS Ubuntu EC2 instance: it's just rendering the same font four times (see below). When I run this locally on OS X, I get four different fonts (as expected).

I think my install steps might just be missing something, but I don't seem to be getting any errors back, and I've tried debugging (via `console.log` and `fprintf`) without being able to find anything that looks wrong.

Here are my reproduction steps:

- Launch EC2 instance with the "Ubuntu Server 14.04 LTS (HVM), SSD Volume Type" image (`ami-29ebb519`).
- Install the node-canvas dependencies:

        sudo apt-get update
        sudo apt-get install libcairo2-dev libjpeg8-dev libpango1.0-dev libgif-dev build-essential g++

- Get the latest node-canvas from GitHub, and install:

        sudo apt-get install git nodejs-legacy npm
        git clone https://github.com/Automattic/node-canvas.git
        cd node-canvas
        npm install

- Run the example:

        cd examples
        node font.js

Here's the output I'm getting on the Ubuntu instance:

![font.png](http://i.imgur.com/0Dcm2vU.png)

Tharit:
I just encountered the same problem when upgrading from a previous version. Seems like Custom Font support is currently broken.

jakeg:
Custom font support is very much working but you need to remove a line if you're using node-canvas with pango support (which you should be). Remove or comment out this line from font.js:
ctx.addFont(pfennigFont);
... and it should work :)

Tharit:
I was just about to post that I figured that out after spending some time debugging the code.
Wish I would have seen your reply sooner - works perfectly indeed, thank you :-)

joefreeman:
I tried moving this `addFont` line, but now I get this:

![font2](https://cloud.githubusercontent.com/assets/535128/6967320/2e687d1e-d97c-11e4-947c-8c08c8ee31b8.png)

I think this isn't using the Pfennig font, so I'm not sure if this is a step forwards of backwards? Do I need to put the font somewhere special for pango to pick it up?

gtlbupt:
I find a solution to solve this problem:  edit the binding.gyp file and turn off the with_pango flags. I  test it on ubuntu 14.04.2 LTS and it work. I hope this may help.

joefreeman:
[...]

PR #543, opened 2015-04-04
Unify the API for working with fonts when Pango is enabled.

joefreeman:
Relating to the discussion of issue #525, I've tried to make sure that the API works the same whether or not Pango is enabled. Previously, calling `addFont` with Pango enabled would cause the font rendering to break. Also, the font had to be referenced by family name. With this change, the `addFont` call is optional, but if it is called, the font can be referenced using the name provided to the `Font` constructor.

So, after setting up the font like this:

    var font = new Font('myFontAlias', '/path/to/MyFont.ttf');
    context.addFont(font);

This:

    context.font = '50px myFontAlias';

Is equivalent to this (assuming the font's family name is 'My Font'):

    context.font = '50px \'My Font\'';

But I'm not familiar with C++, Pango or Cairo, so my changes might be no good.

### DIFF
--- a/examples/font.js
+++ b/examples/font.js
@@ -23,7 +23,7 @@ pfennigFont.addFace(fontFile('PfennigBold.ttf'),   'bold');
 pfennigFont.addFace(fontFile('PfennigItalic.ttf'), 'normal', 'italic');
 pfennigFont.addFace(fontFile('PfennigBoldItalic.ttf'), 'bold', 'italic');
 
-var canvas = new Canvas(320, 320)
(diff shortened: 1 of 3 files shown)


LinusU:
Hi @joefreeman, I'm sorry that this has fallen thru the cracks. Would you mind rebasing on master? I will make sure to review this. Thanks :+1:

joefreeman:
No need - looks like this is working in master now. I'll close this and my issue...

LinusU:
@joefreeman Are you sure? Did you test it with built pango support?

PR #569, opened 2015-06-23
expose ImageData constructor

targos:
I need the ImageData constructor and I currently have to get it `with require('canvas/lib/bindings').ImageData`

Example use: convert a pixel array to dataURL
```js
function toDataURL(width, height, pixels) {
    var imgData = new ImageData(pixels, width, height);
    var canvas = new Canvas(width, height);
    ctx.putImageData(imgData, 0, 0);
    return canvas.toDataURL();
}
```

### DIFF
--- a/src/ImageData.cc
+++ b/src/ImageData.cc
@@ -44,45 +44,65 @@ NAN_METHOD(ImageData::New) {
 
   int width;
   int height;
+  int length;
 
   if (info[0]->IsUint32() && info[1]->IsUint32()) {
     width = info[0]->Uint32Value();
+    if (width == 0) {
+      Nan::ThrowRangeError("The source width is zero.");
+      return;
+    }
     height = info[1]->Uint32Value();
-    int size = width * height;
+    if (height == 0) {
+      Nan::ThrowRangeError("The source height is zero.");
+      return;
+    }
+    length = width * height * 4;
 
 #if NODE_MAJOR_VERSION == 0 && NODE_MINOR_VERSION <= 10
-    Local<Int32> sizeHandle = Nan::New(size);
+    Local<Int32> sizeHandle = Nan::New(length);
     Local<Value> caargv[] = { sizeHandle };
     clampedArray = global->Get(Nan::New("Uint8ClampedArray").ToLocalChecked()).As<Function>()->NewInstance(1, caargv);
 #else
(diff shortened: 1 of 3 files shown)


kangax:
+1 from me, but we should add it to documentation too then.

targos:
I will investigate further because it only works with the internal CanvasPixelArray objects.
`new ImageData(new Uint8ClampedArray(4), 1, 1)` fails

eqyiel:
I also need the ImageData constructor.

The line @targos posted not fail since #604 (now that ImageData is using Uint8ClampedArray).  Can we get this merged?

LinusU:
@targos Could you rebase on master? Also, adding documentation and tests will greatly increase the chances of this getting merged.

targos:
I just found that the constructor does not behave correctly. For example, `new ImageData(2, 2)` creates an array with 4 elements when it should be 16 (width * height * 4)
I will try to fix that and add tests.

zbjornson:
@Targos my mistake (never tested those since they weren't exposed). Let me know if you want help or want me to fix it. I'm going to change that area soon for #637.

targos:
I pushed an update to ImageData.cc + tests. PTAL
For the error messages, I copied the output from Chrome.

zbjornson:
Looks good!

src/ImageData.cc
```diff
@@ -44,46 +44,64 @@ NAN_METHOD(ImageData::New) {
...
-#if NODE_MAJOR_VERSION == 0 && NODE_MINOR_VERSION <= 10
-      height = clampedArray->GetIndexedPropertiesExternalArrayDataLength() / width;
-#else
-      height = clampedArray->Length() / width;
-#endif
+    int size = length / 4;
+    if (width == 0) {
+      Nan::ThrowRangeError("The source width is zero.");
+      return;
+    } else if (size % width != 0) {
```
LinusU: You don't need `else` here, please change to

```c++
}
if (size % width != 0) {
```

LinusU:
One small nit and if you could please rebase on master again (sorry), we have updated the travis config and it will ensure that your tests runs with our new config. Thank you!

targos:
@LinusU done :)

LinusU:
Cool! Merging now :+1:

LinusU:
Released as 1.3.0 \o/

targos:
Thank you!

Issue #572, opened 2015-07-01
npm install canvas failed!


ojlovecd:
Hi, all! I am going to install the `canvas` module for node.js on my OS X. I followed [this post](https://github.com/Automattic/node-canvas/wiki/Installation---OSX)  . Every things were fine until the `Installing node-canvas` step. I got errors below:

```
> canvas@1.2.3 install /Users/Jo/node_modules/canvas
> node-gyp rebuild

  SOLINK_MODULE(target) Release/canvas-postbuild.node
  SOLINK_MODULE(target) Release/canvas-postbuild.node: Finished
  CXX(target) Release/obj.target/canvas/src/Canvas.o
In file included from ../src/Canvas.cc:7:
../src/Canvas.h:19:10: fatal error: 'cairo/cairo.h' file not found
#include <cairo/cairo.h>
         ^
1 error generated.
make: *** [Release/obj.target/canvas/src/Canvas.o] Error 1
gyp ERR! build error 
gyp ERR! stack Error: `make` failed with exit code: 2
gyp ERR! stack     at ChildProcess.onExit (/usr/local/lib/node_modules/npm/node_modules/node-gyp/lib/build.js:269:23)
gyp ERR! stack     at ChildProcess.emit (events.js:110:17)
gyp ERR! stack     at Process.ChildProcess._handle.onexit (child_process.js:1074:12)
gyp ERR! System Darwin 14.3.0
gyp ERR! command "node" "/usr/local/lib/node_modules/npm/node_modules/node-gyp/bin/node-gyp.js" "rebuild"
gyp ERR! cwd /Users/Jo/node_modules/canvas
gyp ERR! node -v v0.12.4
gyp ERR! node-gyp -v v1.0.3
gyp ERR! not ok 
npm ERR! Darwin 14.3.0
npm ERR! argv "node" "/usr/local/bin/npm" "install" "canvas"
npm ERR! node v0.12.4
npm ERR! npm  v2.10.1
npm ERR! code ELIFECYCLE

npm ERR! canvas@1.2.3 install: `node-gyp rebuild`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the canvas@1.2.3 install script 'node-gyp rebuild'.
npm ERR! This is most likely a problem with the canvas package,
npm ERR! not with npm itself.
npm ERR! Tell the author that this fails on your system:
npm ERR!     node-gyp rebuild
npm ERR! You can get their info via:
npm ERR!     npm owner ls canvas
[...]

Issue #514, opened 2015-02-05
doesn't build on OS X with cairo installed with homebrew without command line clang


sma:
We noticed a problem which might be a bug. When trying to install 1.2.0 on OS X it worked on one machine but not on another one. Very frustrating. After 2h of digging, I noticed a subtile difference: I had installed the Xcode development tools via `xcode-select --install`, my colleague never did that but used the tools that came with Xcode. What was different?

We both installed Cairo 1.14.0 via `brew install cairo`. This did copy the library's include files to `/usr/local/include/cairo` as well as `/usr/local/Cellar/cairo/1.14.0/include/cairo/`. A call to `npm install node-canvas` does a lot of magic to build the native library. Frankly, I don't understand that. However, the configuration process adds a lot of include paths for the compiler, including the second one I mentioned above. Then, the source code tries to load .h files using `#include <cairo/cairo.h>`. This cannot work, because this would try to open `/usr/local/Cellar/cairo/1.14.0/include/cairo/cairo/cairo.h` (notice the double `cairo`) I think. On my system, the compiler also added `/usr/local/include` by default so I got the file in question from `/usr/local/include/cairo/cairo.h`. On the other machine, that default include was missing. Probably because that clang wasn't installed as command line tools.

When I removed the useless explicit cairo include to just relay on the default path, I noticed that one .cc file (which I forget, sorry), used `include <name.h>` instead of `include <cairo/name.>` which I'd consider an error that never got noticed because of the second search path added.

Unfortunately, I don't know how to work with gyp or whatever but perhaps this report is still useful for other who have the same strange problem.

amchang:
https://github.com/Homebrew/homebrew/issues/14123,

This patch will fix it.

LinusU:
Should be fixed by #574

Issue #217, opened 2012-10-03
Can't Install.


reggi:
I am having a hard time installing canvas.

	thomas@workstation:~/Desktop/holstground$ sudo npm install canvas
	Password:
	npm info it worked if it ends with ok
	npm info using npm@1.1.61
	npm info using node@v0.8.6
	npm WARN invalid config registry=""
	npm WARN invalid config Must be a full url with 'http://'
	npm info trying registry request attempt 1 at 02:19:29
	npm http GET https://registry.npmjs.org/canvas
	npm http 304 https://registry.npmjs.org/canvas
	npm info install canvas@0.13.1 into /Users/thomas/Desktop/holstground
	npm info installOne canvas@0.13.1
	npm info /Users/thomas/Desktop/holstground/node_modules/canvas unbuild
	npm info preinstall canvas@0.13.1
	npm info build /Users/thomas/Desktop/holstground/node_modules/canvas
	npm info linkStuff canvas@0.13.1
	npm info install canvas@0.13.1
	
	> canvas@0.13.1 install /Users/thomas/Desktop/holstground/node_modules/canvas
	> node-gyp rebuild
	
	gyp info it worked if it ends with ok
	gyp info using node-gyp@0.6.8
	gyp info using node@0.8.6 | darwin | x64
	gyp info spawn python
	gyp info spawn args [ '/Users/thomas/.node-gyp/0.8.6/tools/gyp/gyp',
	gyp info spawn args   'binding.gyp',
	gyp info spawn args   '-f',
	gyp info spawn args   'make',
	gyp info spawn args   '-I',
	gyp info spawn args   '/Users/thomas/Desktop/holstground/node_modules/canvas/build/config.gypi',
	gyp info spawn args   '-I',
	gyp info spawn args   '/usr/local/lib/node_modules/npm/node_modules/node-gyp/addon.gypi',
	gyp info spawn args   '-I',
	gyp info spawn args   '/Users/thomas/.node-gyp/0.8.6/common.gypi',
	gyp info spawn args   '-Dlibrary=shared_library',
	gyp info spawn args   '-Dvisibility=default',
	gyp info spawn args   '-Dnode_root_dir=/Users/thomas/.node-gyp/0.8.6',
	gyp info spawn args   '-Dmodule_root_dir=/Users/thomas/Desktop/holstground/node_modules/canvas',
	gyp info spawn args   '--depth=.',
	gyp info spawn args   '--generator-output',
	gyp info spawn args   'build',
[...]

PR #574, opened 2015-07-02
Fix include paths for cairo

LinusU:
This will need some verification before we can merge, but I think that this is the right thing to do.

`pkg-config` gives the path to a folder that includes the header files for `cairo` right away. This is different for e.g. `fontconfig` which gives a path to a folder that contains a single folder named `fontconfig`.

In the former case `#include <cairo.h>` should be used, but in the latter case `#include <fontconfig/fontconfig.h>` should be used.

This patch fixes the `cairo` includes to point at `<cairo.h>`.

I believe that the reason why it worked before (on linux at least, it seems to be broken on OS X) is because the search path included `/usr/local/include` which in turn contained the folder that (called `cairo`) that `pkg-config` pointed to.

The last proof for that the correct linking should be `<cairo.h>` is the [FAQ page on Cairo's official website](http://cairographics.org/FAQ/).

This pull request can be tested simply by running the following command:

```bash
npm install LinusU/node-canvas#include-paths
```

### DIFF
--- a/src/Canvas.cc
+++ b/src/Canvas.cc
@@ -12,8 +12,8 @@
 #include <string.h>
 #include <node_buffer.h>
 #include <node_version.h>
-#include <cairo/cairo-pdf.h>
-#include <cairo/cairo-svg.h>
+#include <cairo-pdf.h>
+#include <cairo-svg.h>
 #include "closure.h"
 
 #ifdef HAVE_JPEG
--- a/src/Canvas.h
+++ b/src/Canvas.h
@@ -16,7 +16,7 @@
 #if HAVE_PANGO
 #include <pango/pangocairo.h>
 #else
-#include <cairo/cairo.h>
+#include <cairo.h>
 #endif
 
 #include <nan.h>
--- a/src/PNG.h
+++ b/src/PNG.h
@@ -2,7 +2,7 @@
 #define _CANVAS_PNG_H
 #include <png.h>
 #include <pngconf.h>
-#include <cairo/cairo.h>
+#include <cairo.h>
 #include <stdlib.h>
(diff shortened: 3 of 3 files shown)


LinusU:
Just tested this on Ubuntu 13.10 and `pkg-config` gives `/usr/include/cairo` which includes the files right away. I thus believe that this patch is correct on linux as well.

LinusU:
Hmm, the Travis CI build fails on 0.8 for an unrelated reason. It seems like it can't find `nan` on npm?

Anyways, it works on the three other instances so this really seems like the right thing to do :+1:

LinusU:
@TooTallNate Any change of merging this?

It should fix #572, #514, #217 and also fix installation on Mac OS X with cairo from brew.

LinusU:
@TooTallNate Is there anything more you want provided for this to get merged?

wonderdogone:
I built this (node-gyp rebuild) and then included in my node V12.6 project and can confirm canvas now working back as normal. Thanks for this.

LinusU:
Rebased on latest master.

The tests now works on Nodejs 0.8 but fails on iojs because it's trying to get the source for iojs from the nodejs website. It fails on the master branch as well and it's not the fault of this pull request.

@TooTallNate I'm sorry for spamming you but I would really like this merged. We are trying to use node-canvas in production but this is a huge blocker for that.

garthk:
I'm also blocked on this, @TooTallNate, and have to fork two other projects to take a branch dep on @LinusU. Merging would help a great deal.

LinusU:
Thanks :beers:

@TooTallNate Could you please cut a new release that we can depend on as well :+1:

PR #575, opened 2015-07-02
Fix pango detection

LinusU:
On my Ubuntu 13.10 machine the pango detection returns true without me having the library installed. This makes the build fail later.

This patch fixes the detection of pango by using the builtin feature `--exists` of `pkg-config`.

This pull request can be tested simply by running the following command:

```bash
npm install LinusU/node-canvas#detect-pango
```

### DIFF
--- a/binding.gyp
+++ b/binding.gyp
@@ -13,7 +13,7 @@
         'with_jpeg%': '<!(./util/has_lib.sh jpeg)',
         'with_gif%': '<!(./util/has_lib.sh gif)',
         # disable pango as it causes issues with freetype.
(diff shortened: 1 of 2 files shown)


kangax:
I'm not on Linux so can't test. If anyone can confirm this, that would be great!

LinusU:
Btw. the Travis builds fails for an unrelated issue. The node 0.8 test can't handle `^` in the `package.json`.

LinusU:
Rebased on master, I would like to get this merged since I think it will solve some of the install issues that we are seeing...

LinusU:
Superseded by #701

PR #592, opened 2015-07-26
rename package variable to packageJson

miguelmota:
`package` is a future reserved keyword. Throws error in iojs. [MDN Future reserved keywords](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#Future_reserved_keywords)


LinusU:
Merging this :+1:

PR #580, opened 2015-07-13
Upgrade NAN to v2

rvagg:

PR #604, opened 2015-08-01
Replace CanvasPixelArray with Uint8ClampedArray; optimizations; bugfixes

zbjornson:
This updates quite a bit of stuff:

- ImageData's underlying data should be a Uint8ClampedArray, not CanvasPixelArray (issue #403). I think this also unblocks #569 and #580. This replaces PR #498, which also fixed the issue but was not compatible past node 0.10. The API for ImageData should now be fully spec-compliant!

- Optimization to putImageData -- benchmarked 53% faster for a 300 x 600 pixel canvas.

- Fix two failing test cases (putImageData 8 and 9 from the browser suite). All putImageData/getImageData browser tests pass now.

- Updates the browser-based test suite so it runs. Reveals failing test cases that I think correspond to #512 and probably other drawImage bugs (which seems quite borked in the most recent releases of node-canvas).

Tested primarily on node 2.4.0 (iojs) but uses v8 APIs available to node 0.12 and later. Node 3.0 (when v8 upgrades to 4.4) will need a quick API update but I'm poised to make those changes. I can't get node-gyp to work with iojs 3.0 nightlies right now, else I'd put in compiler directives.


zbjornson:
CI failed for node 0.8 and 0.10 as expected. Package.json specifies the engine must be 0.12 or later.

LinusU:
This is really cool! Good job :+1: 

- What is warning 4512?
- Which is the api call only available on >0.12?

Tests is passing on my local machine, code looks good. I would like it if we could support Node.js 0.10 but maybe that just isn't viable.

Could you update the `.travis.yml` file to remove Nodejs 0.8 and 0.10? That way the tests will pass.

LinusU:
Realized from the failed tests that it's `ArrayBuffer` and `UInt8ClampedArray` that isn't available on Node.js 0.10.x so I guess we can't support that...

zbjornson:
Thanks :)

* Warning 4512 is "assignment operator could not be generated" -- google briefly but didn't see a quick fix. Messages were not introduced by my changes, and if I recall only appear on more recent versions of node.
* Ya, the TypedArray stuff (which is what changes again in v8 4.4, ugh).
* Updated .travis.yml

LinusU:
Could you amend the last commit to also remove line 6 of the travis config, that was specific only to node.js version 0.8.

zbjornson:
Done

zbjornson:
As far as node 0.10 support, could make this version 2 and update the readme to be clear that v2 is for node 0.12+ only.

LinusU:
Perfect :+1: 

Looks good to me, I'm happy to merge but I'll give the other maintainers some time to give their comments.

I also think that it might be a good idea to release this as version 2.0.0. Need to think a bit about this.

kangax:
Yep, backwards-incompatible changes need new major version since we follow semver

LinusU:
@zbjornson I think I detected a small "bug" with your performance fix thought. When `a` is `0` all the other values should be zero as well since it uses pre-multiplied values. It might not affect anything since it will probably skip drawing that anyhow.

Maybe it's better to check if `a` is `0` and then just set all the other values to `0`, we don't even need to read them from the source.

zbjornson:
@LinusU I was actually a bit confused about that and should have brought it up. I saw someone else added the equivalent-ish to that to getImageData (https://github.com/Automattic/node-canvas/pull/549/files), but I didn't understand why the same behavior should happen for `a == 0` as for `a == 255`. Because going to/from premultiplied RGB values is lossy, it seems like there's no way to recover the value at all if a == 0, so you just have to return the premultiplied value in that case I guess (which is 0)?

(The biggest gain actually comes from how the source is accessed and destination is set -- working entirely with uint8s instead of bitshifting in/out of a uint32.) 

This look okay to you?
```cpp
// Context2d::PutImageData
if (a == 0) {
  *dstRow++ = 0;
  // ...
} else if (a == 255) {
  *dstRow++ = b;
  // ...
} else {
  *dstRow++ = b * alpha;
  // ...
}
```

LinusU:
That looks very good, I also think that the intent is clearer. When I look at that I immediately understand it.

Also, in GetImageData, couldn't we do something like this to get rid of the bitshifting there?

```diff
-      uint32_t *pixel = row + x + sx;
-      uint8_t a = *pixel >> 24;
-      uint8_t r = *pixel >> 16;
-      uint8_t g = *pixel >> 8;
-      uint8_t b = *pixel;
+      uint8_t *pixel = row + x + sx;
```

And then access as `pixel[0]` instead of `a`, `pixel[1]` instead of `r`, etc... My C++ skills are bit rusty so there might be a reason why we can't do this.

zbjornson:
Yeah, we can do the same thing in GetImageData -- I tried it and it was only slightly faster (not worth the regression risk), but I'll benchmark it again now that I'm more comfortable with what these functions do and have the test suit working. Commit coming in a few hours. Thanks!

zbjornson:
OK, bug fixed in putImageData.

Benchmarked two variations of GetImageData, both of which were slower: `uint8_t *pixel = row + x + sx; uint8_t a = pixel[0];` and `uint8_t *pixel = row + s + sx; uint8_t a = *pixel++;`. Left as-is.

(Aside, I tried using the benchmarking suite in the repo, which gave some really surprising timings. For example, `fillRect` took 3.5 seconds for one iteration and consumed a ton of memory. That looks like essentially a pass-through function to cairo, so wonder if it's a bug in cairo [on Windows].)

LinusU:
This would be very good to merge before any more work goes into moving towards `nan` 2.0.

@zbjornson Is there anything more you would like to work on or do you feel comfortable with merging this?

zbjornson:
@LinusU I'm good with merging. Thanks!

Hopefully sometime soon I can work on some other bugs :)

kkoopa:
This still uses the external array data API, which has been entirely removed.

zbjornson:
Right -- it's removed from the version of v8 used by node 3.0. (That's the API change I was referring to in my original PR message.) As soon as node-gyp works with iojs 3, I will add compiler directives that change which API is used. The API we need to switch to is not present in node 0.12-2.x (the non-externalizing API).

kkoopa:
node-gyp works with io.js 3, so does pangyp

kkoopa:
Would it not be possible to retain support for 0.10 (and / or 0.8) via https://github.com/joyent/node/blob/v0.10.40-release/src/v8_typed_array.cc ?

zbjornson:
Possibly -- let me look at that. Will also try to get node-gyp working with iojs 3 again.

LinusU:
@zbjornson Did you get a chance to look at the part that **kkoopa** mentioned?

zbjornson:
Sorry -- swamped last week. Working on it today.

LinusU:
No problem, tell me if you need help with anything!

zbjornson:
Erm, do I need to fork (or cherry-pick) from https://github.com/Automattic/node-canvas/pull/580 to get a version of nan that works with iojs 3?

@rvagg

LinusU:
I don't think that that branch is done yet. My plan was to get this merged and then I would take a stab at completing the Nan switch.

Do you need anything from Nan 2.x?

zbjornson:
It looks like Nan 2.0 is required for iojs 3.0, so I can only work on supporting node 0.8/0.10 without it.

I'm working on 0.8/0.10 now. Perhaps after I finish that, we merge this branch with support only through 2.x, update/rebase #580 and then I can make the IndexedProperties changes for 3.0 there? Let me know what you want to do.

LinusU:
That sounds perfect :+1:

zbjornson:
Done. No regressions in the browser/visual tests. Set travis to iojs 2.4.0, but 2.5.0 works as well.

src/CanvasRenderingContext2d.cc
```diff
@@ -654,6 +678,125 @@ NAN_METHOD(Context2d::PutImageData) {
...
+
+  clampedArray->SetIndexedPropertiesToExternalArrayData(dst, kExternalPixelArray, size);
+#else
+  Local<ArrayBuffer> buffer = ArrayBuffer::New(Isolate::GetCurrent(), size);
+  Local<Uint8ClampedArray> clampedArray = Uint8ClampedArray::New(buffer, 0, size);
+#if NODE_MAJOR_VERSION < 3 // includes node 0.12
+  clampedArray->SetIndexedPropertiesToExternalArrayData(dst, kExternalUint8ClampedArray, size);
+#else
+  // TODO
+#endif
```
LinusU: Instead of doing nothing if `NODE_MAJOR_VERSION < 3` I would rather that we gave a compiler error. Basically, I would prefer it if line 754, 756, 757 and 758 where gone.

Makes sense?

LinusU:
@zbjornson Could you remove your changes to `.travis.yml` and rebase on `master`? I've added in specific iojs versions there.

zbjornson:
All done!

rvagg:
@zbjornson if you want to jump to NAN v2, you can either battle with merging in my PR or you can do it yourself. Use this script https://gist.github.com/thlorenz/7e9d8ad15566c99fd116 as a first pass but it'll need some love afterwards. Mostly, the `Nan::New<String>().ToLocalChecked()` replacements will get messed up because it doesn't count parens, so you'll have `.ToLocalChecked()` dangling in all the wrong places. Also, try and replace all references to `Handle` to `Local` while you're at, the former is being deprecated in the next V8 so now's the time to move away from it.

zbjornson:
@LinusU are you waiting for anything from me? Probably best to do the nan 2 upgrade in a separate PR.

LinusU:
Sorry for the delay, this is awesome! Merging now :)

I'll release this as a `patch` bump since it's backwards compatible, and since it doesn't introduce any new functionality, sounds good?

zbjornson:
Sounds perfect. Will start working on iojs 3/nan 2 this week. Thanks!

LinusU:
So I somehow managed to miss the actual cutting of a release. I'll get this out asap.

LinusU:
Okay, here is what happened. I tagged the release, pushed it to npm, updated `History.md` and then pushed it *to my own fork*.

I have now pushed it to the main repo :+1:

PR #611, opened 2015-08-23
travis: specify iojs versions

LinusU:
Specify which iojs versions to test against. Maybe I went a bit over board with which versions to test against; but I don't really know which ones I should choose. Maybe only the latest 1.x and 2.x?


LinusU:
Nice to see that it worked on almost every version of iojs, but the tests had to run for over 2 hours :(

I think that only `1.8.4` and `2.5.0` will be best.

LinusU:
I'm merging this now to stop the failing builds.

PR #612, opened 2015-08-23
test: fix up some small issues

LinusU:
- Adds `done` to async tests
- Removes `should` (was only used on three lines, compared to `assert` which was used everywhere)
- Don't allow `Infinity` for `lineWidth` and reenable test for it
- Reenable and fix test for `parseFont`, also trim the resulting `fontFamily`


LinusU:
Merging this

PR #615, opened 2015-08-25
Don't unbind onload/onerror callbacks after invoking them

rochoa:
From "[WHATWG HTML Living Standard]
(https://html.spec.whatwg.org/multipage/indices.html#event-load)", the definition of 'onload':
> Fired at the Window when the document has finished loading; fired at an
element containing a resource (e.g. img, embed) when its resource has finished
loading

To adhere to that specification both handlers should be called as many times as Image loads.

From Google Chrome's Developer Tools:

```js
> function handler(evt) { console.log('img handler: %s', evt.type); }
< undefined
< undefined
> img.onload = handler
< handler(evt)
> img.onerror = handler
< handler(evt)
> img.src = 'https://www.google.com/images/errors/logo_sm_2.png'; true
< true
< img handler: load
> img.src = 'https://www.google.com/images/errors/logo_sm_2.png'; true
< true
< img handler: load
> img.src = 'https://example.com/404.png'; true
< true
< img handler: error
> img.src = 'https://example.com/404.png'; true
< true
< img handler: error
```


rochoa:
Hi @LinusU, what do you think about this one?

test/image.test.js
```diff
@@ -7,6 +7,7 @@ var Canvas = require('../')
   , Image = Canvas.Image
   , assert = require('assert');
 
+var png_checkers = __dirname + '/fixtures/checkers.png';
 var png = __dirname + '/fixtures/clock.png';
```
LinusU: Can you rename this variable to `png_clock`?

test/image.test.js
```diff
@@ -35,6 +36,36 @@ describe('Image', function () {
     assert.strictEqual(320, img.height);
   });
 
+  it('test Image#onload multiple times', function() {
+    var img = new Image
+      , n = 0;
```
LinusU: Can you call this variable `onloadCalled` as in the other tests? This applies to all your uses of `n`.

LinusU:
I like! Just a few nits and then we can :ship:!

rochoa:
Thanks for your feedback! I already applied those changes.

test/image.test.js
```diff
@@ -53,18 +84,48 @@ describe('Image', function () {
...
+    var img = new Image
+      , onloadCalled = 0;
+
+
+    img.onload = function() {
+      assert.fail('called onload');
+    };
+
+    img.onerror = function() {
+      ++onloadCalled;
```
LinusU: Mind changing to `onloadCalled += 1` for consistency with the rest of the file? I personally think it feels a bit weird to put them in front of the variable instead of after, although I understand that technically you don't want it to retain the old value and return that from this particular expression... There is some debate on wether to use `++` and `--` at all, I think that Crockford recommends against it in "Javascript: the good parts". I think that in this case `onloadCalled += 1` shows the intent very clearly.

rochoa:
Glad to change it, it makes sense, specially having in mind your explanation and the context.

I also changed the name for Image#onerror `onerrorCalled` variable, as I was using `onloadCalled`.

LinusU:
Awesome! Thank you for your quick updates, merging now :fireworks:

LinusU:
Released as `1.3.7`

rochoa:
Thank YOU! :+1:

PR #616, opened 2015-08-25
Allow to unbind onload/onerror callback handlers

rochoa:
Assigning `null` to handlers removes them, this mimic the behaviour from Google Chrome.

From Chrome's Developer Tools:

```js
> function handler(evt) { console.log('img handler: %s', evt.type); }
< undefined
< undefined
> img.onload = handler
< handler(evt)
> img.onerror = handler
< handler(evt)
> img.src = 'https://www.google.com/images/errors/logo_sm_2.png'; true
< true
< img handler: load
> img.src = 'https://www.google.com/images/errors/logo_sm_2.png'; true
< true
< img handler: load
> img.src = 'https://example.com/404.png'; true
< true
< img handler: error
> img.src = 'https://example.com/404.png'; true
< true
< img handler: error
> img.onload = null
< null
> img.onerror = null
< null
> img.src = 'https://www.google.com/images/errors/logo_sm_2.png'; true
< true
> img.src = 'https://example.com/404.png'; true
< true
```


rochoa:
Travis build for node 0.8.x failed due:

```
curl: (18) transfer closed with 34544884 bytes remaining to read
```

LinusU:
Restarted the build to see if it works now, would be good to rebase on master to get full test suite running thought.

LinusU:
Hi @rochoa, I would like to land this patch. Could you rebase on master? We have changed how the tests is structured but it should be an easy update :+1:

rochoa:
Hi @LinusU I finally took the time to update the PR with latest master.

Hope that helps.

test/image.test.js
```diff
@@ -150,4 +150,56 @@ describe('Image', function () {
     image.src = new Buffer(0);
     image.src = new Buffer('');
   });
+
+  it('should unbind Image#onload', function() {
+    var img = new Image
+      , onloadCalled = 0;
+
+    img.onload = function() {
+      ++onloadCalled;
```
LinusU: If you wouldn't mind, please change this to `onloadCalled += 1` because of the reasons we discussed before... Thanks

LinusU:
Just one small nit, thank you for your patience!

rochoa:
Sorry, that was on my side. I should have considered updating it after previous PR feedback :smile:.

src/Image.cc
```diff
@@ -268,6 +268,12 @@ NAN_SETTER(Image::SetOnload) {
   if (value->IsFunction()) {
     Image *img = Nan::ObjectWrap::Unwrap<Image>(info.This());
     img->onload = new Nan::Callback(value.As<Function>());
+  } else if (value->IsNull()) {
+    Image *img = Nan::ObjectWrap::Unwrap<Image>(info.This());
+    if (img->onload) {
+      delete img->onload;
+    }
+    img->onload = NULL;
```
LinusU: Sorry, one last thing again :laughing: 

Could you replace `NULL` with `nullptr` since we are always building with C++11. This is better because `NULL` is just a macro for `0` whereas `nullptr` is a value of a the special type `std::nullptr_t`. Because of that it can never accidentally be used as an int, or really anything other than a pointer.

I promise I will merge and release after this :dancer:

rochoa: I was using NULL here as it is used when initialising a new Image :smiling_imp:.

Shouldn't we use nullptr across all the project then?

LinusU: Looked thru the codebase and you are absolutely right, I'll send another PR for that.

LinusU:
I'm going to release this a patch release since the previous behaviour was very unexpected and since it wasn't following the spec. Therefor I consider this a bugfix.

LinusU:
Thanks for all your work!

LinusU:
Shit I really managed to screw this up :( I'll manually merge this now...

LinusU:
Released as 1.3.9 \o/

rochoa:
Thank you @LinusU again, for your patience and good feedback :+1:

Issue #624, opened 2015-09-11
Cannot use custom font


sungwoncho:
I am following the `example/font.js` to use a custom font, but the font does not show up.

Here is what I have so far:

```javascript
var Canvas = require('canvas')
    , Image = Canvas.Image
    , Font = Canvas.Font
    , path = require('path');

var gubblebum = new Font('gubblebum', fontFile('GUBBLO___.ttf'));

function fontFile(name) {
  return path.join(__dirname, '../fonts', name);
}

function draw() {
  var canvas = new Canvas(100, 100)
    , ctx = canvas.getContext('2d');

  ctx.addFont(gubblebum);
  ctx.font = 'bold 40 gubblebum';
  ctx.fillText('test', 25, 45);

  return canvas;
}

module.exports = draw;
```

When rendered in a browser, the font is left unchanged from the default, as well as the size.

Interestingly, when I do `ctx.font = 'bold 40 someGibberish';` the font size is applied correctly.

chearon:
is this happening on OS X? I'm trying to narrow down on the issue (see #628)

sungwoncho:
Yes it happened on OSX Yosemite.

Issue #618, opened 2015-08-28
font loading issue in latest version


diwu1989:
I see this when deploying 1.2.x to heroku with the proper buildpack
```
2015-08-28T19:01:32.628424+00:00 app[web.1]: /app/node_modules/canvas/lib/canvas.js:80
2015-08-28T19:01:32.628428+00:00 app[web.1]:     var face = new FontFace(path, idx || 0);
2015-08-28T19:01:32.628430+00:00 app[web.1]:                ^
2015-08-28T19:01:32.628432+00:00 app[web.1]: Error: Could not load font in FontConfig
2015-08-28T19:01:32.628433+00:00 app[web.1]:     at Error (native)
2015-08-28T19:01:32.628437+00:00 app[web.1]:     at Font.addFace (/app/node_modules/canvas/lib/canvas.js:80:16)
2015-08-28T19:01:32.628438+00:00 app[web.1]:     at new Font (/app/node_modules/canvas/lib/canvas.js:73:10)
2015-08-28T19:01:32.628440+00:00 app[web.1]:     at Object.<anonymous> (/app/routes/index.js:13:12)
2015-08-28T19:01:32.628441+00:00 app[web.1]:     at Module._compile (module.js:460:26)
2015-08-28T19:01:32.628443+00:00 app[web.1]:     at Object.Module._extensions..js (module.js:478:10)
2015-08-28T19:01:32.628444+00:00 app[web.1]:     at Module.load (module.js:355:32)
2015-08-28T19:01:32.628445+00:00 app[web.1]:     at Function.Module._load (module.js:310:12)
2015-08-28T19:01:32.628447+00:00 app[web.1]:     at Function.cls_wrapMethod (/app/node_modules/newrelic/lib/shimmer.js:230:38)
2015-08-28T19:01:32.628449+00:00 app[web.1]:     at Module.require (module.js:365:17)
2015-08-28T19:01:32.635535+00:00 app[web.1]: /app/node_modules/canvas/lib/canvas.js:80
2015-08-28T19:01:32.635542+00:00 app[web.1]:     var face = new FontFace(path, idx || 0);
```
switching back to 1.1.x works

dotnil:
I've got this issue with 1.3.5 on RHEL too. With pango disabled as the way detailed in #628 , this issue no longer persist. So +1 from me for disabling pango.

Issue #570, opened 2015-06-24
Pango nightmares


rvagg:
I just upgraded https://nodei.co and had to yak shave my way to find a solution to my custom fonts not showing up any more. I believe related to this problem are #548 and #561. The workaround was to revert pango being enabled as it was in 7522aca2ea7d4e38752cfef807b31ba1a419d58f at which point everything worked again.

Scanning some of the recent font related problems in the issues I wouldn't be surprised if this is biting a large number of people and perhaps we should view it as a breaking change and revert that commit (or at least pango being enabled).

jakeg:
Personally I'd like to see making pango a requirement.

As you say, lots of issues/frustrations have come up recently in relation to font problems, and I think part of this is that some people are installing with pango, others without, and without pango some things (e.g. fallback fonts) just aren't possible.

Is there a good reason for not making pango a requirement?

Currently we (creating printed yearbooks for hundreds of customers) use our fork of node-canvas that I haven't touched the code on since December because I know it works and I'd be very nervous about upgrading to the non-forked version with the plethora of font-related bugs I've seen lately.

rvagg:
if it's to become a requirement then we should bump major version to indicate breakage - my use is simple, load a font by pointing to a particular path and then explicitly insert that to the context and refer to it by that name - if pango become necessary then we'd need a migration path for that usage to a pango-happy usage and tbh I don't know what that is (I've messed with fontconfig to no avail)

chearon:
Closed via #715

PR #628, opened 2015-09-14
disable pango

rvagg:
I can't get custom fonts (in custom locations) working when pango is enabled and there's no easy way to turn it off other than editing binding.gyp. This commit basically reverts https://github.com/Automattic/node-canvas/commit/7522aca2ea7d4e38752cfef807b31ba1a419d58f

See also discussion @ https://github.com/Automattic/node-canvas/issues/570

Unless custom font loading can be fixed with Pango or good documentation provided then I'd like to see it disabled.

Likely related are https://github.com/Automattic/node-canvas/issues/624 and https://github.com/Automattic/node-canvas/issues/618

### DIFF
--- a/binding.gyp
+++ b/binding.gyp
@@ -13,7 +13,7 @@
         'with_jpeg%': '<!(./util/has_lib.sh jpeg)',
         'with_gif%': '<!(./util/has_lib.sh gif)',
         # disable pango as it causes issues with freetype.
-        'with_pango%': '<!(./util/has_lib.sh pangocairo)',
+        'with_pango%': 'false',
         'with_freetype%': '<!(./util/has_cairo_freetype.sh)'
       }
     }]


LinusU:
Have you looked at how this relates to #410 and #543?

rvagg:
haven't looked in detail, but something needs to happen because 7522aca was a breaking change that crept in on a minor and by the issues filed here it's clear that it's causing a lot of pain

LinusU:
Also, on my computer (OS X 10.11, was also present in 10.10) trying to render text to the canvas, when built with pango, breaks every further draw call without giving any errors. If I start by rendering text, I get a blank canvas, otherwise I get whatever I drew up 'til the first text call.

I'm very much +1 on merging this until we can get it working stable across the board.

jakeg:
Very much against this. We use pango for font fallbacks and not having it would be a show stopper for us. Is it just on osx where people are having problems with this? My usage is Linux (Ubuntu) only.

jakeg:
Isn't this caused by the issue shown in #525 ? Or is it something completely different?

rvagg:
my problem is Linux, at this stage I have no clue how to make custom fonts work for nodei.co without this patch applied, Pango has entirely screwed up the config, this should have been done in a minor version bump if at all

LinusU:
@jakeg What do you mean with font fallbacks? I'm not entirely sure myself on what pango does for us so it would be nice with some insight :)

jakeg:
@linusu firstly, big thank you for all your work the last few weeks, the project was looking rather abandoned before you turned up!

Try using a glyph that isn't within the currently set font. Pango enables you to set multiple fonts to use as fallbacks for those glyphs, but without pango the glyph doesn't get rendered at all! Try eg Chinese, Korean or Arabic letters.

LinusU:
Thank you, it's always nice to know that your work is appreciated :)

This helps me understand the issue as well, I'm gonna try and dig deeper into the issues I'm having and see if there is a way to properly solve it.

jacobp100:
+1 for this being applied. Using Ubuntu, it doesn't crash, it just doesn't load custom fonts.

dotnil:
Guys, any update on this? :confused:

tomerb15:
@LinusU or @jakeg Is this fixed? I have node installed on a Ubuntu server and I'm having issues with using custom fonts.
This works great on my development computer (macbook pro with the latest OSX)

What I'm experiencing is that the custom font which is located in the same files hierarchy is completely ignored, and the default font is used. Also the font size is also ignored.

Any idea how can i fix it on the Ubuntu server?

LinusU:
@tomerb15 I had the exact problem myself, with the same setup (OS X locally, Ubuntu in the cloud). Manually changing the `binding.gyp` did work for me though.

I would love to see if #410 + #543 could solve this, it would be better then disabling pango since some of our users relay on it.

tomerb15:
What solved the problem was commenting out:
ctx.addfont(Fontobj) on the Ubuntu server.
Locally on the Osx this line was still needed.

LinusU:
Did you recompile (`node-gyp rebuild`) after editing `binding.gyp`?

tomerb15:
Nope :-) probably should have. But like I said just commenting out:
ctx.addfont
Did the trick as well

LinusU:
Then it's probably loading the font from the system fonts instead of from your file.

chearon:
+1

Or get Pango working somehow. Seems like a lot of issues can be closed by getting Pango to work or removing it.

Also:

* The changes from #410 are already in master from @jakeg's changes in 7522aca2. it should be closed
* #543 did not help this problem for me. The only thing that gets custom fonts working for me is disabling Pango, but I hate having to tell everyone to do that on their computers and then avoid updating

jakeg:
Ok, going through my project files looking for all relevant stuff for this as I can see this is really coming to a head for some people now...

As far as I can see, you can use Pango without system-installed fonts, but fallback fonts do need to be system installed (and I think you need to have/use fontconfig for them too?) if you want them to work. You don't get fallback fonts without pango, so no loss there if you can't do system installed fonts.

What are fallback fonts? So if you're using a font and a glyph doesn't exist in that font, you want that glyph displayed in a different font - the fallback font - instead (otherwise it will just be e.g. an empty rectanglle). I run this command to install a ton of fallback fonts on my Ubuntu server:

`sudo apt-get install fonts-arphic-uming fonts-ipafont-mincho ttf-wqy-zenhei ttf-unfonts-core fonts-takao fonts-dejavu ttf-mscorefonts-installer`

So as well as system-installing those fallback fonts, I go and grab all the Google web fonts too with this handy little script I made:
https://twitter.com/AllYearbooks/status/512937396552212480

Note that for sanity, I use FVD naming (see https://github.com/typekit/fvd) for fonts and my script above handles that perfectly, renaming all the Google web fonts with sane FVD naming instead.

With those fonts all installed, I then use something like this to actually use the fonts. Remember, as we're using pango, we can't do `ctx.addFont()`. `new Canvas.font()` yes `myfont.addFace()` yes, `ctx.addFont()` no.

    var myfonts = [];
    var addedfonts = [];

    // then for each font
    var font = 'Roboto Slab:n7'; // example font using FVD naming as above
    var bits = font.split(':');
    var name = bits[0].replace(/ /g, '_'); // global match else just replaces first
    var filename = name + '.' + bits[1] + '.ttf';

    if (fs.existsSync(dir + filename)) {
      if (!myfonts[bits[0]]) {
        // add first variant for this font
        //console.log('adding font', bits[0]);
        myfonts[bits[0]] = new Canvas.Font(bits[0], dir + filename);
        addedfonts[bits[0]] = {};
      }

      // add face/variant to existing font
      // ... this bit used to be in else block but surely that wouldn't work sometimes?
      if (!addedfonts[bits[0]][bits[1]]) {
        //console.log('adding face', bits[0], bits[1]);
        myfonts[bits[0]].addFace(dir + filename, bits[1][1] + '00', bits[1][0] == 'i' ? 'italic' : 'normal'); // path, weight, style
        addedfonts[bits[0]][bits[1]] = true;
      }
    }
    else {
      // font file not saved on server yet. fetch it then use it
      // TODO: as this is async, if font hadn't already been saved then won't use it for this request, only subsequent ones. fix this
      console.log('font file not found. try running google_fonts_downloader.js again?', dir + filename);
    }

Note that if you installed all the fallback fonts, no matter what font you are currently set to in ctx the following should output fine as all the missing glyphs in that font will find one to fallback on:

    ctx.fillStyle = '#000';
    ctx.fillText("English: Some β text in Impact.", 10, 10);
    ctx.fillText("Japanese: 図書館の中では、静かにする。", 10, 30);
    ctx.fillText("Arabic: اللغة العربية هي أكثر اللغات تحدثا ضمن", 10, 50);
    ctx.fillText("Korean: 모타는사라미 못하는 사람이", 10, 70);

jakeg:
If it wasn't clear from my post above, only the fallback fonts need to be installed system-wide, not your regular usage fonts. All the other fonts can go wherever you want (though may require fontconfig being used/installed?) and get added with `new Canvas.Font('Roboto slab', '/path/to/somefont.ttf')`. I believe being consistent with the font naming is crucial though, hence one reason I use FVD.

chearon:
The fallback font stuff comes from FontConfig, though, no? The way the code is now you would get the font substitution like you said which is great.

I understand that with Pango support you just use `new Canvas.Font` and it should work with a custom TTF. I spent hours reading the source yesterday; calling the Font constructor calls `FontFace`, which is what adds it to FontConfig [here](https://github.com/Automattic/node-canvas/blob/master/src/FontFace.cc#L84). Those lines are executing but Pango still seems to end up with the wrong font. Something is going wrong between FontConfig and Pango for those of us that can't get it working

chearon:
OH I think this is an OS X issue. Pango doesn't use FontConfig on OS X! It needs to use CoreText somehow in this case... that'd explain why the people who have no problem with it are always using Ubuntu...

jakeg:
@chearon pango can't use fontconfig or you mean because fontconfig isn't installed on OSX?

http://apple.stackexchange.com/questions/118737/does-osx-10-9-1-have-fontconfig-installed-by-default
... suggests you can install fontconfig using MacPort (`sudo port install fontconfig`) or Homebrew (`sudo brew install fontconfig`).

chearon:
@jakeg I suspected that OS X Pango was using something other than FontConfig to look up fonts. Now I'm not sure about that, but the problem lies somewhere with FontConfig + Pango. I tried making a standalone C program that loaded a font from a file and added it to FontConfig exactly the way the creator of Pango recommends, and it did not render on Mac OS :(

LinusU:
@jakeg Please never use `sudo` with `brew`!

@chearon Appreciate you looking into this! I should add that loading fonts from ttf-files doesn't work for me on Ubuntu 14.04 when built with pango. When building without it works. I have the same problem on OS X as well, so I don't think it's only an OS X issue...

chearon:
@LinusU you didn't call `ctx.addFont`, right? If you use `addFont` then when you set `ctx.font` later, it will not update the Pango context, instead using the old Cairo methods which basically has no effect. Another thing I think should be fixed

zbjornson:
Since v2.0 made Pango *required,* can this be closed?

chearon:
Yup! I'm going to go through now and close all the issues I listed forever ago in #715

PR #635, opened 2015-09-28
Fix segfault in Canvas#jpegStream. Fix #629

### LINKED ISSUE #629: jpegStream causes segfault
On Node v4.0.0.
There is no crash with 0.12 but the resulting file is not a valid jpeg.

```js
var fs = require('fs');

var canvas = new Canvas(640, 480);
var out = fs.createWriteStream('./test.jpg');
var stream = canvas.jpegStream();
stream.pipe(out);
```

http://pastebin.com/raw.php?i=qpi1Dzfv

zbjornson:
Fixes #629 and adds a test case for JPEGstream.

### DIFF
--- a/src/JPEGStream.h
+++ b/src/JPEGStream.h
@@ -31,13 +31,17 @@ boolean
 empty_closure_output_buffer(j_compress_ptr cinfo){
   Nan::HandleScope scope;
   closure_destination_mgr *dest = (closure_destination_mgr *) cinfo->dest;
+
   Local<Object> buf = Nan::NewBuffer((char *)dest->buffer, dest->bufsize).ToLocalChecked();
-  Local<Value> argv[3] = {
+
+  // emit "data"
+  Local<Value> argv[2] = {
       Nan::Null()
     , buf
-    , Nan::New<Integer>(dest->bufsize)
   };
-  Nan::MakeCallback(Nan::GetCurrentContext()->Global(), (v8::Local<v8::Function>)dest->closure->fn, 3, argv);
+  Nan::MakeCallback(Nan::GetCurrentContext()->Global(), (v8::Local<v8::Function>)dest->closure->fn, 2, argv);
+
+  dest->buffer = (JOCTET *)malloc(dest->bufsize);
   cinfo->dest->next_output_byte = dest->buffer;
   cinfo->dest->free_in_buffer = dest->bufsize;
   return true;
@@ -47,26 +51,24 @@ void
 term_closure_destination(j_compress_ptr cinfo){
   Nan::HandleScope scope;
   closure_destination_mgr *dest = (closure_destination_mgr *) cinfo->dest;
-  /* emit remaining data */
-  size_t remaining = dest->bufsize - cinfo->dest->free_in_buffer;
-  Local<Object> buf = Nan::NewBuffer((char *)dest->buffer, remaining).ToLocalChecked();
 
-  Local<Value> data_argv[3] = {
+  /* emit remaining data */
+  Local<Object> buf = Nan::NewBuffer((char *)dest->buffer, dest->bufsize).ToLocalChecked();
+
+  Local<Value> data_argv[2] = {
       Nan::Null()
     , buf
-    , Nan::New<Number>(remaining)
   };
 
-  Nan::MakeCallback(Nan::GetCurrentContext()->Global(), (v8::Local<v8::Function>)dest->closure->fn, 3, data_argv);
+  Nan::MakeCallback(Nan::GetCurrentContext()->Global(), (v8::Local<v8::Function>)dest->closure->fn, 2, data_argv);
 
   // emit "end"
-  Local<Value> end_argv[3] = {
+  Local<Value> end_argv[2] = {
       Nan::Null()
     , Nan::Null()
-    , Nan::New<Integer>(0)
   };
 
-  Nan::MakeCallback(Nan::GetCurrentContext()->Global(), (v8::Local<v8::Function>)dest->closure->fn, 3, end_argv);
+  Nan::MakeCallback(Nan::GetCurrentContext()->Global(), (v8::Local<v8::Function>)dest->closure->fn, 2, end_argv);
 }
 
 void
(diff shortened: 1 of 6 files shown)


LinusU:
Could this maybe be solved by modifying the `jpeg_free_custom_allocations` to not free the buffer and instead let node do that (which it will if we are using `NewBuffer` instead of `CopyBuffer`)?

That way we would skip having to copy all the bytes in memory.

LinusU:
Than we would also have to change it to tell libjpeg to write to a new buffer instead of reusing the same. It should still be more performant thought since we are avoiding a copy to a newly allocated buffer that then gets freed.

```diff
-   cinfo->dest->next_output_byte = dest->buffer;
+   cinfo->dest->next_output_byte = malloc(dest->bufsize);
    cinfo->dest->free_in_buffer = dest->bufsize;
```

LinusU:
Than we could probably avoid the entire `jpeg_free_custom_allocations` function which is nice :+1:

LinusU:
This should do it:

```diff
diff --git a/src/JPEGStream.h b/src/JPEGStream.h
index a65fccf..2814890 100644
--- a/src/JPEGStream.h
+++ b/src/JPEGStream.h
@@ -31,14 +31,14 @@ boolean
 empty_closure_output_buffer(j_compress_ptr cinfo){
   Nan::HandleScope scope;
   closure_destination_mgr *dest = (closure_destination_mgr *) cinfo->dest;
-  Local<Object> buf = Nan::CopyBuffer((char *)dest->buffer, dest->bufsize).ToLocalChecked();
+  Local<Object> buf = Nan::NewBuffer((char *)dest->buffer, dest->bufsize).ToLocalChecked();
   Local<Value> argv[3] = {
       Nan::Null()
     , buf
     , Nan::New<Integer>(dest->bufsize)
   };
   Nan::MakeCallback(Nan::GetCurrentContext()->Global(), (v8::Local<v8::Function>)dest->closure->fn, 3, argv);
-  cinfo->dest->next_output_byte = dest->buffer;
+  cinfo->dest->next_output_byte = (JOCTET *)malloc(dest->bufsize);
   cinfo->dest->free_in_buffer = dest->bufsize;
   return true;
 }
@@ -49,7 +49,7 @@ term_closure_destination(j_compress_ptr cinfo){
   closure_destination_mgr *dest = (closure_destination_mgr *) cinfo->dest;
   /* emit remaining data */
   size_t remaining = dest->bufsize - cinfo->dest->free_in_buffer;
-  Local<Object> buf = Nan::CopyBuffer((char *)dest->buffer, remaining).ToLocalChecked();
+  Local<Object> buf = Nan::NewBuffer((char *)dest->buffer, remaining).ToLocalChecked();
 
   Local<Value> data_argv[3] = {
       Nan::Null()
@@ -97,16 +97,6 @@ jpeg_closure_dest(j_compress_ptr cinfo, closure_t * closure, int bufsize){
 }
 
-void
-jpeg_free_custom_allocations(j_compress_ptr cinfo){
-  closure_destination_mgr * dest;
-  dest = (closure_destination_mgr *) cinfo->dest;
-  if (dest->buffer) {
-    free(dest->buffer);
-    dest->buffer = NULL;
-  }
-}
-
 void
 write_to_jpeg_stream(cairo_surface_t *surface, int bufsize, int quality, bool progressive, closure_t *closure){
   int w = cairo_image_surface_get_width(surface);
   int h = cairo_image_surface_get_height(surface);
@@ -148,7 +138,6 @@ write_to_jpeg_stream(cairo_surface_t *surface, int bufsize, int quality, bool pr
   }
   free(dst);
   jpeg_finish_compress(&cinfo);
-  jpeg_free_custom_allocations(&cinfo);
   jpeg_destroy_compress(&cinfo);
 }
 ```

Also, the test didn't actually fail for me before applying the changes :sob:

**Comment by zbjornson** [comment]
@LinusU  sorry for the delay. Just added your changes, thanks. There's some slow, perpetual memory growth when I run this a lot, but it is porbalby from the streams/callbacks not being disposed of well -- I assume `jpeg_destroy_compress` frees `dest->next_output_byte`, and otherwise this looks safe.

Re: test not failing -- It failed for me with libjpeg-turbo on Windows running node 4.0, but I had no crash in that environment. Up to you if you want to repro and describe your fail case so we can make the test fail.

**Comment by LinusU** [comment]
No problem!

I don't think that `jpeg_destroy_compress` should free that since it's up to V8 to do it once the ownership has been transferred with `Nan::NewBuffer`. It might be a good idea to take a closer look at it, I'm not sure that the code a posted is 100% correct.

We should also drop the third argument in the `data` and `end` emit, it's not used as far as I can tell. See the stream documentation of node.js.

Hopefully I can have another look at this this Saturday.

**Comment by zbjornson** [comment]
I removed the undocumented third args of the emitters per your suggestion, but I'm wondering if they are used by anyone. (Your comment made me originally think they were required for the implementation of a Stream from C++.) They're not simply the length of the emitted chunk, but the amount of buffer space available:

```cpp
dest->bufsize - cinfo->dest->free_in_buffer
```

I made some more revisions to your CL -- the test might have passed but the output was wrong when I tried it again. This was required:

```diff
@@ -31,14 +31,18 @@ boolean
 empty_closure_output_buffer(j_compress_ptr cinfo){
//...
-  cinfo->dest->next_output_byte = (JOCTET *)malloc(dest->bufsize);
+  dest->buffer = (JOCTET *)malloc(dest->bufsize);
+  cinfo->dest->next_output_byte = dest->buffer;
```

I'm not sure if it's possible to add a better test to the mocha suite -- JPEG compression is apparently* nondeterministic so checking the bytes is not an option (*when I tried several times the results were always different). Instead I added a third column to the browser tests to show JPEG results. Because JPEG doesn't support alpha transparency many of the results look wrong. Could add a white `fillRect` to blank the background out? Or, it's probably unnecessary to have JPEG results for all of the test cases, so I could just add one row to the bottom with a JPEG result. Preference?

Hunting for memory leaks now, but otherwise I have nothing more to add.

zbjornson:
Don't see any mem leaks after creating 1000 JPEG streams.

zbjornson:
@LinusU ping :) Let me know if there's anything else needed here. After this is merged I'll do #637.

src/JPEGStream.h
```diff
@@ -148,7 +140,7 @@ write_to_jpeg_stream(cairo_surface_t *surface, int bufsize, int quality, bool pr
   }
   free(dst);
   jpeg_finish_compress(&cinfo);
-  jpeg_free_custom_allocations(&cinfo);
+  //jpeg_free_custom_allocations(&cinfo);
```
LinusU: @zbjornson Could you remove this line and I'll merge and release right away :)

zbjornson:
Done, thanks :)

LinusU:
Cool :sunglasses:

Just gonna let the build go thru, shouldn't be long now :+1:

LinusU:
Released as 1.2.10, thanks again for your help! :100:

PR #640, opened 2015-10-13
tests: use standard mocha ui

LinusU:
This switches the `ui` of mocha to the default (`describe` / `it`) instead of exports (`exports['...']`).

I also switched back to the default reporter because I think it's nice and it removed some lines from the makefile. I don't have a strong opinion on this part so happy to change back is someone loves the dot reporter.

Looks good?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,15 +1,11 @@
 
 ADDON = build/Release/canvas.node
-REPORTER = dot
 
 $(ADDON): src/*.cc
 	npm install
 
 test: $(ADDON)
-	@./node_modules/.bin/mocha \
-		--reporter $(REPORTER) \
-		--ui exports \
-		test/*.test.js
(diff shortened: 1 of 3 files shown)


zbjornson:
LGTM, although it doesn't use the sentence-like syntax describe/it is intended to use (`describe("Canvas#getImageData") { it("returns a Uint8ClampedArray with size 4*w*h")...`). Because I don't have make on Windows I've been running a command similar to what you changed the makefile to use here. Bonus points if you change package.json to directly invoke mocha instead of make!

LinusU:
Yeah, it would be nice with better test names, but I figured that this was at least a step in the right direction.

Hmm, using make actually gives us one benefit. It will automatically recompile the module before running the tests if, and only if, the source files have been changed. That's actually really nice when you are developing...

Let's merge this now and fix the other issues elsewhere, I'll open a bug for the make part.

Issue #645, opened 2015-10-14
image onload callback firing twice on visual tests


inssein:
Edit: I made the wrong assumption the first time around..

Just checked out master and was going to try and implement glow and shadow on images, but running into a lot of issues.

First, I am trying to run the visual tests. The server kept crashing because headers were already sent. I commented out every test except for "image shadow", and it still crashed. I quickly debugged server.js and realized that in /render and /jpeg, done() fires twice.

Can anyone else confirm?

inssein:
Not sure if this is new to node v4.2.1, but either way, cleaned it up and fixed the error

PR #646, opened 2015-10-14
add helper function to execute test and only fire callback once

inssein:
Fixes https://github.com/Automattic/node-canvas/issues/645

### DIFF
--- a/test/server.js
+++ b/test/server.js
@@ -37,6 +37,15 @@ function testFn(req){
   return eval('(' + req.body.fn + ')');
 }
 
+function executeTestFn(ctx, fn, done) {
+  if(2 === fn.length) {
+    fn(ctx, done);
+  } else {
+    fn(ctx);
+    done();
+  }
+}
+
 function createCanvas(req, type){
   var width = req.body.width
     , height = req.body.height;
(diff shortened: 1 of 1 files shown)


LinusU:
I don't understand, how does this make sure that callbacks only fire once? Could you please explain your code :)

inssein:
@LinusU I just wrote a simple testbed, and tested it in 4 different node versions. Looks like it never worked, or I am missing something.

```js
var done = function() {
  console.log('done');
};

var fn = function(ctx, done) {
  done();
};

2 == fn.length ? fn('test', done) : fn('test'), done();```

The above code will print ```done``` twice. Once by going through fn(), and once by being called at the end of the last line. This breaks the test for me as /jpeg and /render will try to send the result twice.

LinusU:
Cool, yes I have seen this behaviour as well actually

inssein:
@LinusU what's the process on getting this merged?

LinusU:
Merging now, sorry for the delay

inssein:
@LinusU thanks!

LinusU:
Thank you, help is always appreciated!

PR #648, opened 2015-10-14
implement blur on images (drawImage)

inssein:
Quickly implemented blur on images based on @woodcoder's advice. This PR fixes https://github.com/Automattic/node-canvas/issues/133.

I have done a few things that I do not fully understand yet. In order to replicate the blur, I:
* Divided the shadowBlur by 2 when calling blur()
* Offset the surface by 1.15

@woodcoder: In your implementation of shadow blur, you offset the surface by 1 (https://github.com/woodcoder/node-canvas/blob/8c6984a0999253a22fbd74713d4823da8b2cdd48/src/CanvasRenderingContext2d.cc#L395). Any insight as to why 1? That's what I did initially as well, but it was slightly off.

### DIFF
--- a/src/CanvasRenderingContext2d.cc
+++ b/src/CanvasRenderingContext2d.cc
@@ -895,7 +895,10 @@ NAN_METHOD(Context2d::DrawImage) {
       context->blur(shadow_surface, context->state->shadowBlur / 2);
 
       // paint
-      // @todo figure out where the 1.4 comes from
+      // @note: ShadowBlur looks different in each browser. This implementation matches chrome as close as possible.
+      //        The 1.4 offset comes from visual tests with Chrome. I have read the spec and part of the shadowBlur
+      //        implementation, and its not immediately clear why an offset is necessary, but without it, the result
+      //        in chrome is different.
       cairo_set_source_surface(ctx, shadow_surface,
         dx - sx + context->state->shadowOffsetX - pad + 1.4,
         dx - sx + context->state->shadowOffsetY - pad + 1.4);
--- a/src/CanvasRenderingContext2d.cc
+++ b/src/CanvasRenderingContext2d.cc
@@ -900,6 +900,10 @@ NAN_METHOD(Context2d::DrawImage) {
         dx - sx + context->state->shadowOffsetX - pad + 1.15,
         dx - sx + context->state->shadowOffsetY - pad + 1.15);
       cairo_paint(ctx);
+      
+      // cleanup
+      cairo_destroy(shadow_context);
+      cairo_surface_destroy(shadow_surface);
     } else {
       context->setSourceRGBA(context->state->shadow);
       cairo_mask_surface(ctx, surface,
--- a/src/CanvasRenderingContext2d.cc
+++ b/src/CanvasRenderingContext2d.cc
@@ -883,10 +883,29 @@ NAN_METHOD(Context2d::DrawImage) {
   }
 
   if (context->hasShadow()) {
-    context->setSourceRGBA(context->state->shadow);
-    cairo_mask_surface(ctx, surface,
-      dx - sx + context->state->shadowOffsetX,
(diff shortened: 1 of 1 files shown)


src/CanvasRenderingContext2d.cc
```diff
@@ -883,10 +883,33 @@ NAN_METHOD(Context2d::DrawImage) {
...
+      cairo_surface_t *shadow_surface = cairo_image_surface_create(CAIRO_FORMAT_ARGB32, dw + 2 * pad, dh + 2 * pad);
+      cairo_t *shadow_context = cairo_create(shadow_surface);
+
+      // mask and blur
+      context->setSourceRGBA(shadow_context, context->state->shadow);
+      cairo_mask_surface(shadow_context, surface, pad, pad);
+      context->blur(shadow_surface, context->state->shadowBlur / 2);
+
+      // paint
+      // @todo figure out where the 1.15 comes from
```
LinusU: I would love for this to be solved before merging, where did you find the value?

woodcoder:
The comment on the [commit](https://github.com/woodcoder/node-canvas/commit/2b366c684e16fda05ca55e4da532445efeaaed8d) was "The +1 on the offset seems to match the browser's output better, but I can't work out why it would be needed (unless it's pixel alignment related)."  So I think I just did it by eye, comparing the output against the test canvases viewed in Firefox.

You may find something in the canvas W3C spec?  I seem to remember finding the odd constant here and there when I was looking at the shadowing blurring.  HTH

inssein:
Reading the spec (http://www.w3.org/TR/2015/PR-2dcontext-20150924/#shadows) and haven't found anything about the offset yet.

I did find something to do with the dividing the shadowBlur by 2.
```
If shadowBlur is greater than 0:
* Let σ be half the value of shadowBlur.
* Perform a 2D Gaussian Blur on B, using σ as the standard deviation.
```

If the blur function we are using is a gaussian blur, and the radius is the standard deviation, then we should be following the spec.

inssein:
@LinusU The blur implementation seems to be different in every browser... not sure what the right answer here is, is there one we try to follow?

Also, I still can't figure out what the offset is. It's closer to 1.4 after testing with some larger images, but its no more than an educated guess.

With the offset at 1.4, I can get a 99% (subjective) match with current chrome (and canary).

zbjornson:
For what it's worth, FF's shadow looks different than Chrome's, and furthermore the four different rendering engines for FF look slightly different. Perhaps it's futile to match perfectly. (I would download these and flip through them using some preview application.)
Chrome:
chrome
FF-Skia:
skia
FF-Cairo:
cairo
FF-Direct2D:
direct2d
FF-Direct2D1.1:
direct2d11

inssein:
@zbjornson wow those are different. If you put them in photoshop and toggle layers, the colors of the leaves change.

inssein:
Whats the next step here? Do we pick a rendering engine to match best to?

Also I have found an answer as to why the offset. Maybe someone with more experience in this area could help me out.

Either way, I think what we have now is better than doing nothing. It matches chrome pretty well in all the tests that I have tried.

LinusU:
Could you add a test case to the `test-server` thing? I then believe that we could merge it.

LinusU:
Oh, and also, if you could update the comment to say something along "The 1.4 value was chosen to best match how the shadow looks in Google chrome currently", the more verbose the better.

inssein:
@LinusU there was already a failing test added by @woodcoder previously, its keyed "image shadow". I will add the comment asap.

inssein:
@LinusU not sure why the travis build failed, I only modified a comment.

LinusU:
@inssein I've restarted the build, lets see if it helps

inssein:
@LinusU restart did it, thanks.

LinusU:
No problem, let's merge this!

LinusU:
Released as `1.2.11`

Issue #644, opened 2015-10-14
Try to upgrade to container-based Travis builds


zbjornson:
If we can get this to work, would save a ton of time on builds.

(Happy to work on this sometime. Obviously low-priority.)

LinusU:
Yeah I've been thinking about this as well, if we install the libs locally instead of globally it should work.

LinusU:
Closed by #652

PR #652, opened 2015-10-24
simplify travis config

targos:
Currently, CI takes a long time because it has to compile all dependencies before running the test.
Use Ubuntu packages instead of building them from source.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,13 +12,16 @@ addons:
     sources:
       - ubuntu-toolchain-r-test
     packages:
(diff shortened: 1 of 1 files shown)


targos:
test time is divided by 3 with this change. Now the limiting step is `npm run benchmark`.

zbjornson:
Cool. Will fix #644.

LinusU:
Nice! `sudo: false` is awesome since it lets the builds run on their new Docker container platform.

Merging now :+1:

PR #655, opened 2015-10-28
fix the install script to current repository

outsideris:
Installation script url still pointed LearnBoost that doesn't exist, so I updated it.


LinusU:
Thanks! :shipit:

PR #660, opened 2015-11-04
Wrap std::min calls in paranthesis to prevent macro expansion on windows

### LINKED ISSUE #659: Error compiling CanvasRenderingContext2d.cc using VS2013
There seems to be an issue with std::min conflicting with the min macro defined by windows.h, same issue as [explained here](http://stackoverflow.com/questions/5004858/stdmin-gives-error). I dont have the error output available right now, but let me know if you want to see it.

I'm building using nw-gyp (as opposed to node-gyp) for usage with node-webkit, which may be part of the issue.

I propose we use [this solution](http://stackoverflow.com/a/22023122). I'll send a pull request in a minute.

deadbeef84:
Fixes #659.

### DIFF
--- a/src/CanvasRenderingContext2d.cc
+++ b/src/CanvasRenderingContext2d.cc
@@ -619,6 +619,8 @@ NAN_METHOD(Context2d::PutImageData) {
       if (dx < 0) sw += dx, sx -= dx, dx = 0;
       if (dy < 0) sh += dy, sy -= dy, dy = 0;
       // clamp width at canvas size
+      // Need to wrap std::min calls using parens to prevent macro expansion on
+      // windows. See http://stackoverflow.com/questions/5004858/stdmin-gives-error
       cols = (std::min)(sw, context->canvas()->width - dx);
       rows = (std::min)(sh, context->canvas()->height - dy);
       break;
--- a/src/CanvasRenderingContext2d.cc
+++ b/src/CanvasRenderingContext2d.cc
@@ -592,8 +592,8 @@ NAN_METHOD(Context2d::PutImageData) {
   switch (info.Length()) {
     // imageData, dx, dy
     case 3:
-      cols = std::min(imageData->width(), context->canvas()->width - dx);
-      rows = std::min(imageData->height(), context->canvas()->height - dy);
+      cols = (std::min)(imageData->width(), context->canvas()->width - dx);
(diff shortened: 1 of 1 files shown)


LinusU:
Hmm, that's strange...

@zbjornson Any comments on this? I'm okay with merging it otherwise

zbjornson:
I don't see this issue using node-gyp with either VS2013 or 2015, but also no problem merging. (I tried reproducing with nw-gyp, but couldn't get nw-gyp to run. Wonder if this happens with electron, too.) Curious what's including windows.h or windef.h and causing this problem.

Might add a comment saying why the otherwise unnecessary parens are there so they aren't removed accidentally in the future.

LinusU:
That sounds good, @deadbeef84 could you add a comment to the code?

deadbeef84:
Sure!

Btw, I'm using VS2013 express if that makes any difference.

LinusU:
Is there a reason why we aren't using the `#define NOMINMAX` fix instead?

deadbeef84:
Not really, I just thought parens were less hacky, but I guess they're both a bit hacky :) If you want me to change to `#define NOMINMAX` I'll do that. Perhaps we could even put the define in [binding.gyp](https://github.com/deadbeef84/node-canvas/blob/windows-std-min-fix/binding.gyp#L66), if you prefer that?

LinusU:
I think I would prefer it in the binding.gyp for windows only. I feel like we would never want to use the `min` macro since we now have `std::min` function.

Maybe the binding.gyp can do it conditionally only for VS2013 Express...

Or maybe it's better to have `#define NOMINMAX` before the `#include <windows.h>`, it makes it clear what it's for.

deadbeef84:
Except windows.h is not included directly from node-canvas, but probably by one of the libraries. Which is also why I prefer the parens-solution, assume any of the libraries use the min/max macros in their header-files, then defining NOMINMAX would break them.

deadbeef84:
Any thoughts?

LinusU:
Sorry for the delay, I hate problems like this :)

Could you duplicate the comment to the other callsite as well, then I'll merge :+1:

deadbeef84:
I know what you mean :) Thanks!

LinusU:
Released as `1.3.1`

Issue #668, opened 2015-11-17
Shadow on images is incorrect if the image is scaled


inssein:
Can't believe I didn't see this before, but shadow position is wrong if the image is scaled. Will try fixing this tomorrow, and hopefully include a test as well.

inssein:
Also, the Y position of the shadow and glow is wrong as well, there was a mistake in the PR which I will be rectifying tomorrow.

LinusU:
Nice catch, thanks for the help! I'll be ready to review :+1:

denkan:
Sorry for hi-jacking an old issue.

I'm having trouble with shadow when using scale() on context; shouldn't the shadowBlur value increase accordingly to scale as well? 

My tests seems to be leaving the blur untouched, making the blur/shadow effect barely visible when scaling up. 

Am I doing something wrong or do we have a valid issue here? :)

LinusU:
Seems like a valid issue to me, could you try it out in a web browser?

denkan:
@LinusU, actually didn't try out in browser. 
As far as I can see, the browser doesn't scale the blur value either. 

See fiddle:
https://jsfiddle.net/5aajvm3u/

A bit odd that shadowBlur is the only value that needs manually handling, if you ask me. But node-canvas is doing good (Y) :)

PR #669, opened 2015-11-17
fix incorrect Y offset for shadows, and also fix scaled shadows

inssein:
This PR fixes two things:
* In my initial commit, the shadowOffsetY was incorrect when there was a blur
* Shadow never worked properly for scaled images

I also added a quick visual test.

Fixes https://github.com/Automattic/node-canvas/issues/668


inssein:
I found one more bug! shadowBlur = 1 seems to draw nothing. Looking into it.

LinusU:
Landed as `1.3.2`, thank you!

PR #671, opened 2015-11-19
Use Nan::TypedArrayContents, fix a signed/unsigned comp warning

### LINKED ISSUE #637: Can we use Nan::TypedArrayContents
@zbjornson I think that this can help some of your ifdefs, it will be released with Nan 2.1

nodejs/nan#486

zbjornson:
Fixes #637, plus a compiler warning. Was also able to remove some unnecessary code that creates an ArrayBuffer explicitly in CanvasRenderingContext2d.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "test-server": "node test/server.js"
   },
   "dependencies": {
-    "nan": "^2.0.9"
+    "nan": "^2.1.0"
   },
   "devDependencies": {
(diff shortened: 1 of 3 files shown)


zbjornson:
(5.0 fails as expected)

LinusU:
Tested Node 5.x manually and it works for me, pulling.

Very nice, thanks for the help :100: :+1:

PR #672, opened 2015-11-19
Add Appveyor CI for Windows

zbjornson:
This mostly works. VS2015 fails without #670. We need #458 merged for any to fully pass -- see log:
https://ci.appveyor.com/project/zbjornson/node-canvas/build/job/x3domgfjsq77445m After that I need to change the script to install libjpeg-turbo.

For now there's only one node env. Think it's unnecessary to have more than one, but it's easy to add.

(Sorry I've got 83a8b27 upstream from #671 in here.)

PR #676, opened 2015-11-21
package: remove pre-scripts

LinusU:
See discussion in #672.

@rvagg, @kangax would you be sad to see this go?

ping @zbjornson

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,11 +21,8 @@
   "homepage": "https://github.com/Automattic/node-canvas",
   "repository": "git://github.com/Automattic/node-canvas.git",
   "scripts": {
-    "prebenchmark": "node-gyp build",
     "benchmark": "node benchmarks/run.js",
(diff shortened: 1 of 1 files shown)


zbjornson:
:+1:

(BTW you added the scripts in d4fab578, although they were 1:1-ish replacements for the make targets.)

LinusU:
Yeah, I just added them to do exactly what the make-file did...

targos:
With this change, is it still possible to work on the C++ code and `npm test` without manually recompiling ?

LinusU:
@targos Unfortunately that would no longer work.

The problem is that you don't want that in all instances. I think that "test" should only do testing.

But it would of course be nice to have one command to type when doing development. Some ways to do this would be 1) bash function/alias 2) another npm script, maybe `build-and-test` 3) another npm package that could work with any npm module, it would be installed by the developer with `npm install -g ___`.

PR #679, opened 2015-11-25
suppress duplicate `end` events

vasili-zolotov:
The `end` event only needs to be raised once upon reaching EOF.

### DIFF
--- a/lib/jpegstream.js
+++ b/lib/jpegstream.js
@@ -47,8 +47,10 @@ var JPEGStream = module.exports = function JPEGStream(canvas, options, sync) {
       } else if (chunk) {
         self.emit('data', chunk);
       } else {
-        self.emit('end');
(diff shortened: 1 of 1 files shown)


LinusU:
When do you see multiple `end`-events? Seems like there might be a bug on the C++ side here...

kliu12:
We see multiple `end` events when streaming the following image through `jpegStream()`:

http://placelocal-photos.s3.amazonaws.com/71675-505833-38d2ea51-e46f-c228-9c08-70da44996eca_600.png

zbjornson:
@kliu12 and @vasili-zolotov can you show the code you're using so we can write a test case please?

LinusU:
I think that this was fixed by #740, please feel free to reopen if that isn't the case

PR #682, opened 2015-12-03
Unused variable `Image` in the example.

aichholzer:


LinusU:
Hmm, I kind of like having it there since it shows how to get ahold of `Image` which might not be 100% obvious. Maybe we could append something that uses `Image` to the example?

aichholzer:
Well, `Canvas.Image` returns a function, and it does not equal `new Image` so it might be even more confusing to have it there. I would suggest updating the corresponding section for that effect.

PR #687, opened 2015-12-07
Replace unescpaed \ with / in browser tests

zbjornson:
Oddly, #646 seems to have broken browser tests (presumably on Windows only). I bisected the history back to that commit, although I don't see why it would have broken it.

Without this, the tests appear to not complete -- the [empty onError hooks](https://github.com/Automattic/node-canvas/blob/master/test/public/tests.js#L1536) were suppressing a cairo error ("failed to read from input stream").

### DIFF
--- a/test/server.js
+++ b/test/server.js
@@ -31,7 +31,8 @@ function testFn(req){
   req.body.fn = req.body.fn
     .replace("'state.png'", "'" + __dirname + "/public/state.png'")
     .replace("'face.jpeg'", "'" + __dirname + "/public/face.jpeg'")
-    .replace("'star.png'", "'" + __dirname + "/public/star.png'");
+    .replace("'star.png'", "'" + __dirname + "/public/star.png'")
+    .replace(/\\/g, "/"); // unescaped \ path sep causes issues on Windows, at least
 
   // Do not try this at home :)
   return eval('(' + req.body.fn + ')');


LinusU:
Would it be better to solve this using `path.join` or `path.resolve`?

LinusU:
Also, I don't see the empty onError hooks in your link, did you maybe misslink?

zbjornson:
Oops, misslinked. Edited original post.

zbjornson:
Just played a bit, but `path.resolve` doesn't appear to fix incorrect path seps. `__dirname` seems to have unescaped `\`, which is the problem.

zbjornson:
@LinusU your test cleanup PR reminded me of this. I didn't explain this problem too well: on Windows, `__dirname` is something like `C:\\foo\\bar\\node-canvas`, and after the normalization in server.js the test functions contain paths like `C:\foo\bar\node-canvas/public/star.png` (when printed with `if (/star/.test(req.body.fn)) console.log(req.body.fn);`).

These work:

* `.replace(/\\/g, "/")`
* `.replace(/\\/g, "\\\\")`

These do not:

* `path.join(__dirname, "public", "star.png")`
* `path.posix.join(__dirname, "public", "star.png")`
* `path.win32.join(__dirname, "public", "star.png")`
* `path.resolve(__dirname, "public/star.png")`

which is to say, `path.join` and `resolve` won't touch the separators in `__dirname`.

LinusU:
Ahh, I did not see the `eval` below, that explains it...

I'll take a look at this when I get home from work, sorry for the delay

LinusU:
I fixed this another way in #830

PR #689, opened 2015-12-11
using node-pre-gyp on windows

wsw0108:


LinusU:
`Could not resolve host: maptalks.org`

wsw0108:
For now, I have no public storage to place it.
I will get one and change the remote host.

LinusU:
I think that this needs to be hosted by Automattic in a more official way. This is binary code that will be run on the end users machine after all.

wsw0108:
I agree that. Hope it will be come soon.

LinusU:
I think we still can have this open. If we manage to get everything in place so that it works, we can resolve the hosting issue later.

We must think about all the libraries that we are using though, would it be feasible to bundle cairo, pango, libpng, etc?

Mithgol:
@wsw0108 If you do not have a cloud storage in your avail, for now consider hosting binaries in a content-addressable distributed file system — [IPFS](https://ipfs.io/).

PR #685, opened 2015-12-07
Add data URI jpeg support (fixes #31)

zbjornson:

PR #695, opened 2015-12-28
Allow optional toDataURL arguments to be undefined.

zbjornson:
Per discussion in #685, allow optional arguments to toDataURL to be undefined. (Was a regression from adding image/jpeg support.)

cc @domenic

### DIFF
--- a/History.md
+++ b/History.md
@@ -1,3 +1,7 @@
+future / future
+==================
+  * Allow optional arguments in `toDataURL` to be `undefined` (#695)
+
 1.3.5 / 2015-12-07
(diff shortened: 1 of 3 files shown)


LinusU:
@zbjornson Does #696 replace this one? Or do you want this one merged before?

zbjornson:
#696 would replace this one (it has 0806071 upstream). If you want to merge #696 I can squash those two commits -- let me know.

LinusU:
Let's go for #696! I left a small comment to address as well :)

PR #696, opened 2015-12-28
Make toDataURL more spec-compliant

zbjornson:
I personally don't care if this gets merged or not, but while making #695 I made this, which continues to improve spec compliance. Because the error type and message changes, this is breaking, however.

1. A TypeError is thrown after the first offending argument is processed, per the [IDL overload resolution algorithm](http://heycam.github.io/webidl/#dfn-overload-resolution-algorithm) step 11.21. The messages match those of Chrome. Note that in doing this, I had to remove some of the helpful messages that are specific to node-canvas, e.g. lib/canvas.js#L229 and lib/canvas.js#L231. **Edit**: after further revision, there is now only one TypeError thrown. The toDataURL spec says "Other arguments must be ignored and must not cause the user agent to throw an exception." node-canvas complicates the design a bit by supporting async.
1. Support `toDataURL("image/jpeg", 0.5, fn...)` -- the second argument is the quality from 0 to 1. Note that this is a gotcha because the node-canvas options for JPEG use 0-100. ([spec](http://www.w3.org/TR/html5/scripting-1.html#a-serialization-of-the-bitmap-as-a-file), table)
1. If the canvas has no pixels, the minimal string "data:," is returned ([spec](http://www.w3.org/TR/html5/scripting-1.html#dom-canvas-todataurl), number 2).
1. If a non-supported format is requested, fall through to "image/png" ([spec](http://www.w3.org/TR/html5/scripting-1.html#a-serialization-of-the-bitmap-as-a-file), paragraph 5).
1. `toLowerCase` the type before checking if it's supported ([spec](http://www.w3.org/TR/html5/scripting-1.html#a-serialization-of-the-bitmap-as-a-file), paragraph 6).

This could be split up if desired. Also note that this is downstream of the commit for #695, the minimal PR to just fix the regression to support 'undefined'. Can squash them if this is the preferred resolution.

Apologies that my original PR to add image/jpeg didn't follow these specs.

### DIFF
--- a/History.md
+++ b/History.md
@@ -1,3 +1,8 @@
+future / future
+==================
+
+  * Allow optional arguments in `toDataURL` to be `undefined` and improve `toDataURL`'s spec compliance (#690)
+
 1.3.5 / 2015-12-07
 ==================
 
--- a/lib/canvas.js
+++ b/lib/canvas.js
@@ -18,7 +18,8 @@ var canvas = require('./bindings')
   , JPEGStream = require('./jpegstream')
   , FontFace = canvas.FontFace
   , fs = require('fs')
-  , packageJson = require("../package.json");
+  , packageJson = require("../package.json")
+  , FORMATS = ['image/png', 'image/jpeg'];
 
 /**
  * Export `Canvas` as the module.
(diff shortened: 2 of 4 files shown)


zbjornson:
Addresses/replaces #517 and #515.

LinusU:
I personally don't think that changing an error type and message is enough to bump a major version. Also, one could argue that the correct behaviour have always been the spec and that it's been a bug not to follow it up to now. Therefor this should just be a patch bump since we are addressing a bug.

I would be glad to merge this!

lib/canvas.js
```diff
@@ -186,61 +186,60 @@ Canvas.prototype.createSyncJPEGStream = function(options){
...
-      }
-      break;
-    case 3:
-      if ('image/jpeg' === a1 && 'object' === typeof a2 && 'function' === typeof a3) {
-        type = a1;
+  if ('function' === typeof a1) {
+    fn = a1;
+  } else {
+    if ('string' === typeof a1 && ('image/png' === a1.toLowerCase() ||
+      'image/jpeg' === a1.toLowerCase())) {
```
LinusU: Can we store the supported image formats in an array as a constant at the top of the file? `var FORMATS = ['image/png', 'image/jpeg']` and `FORMATS.contains(a1.toLowerCase())`?

zbjornson:
Yeah, agreed, silly reason to bump major, but wanted to point it out anyway because someone might be catching errors and testing types. :)

History.md
```diff
@@ -1,3 +1,7 @@
+future / future
+==================
+  * Allow optional arguments in `toDataURL` to be `undefined` (#690)
```
LinusU: There should be an empty line between line 2 and 3

lib/canvas.js
zbjornson: Done, but used `indexOf` to preserve support for earlier versions of node.

LinusU:
Awesome work as always!! :+1: 

Trying it out locally now

zbjornson:
Thanks!

Sorry, just force-pushed a capitalization fix in readme.md.

LinusU:
No problem

Does it make sense to break out the `toDataURL` tests into it's own file? They have there own describe and are more than 100 lines of code...

zbjornson:
Up to you. If I make `canvas.toDataUrl.test.js` with `describe("Canvas", ... describe("#toDataURL", ...))` that makes the output look like

```
  Canvas
    √ .version
    √ .cairoVersion
    √ .parseFont()
...

  Canvas
    #toDataURL()
      √ toDataURL() works and defaults to PNG
      √ toDataURL(0.5) works and defaults to PNG
      √ toDataURL(undefined) works and defaults to PNG
...
```

I tend to split test files by class/object I guess.

LinusU:
Yeah, the split per class is very logical. It's just becoming so big of a file :laughing: 

Let's keep it for now, merging time!

LinusU:
Released as `1.3.6`, our first release for 2016 :fireworks:

PR #700, opened 2016-01-06
remove cairo_include.sh

LinusU:
This file was only used for one day in 2013, it's time to let it go...


LinusU:
Superseded by #701

PR #600, opened 2015-07-28
gyp: fix pangocairo detection

LinusU:
*Fixes #599*

This should fix the detection of pangocairo to ask `pkg-config` wether or not it has it, instead of trying to detect it ourselves.

Try it out with the following command:

```sh
npm install LinusU/node-canvas#pangocairo-test
```

PR #701, opened 2016-01-06
cleanup build scripts

LinusU:
Replaces #700, #600 and #575 

Also cleans up the script to fix warnings from [shellcheck](http://www.shellcheck.net).

Test by running:

```sh
npm install LinusU/node-canvas#build-cleanup
```


LinusU:
Note to self: stop using `which` since it breaks on OS X because it outputs stuff to stderr...

targos:
:+1: works perfectly on Fedora 23

LinusU:
I have multiple confirmation + travis that this works, let's merge!

main-qimg-c987e575dca535436a6b811c189e7294

PR #706, opened 2016-01-27
Allow to unbind onload/onerror callback handlers

LinusU:
Replaces #616, I screwed up when I did the merge...


LinusU:
Released as 1.3.9

PR #712, opened 2016-02-07
fix segfault on node 0.10.x

### LINKED ISSUE #688: Segmentation fault only on Linux
Heatmap.js (https://github.com/amytych/node-heatmaps) uses canvas for creating heatmaps. getImageData(..) funciton of the canvas module is reporting segmentation fault, but only in Linux. I tested the code with same data set and it works fine on windows. After some debugging, I narrowed it down to "image = actx.getImageData(left, top, right - left, bottom -  top);" of colorize function of the heatmap.js module:

```javascript
 Heatmap.prototype.colorize = function (x, y) {
  // get the private variables
  var me = this,
    width = me.get('width'),
    radius = me.get('radius'),
    height = me.get('height'),
    actx = me.get('actx'),
    ctx = me.get('ctx'),
    x2 = radius * 3,
    premultiplyAlpha = me.get('premultiplyAlpha'),
    palette = me.get('gradient'),
    opacity = me.get('opacity'),
    bounds = me.get('bounds');
    var left =0, top=0, bottom=0, right=0;
    var  image = {}, imageData = {} , length =0, alpha =0, offset =0 , finalAlpha =0 ;

  if (x != null && y != null) {
    if (x + x2 > width) {
      x = width - x2;
    }
    if (x < 0) {
      x = 0;
    }
    if (y < 0) {
      y = 0;
    }
    if (y + x2 > height) {
      y = height - x2;
    }
    left = x;
    top = y;
    right = x + x2;
    bottom = y + x2;

  } else {
    if (bounds.l < 0) {
      left = 0;
    } else {
      left = bounds.l;
    }
    if (bounds.r > width) {
      right = width;
    } else {
      right = bounds.r;
    }
    if (bounds.t < 0) {
      top = 0;
    } else {
      top = bounds.t;
    }
    if (bounds.b > height) {
      bottom = height;
    } else {
      bottom = bounds.b;
    }
  }

  image = actx.getImageData(left, top, right - left, bottom -  top);
  imageData = image.data;
  length = imageData.length;

  // loop thru the area
  for (var i = 3; i < length; i += 4) {

    // [0] -> r, [1] -> g, [2] -> b, [3] -> alpha
    alpha = imageData[i];
    offset = alpha * 4;

    if (!offset) {
      continue;
    }

    // we ve started with i=3
    // set the new r, g and b values
    finalAlpha = (alpha < opacity) ? alpha : opacity;
    imageData[i - 3] = palette[offset];
    imageData[i - 2] = palette[offset + 1];
    imageData[i - 1] = palette[offset + 2];

    if (premultiplyAlpha) {
      // To fix browsers that premultiply incorrectly, we'll pass in a value scaled
      // appropriately so when the multiplication happens the correct value will result.
      imageData[i - 3] /= 255 / finalAlpha;
      imageData[i - 2] /= 255 / finalAlpha;
      imageData[i - 1] /= 255 / finalAlpha;
    }

    // we want the heatmap to have a gradient from transparent to the colors
    // as long as alpha is lower than the defined opacity (maximum), we'll use the alpha value
    imageData[i] = finalAlpha;
  }
  // the rgb data manipulation didn't affect the ImageData object(defined on the top)
  // after the manipulation process we have to set the manipulated data to the ImageData object
  image.data = imageData;
  ctx.putImageData(image, left, top);
};

LinusU:
A segfault occurred in `getImageData(...)` on Node.js version 0.10.x because of an incorrect length specified to a function call. V8 then thought that the array was longer than it was and tried to access memory outside of it.

This fixes #688 

I think that the reason this haven't been caught in CI is because it doesn't always happen. My guess is that sometimes the memory after the array is owned by V8 and no segfault occurs. Since the `Uint8ClampedArray` constructor ignores any arguments after the length it probably wasn't a problem to pass in more arguments...

### DIFF
--- a/src/CanvasRenderingContext2d.cc
+++ b/src/CanvasRenderingContext2d.cc
@@ -731,13 +731,13 @@ NAN_METHOD(Context2d::GetImageData) {
   int dstStride = sw * 4;
 
   uint8_t *src = canvas->data();
-  
+
 #if NODE_MAJOR_VERSION == 0 && NODE_MINOR_VERSION <= 10
   Local<Object> global = Context::GetCurrent()->Global();
 
   Local<Int32> sizeHandle = Nan::New(size);
   Local<Value> caargv[] = { sizeHandle };
-  Local<Object> clampedArray = global->Get(Nan::New("Uint8ClampedArray").ToLocalChecked()).As<Function>()->NewInstance(3, caargv);
+  Local<Object> clampedArray = global->Get(Nan::New("Uint8ClampedArray").ToLocalChecked()).As<Function>()->NewInstance(1, caargv);
 #else
   Local<ArrayBuffer> buffer = ArrayBuffer::New(Isolate::GetCurrent(), size);
   Local<Uint8ClampedArray> clampedArray = Uint8ClampedArray::New(buffer, 0, size);
@@ -745,7 +745,7 @@ NAN_METHOD(Context2d::GetImageData) {
 
   Nan::TypedArrayContents<uint8_t> typedArrayContents(clampedArray);
   uint8_t* dst = *typedArrayContents;
-  
+
   // Normalize data (argb -> rgba)
   for (int y = 0; y < sh; ++y) {
     uint32_t *row = (uint32_t *)(src + srcStride * (y + sy));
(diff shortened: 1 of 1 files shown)


LinusU:
Released on npm as `1.3.10`

Issue #694, opened 2015-12-22
is it possible to have right-to-left text drawing capabilities?


tomerb15:
I was wondering if it is possible to have right-to-left text drawing capabilities?

chearon:
I think this would be possible when we get #715 merged in, I'll make sure to test it

chearon:
Closed via #715

chearon:
Actually need to confirm this. Pango should be detecting automatically but we could probably also add the [direction](https://html.spec.whatwg.org/multipage/canvas.html#text-preparation-algorithm) option to the context.

chearon:
Fixed by #2482

PR #578, opened 2015-07-07
Enable freetype for windows

ldhertert:
Enable freetype flag for windows and set the necessary compiler flags.  Fixes #547 Canvas.Font on Windows.

While this "works on my machine", I'd recommend another couple people test it out prior to merging this.

Issue #547, opened 2015-04-14
What's the story with Canvas.Font on Windows?


ghost:
Hi,

I am trying get support for Canvas.Font on Windows, but I don't get far. From what I see, currently both with_pango and with_freetype are disabled so the Canvas object does not even have a Font property.
When I try to compile with either one of them, then I get into a myriad of dependency issues with sources of various libs - glib, cairo-ft, etc...

Do you perhaps have an easy setup of compiling with either pango or freetype on windows?

chearon:
Closed via #715

PR #715, opened 2016-02-11
Require Pango over "toy" Cairo font methods. Redo the custom font API

chearon:
# Custom font problems

I have spent a *lot* of time on the custom fonts issue plaguing this project. They work for the regular build, but it does not work for the Pango build because of 3 issues:

* Pango uses the system's font resolver, but there is only code to add custom fonts to FontConfig. There are no win32 calls or calls to CoreText so it could not possibly ever work on OS X or Windows. The following 2 points assume you're on Linux/FontConfig
* `CanvasRenderingContext2d#addFont` breaks use of custom fonts *unless you put something other than family name* as the first argument. How wrong!
* The family name given to `ctx.font` needs to be the real family name embedded in the font, not what you pass to `new Canvas.Font`

**TL;DR: the Pango and non-Pango builds work completely differently and have incompatible APIs**

# Pango

I propose that we require Pango. I think there is a strong case for it:

* The Cairo font methods currently in use [are not supposed to be used seriously](http://cairographics.org/FAQ/#using_pango).
* Wherever Cairo is available, Pango is going to be available
* It's more powerful, so going forward we'll be able to match the HTML5 canvas spec more closely

**TL;DR: Pango is better for fonts than Cairo** 

# New custom font API

There will have to be a new API which is `Canvas.registerFont(ttf)` (equivalent to what people are currently doing, `new Canvas.Font('doesnt do anything', 'font.ttf');`). It's implemented in Node C++, and will delegate to either FontConfig, CoreText, or WinGDI. Note that the current font API isn't even documented so this isn't the end of the world

# Issues this PR addresses
I believe these can all be closed because they become obsolete or because they are fixed: #624, #570, #547, #578, #628, #543, #525, #733, #694, #783, #266


examples/font.js
```diff
@@ -4,42 +4,32 @@
...
   , fs = require('fs')
   , path = require("path");
 
-if (!Font) {
-	throw new Error('Need to compile with font support');
-}
-
 function fontFile(name) {
   return path.join(__dirname, '/pfennigFont/', name);
 }
```
jakeg: Suggest adding a comment here to explain that the file name itself is irrelevant and that the font is found based on the name of the font in the ttf file itself.

jakeg:
**Yes, yes. 100x yes. Massive thank you for working on the code for this :D**

This is the right decision. It's either this or remove Pango support, and this is clearly the correct option.

I'll try to look over the code in a bit myself.

One question: does this (and how does this) cope with multiple variants (faces) of a font? e.g. different weights and italics? Currently addFace(ttf, weight, italic) is used for that (see https://github.com/Automattic/node-canvas/pull/628#issuecomment-181570008), what would the new usage be?

... actually, I just looked over your changes to font.js and see how you intend to do that. Looks great!

    Canvas.registerFont(fontFile('PfennigBold.ttf'));
    ctx.font = 'bold 50px Pfennig';

I presume the word 'Pfennig' is the name of the font within the font file and the file name itself is completely irrelevant? Are you certain that in ``new Canvas.Font('doesnt do anything', 'font.ttf')`` previously the first argument did indeed do nothing? And that in your new usage ``registerFont(ttf)`` will pick up the correct variant/face, including if weights are given as numbers (400, 700 etc) instead of as simply 'bold', 'normal' etc? So the following should also work:

    Canvas.registerFont(fontFile('BlahBlah:n7.ttf')); // same font file just renamed
    ctx.font = '700 50px Pfennig'; // Pfennig' still works as that's the name within the font file. 700 is alias for bold weight

Once again, huge thank you @chearon !!!

examples/font.js
```diff
@@ -4,42 +4,32 @@
...
 
-var pfennigFont = new Font('pfennigFont', fontFile('Pfennig.ttf'));
-pfennigFont.addFace(fontFile('PfennigBold.ttf'),   'bold');
-pfennigFont.addFace(fontFile('PfennigItalic.ttf'), 'normal', 'italic');
-pfennigFont.addFace(fontFile('PfennigBoldItalic.ttf'), 'bold', 'italic');
-
-var canvas = new Canvas(320, 320)
-var ctx = canvas.getContext('2d')
+Canvas.registerFont(fontFile('Pfennig.ttf'));
```
jakeg: Suggest adding a comment here to explain that the file name itself is irrelevant and that the font is found based on the name of the font in the ttf file itself.

jakeg: Suggest adding a comment here to explain that the file name itself is irrelevant and that the font is found based on the name of the font in the ttf file itself.

chearon:
> does this (and how does this) cope with multiple variants (faces) of a font

We no longer have any control over that. It's up to the font host layer of the operating system to match the style and weights we give to Pango (and Pango gives to the font layer) with the correct font. It sounds like a bad thing at first but actually that matches the behavior of non-custom fonts, so I think it's better!

>I presume the word 'Pfennig' is the name of the font within the font file and the file name itself is completely irrelevant? 

Exactly. 

>Are you certain that in new Canvas.Font('doesnt do anything', 'font.ttf') previously the first argument did indeed do nothing?

I'm certain - if you look at the old code, `addFont` was implemented in JS. It would add the font to an internal list. When the setter for `ctx.font` was called later, if it had a matching font that was passed to `addFont`, it would call `ctx._setFace` (implemented in C++) instead of `ctx._setFont` (also C++). Only problem was, the former never had a Pango implementation! So you would end up with nothing happening. #543 attempted I think would have fixed it but it didn't quite get it

> And that in your new usage registerFont(ttf) will pick up the correct variant/face, including if weights are given as numbers (400, 700 etc) instead of as simply 'bold', 'normal' etc? So the following should also work

I believe so, if the font is formatted correctly - would you be able to try that for me?

examples/font.js
chearon: done in d2d6324 👍

chearon:
I just got `Canvas.addFont` working in between `new Canvas` calls across all platforms! @LinusU is there any chance of getting this accepted? I know it's a pain because it breaks backwards compatibility, but I believe in this enough to maintain my own fork if I must

LinusU:
I absolutely want to get this merged, just give me some time to look thru the code. Hopefully I'll get some time this weekend. Thank you for your hard work!

chearon:
Awesome, glad to help!

LinusU:
This looks very awesome, but there is one thing that I don't like:

> When you set `ctx.font`, refer to the styles and the family name as it is embedded in the TTF. If you aren't sure, open the font in FontForge and visit Element -> Font Information and copy the Family Name

I don't have FontForge installed, furthermore, it seems very fragile to rely on the name inside the TTF to be used as an identifier inside the canvas api.

I would like us to be as close to the web canvas as possible. So starting at the [`@font-face` specification](https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face), what do you think of an api like this?

```javascript
// 1
Canvas.registerFontFace({
  fontFamily: 'Awesome Font',
  src: 'fonts/Pfennig.ttf'
})

// 2
Canvas.registerFontFace({
  fontFamily: 'Awesome Font',
  src: 'fonts/PfennigBold.ttf',
  fontWeight: 'bold'
})

ctx.font = '50px "Awesome Font"' // 1
ctx.font = 'bold 50px "Awesome Font"' // 2
```

src/ImageData.h
```diff
@@ -10,7 +10,7 @@
 
 #include "Canvas.h"
 #include <stdlib.h>
-#include "v8.h"
+#include <v8.h>
```
LinusU: Good catch

src/register_font.cc
```diff
@@ -0,0 +1,34 @@
...
+  #elif defined(_WIN32)
+  success = AddFontResourceEx((LPCSTR)filepath, FR_PRIVATE, 0) != 0;
+  #else
+  success = FcConfigAppFontAddFile(FcConfigGetCurrent(), (FcChar8 *)(filepath));
+  #endif
+  
+  // Tell Pango to throw away the current FontMap and create a new one. This
+  // has the effect of registering the new font in Pango by re-looking up all
+  // font families.
+  if (success) pango_cairo_font_map_set_default(NULL);
```
LinusU: The flow in this function seems a bit "off". What do you think of something like this?

```c
bool register_font(unsigned char *filepath) {
  bool success;

  // the ifdefs, set success

  if (!success) return false

  // Tell Pango to throw away the...
  pango_cairo_font_map_set_default(NULL);

  return true;
}
```

chearon: You got it. I just sometimes try to avoid a `return` that's not at the end

chearon:
> This looks very awesome, but there is one thing that I don't like

See my comment to @jakeg, it could actually be better, however the main reason is that we are now offloading font resolution to the OS, so that's impossible. Unless I were to parse the font file with FT, modify the family name, then send it to the OS as a font buffer instead of a file. Hmm, that sounds fun. I could take a crack at it if I get time tomorrow.

chearon:
Been reading a lot about fonts again. I think I would have to add a third party library like [opentype.js](https://github.com/nodebox/opentype.js) (quite an impressive library) to parse and write the font to match the user's font spec so that the OS picks it up correctly. I think it would also have to write a new OTF/TTF/WOFF to a temporary directory, which makes me very sad. Does that seem OK?

terrancesnyder:
+1

LinusU:
Hmm, I see the problem, it would be very interesting to see how Firefox made it work since, I believe, that they have a backend for Cairo in Moz2D...

Maybe this is good as it is, I would love some opinions from other maintainers.

This will have to be released under a new major bump since we now require pango.

chearon:
It would be great if we matched web fonts as closely as possible, it's just going to be hard.

I looked at FX but I can't remember what I found. WebKit holds its own native font handles and does its own selection before it defers to the OS for non webfonts. Which is exactly what we should do since CSS3 specifies the font selection algorithm.

It would involve creating our own PangoFontMap though, which isn't currently possible. Luckily I got in contact with the Pango maintainer so we can get some solid feedback, he's on vacation right now tho

LinusU:
> Luckily I got in contact with the Pango maintainer so we can get some solid feedback, he's on vacation right now tho

Mad props, awesome :raised_hands: 

Thank you for taking your time on this, it's highly appreciated!

veltman:
I've been struggling with this as well. I'm currently using a bunch of OS-specific patch code to harmonize font loading between OS X and Ubuntu, so thanks @chearon for all the work on it.

FWIW I think requiring the font name used to match the family name internal to the font file solves this problem pretty well and is consistent with how it would handle pre-existing fonts. I don't think failing to match the web spec perfectly is a killer, especially if doing so would require writing out new temporary font files to alias them.

chearon:
NP, I don't think the PangoFontMap thing I was going to try will work since all over the code, Pango expects the system to have all of the fonts.

But I was thinking that I could let the user specify the family name, weight etc, another way: just maintain a mapping between the user-set name and style, and what is actually in the file. Then I could send the right attributes to Pango when user-specified family names are sent to `ctx.font`.

There would be problems if a user's font shared a family name with the OS, and it would add an external dependency to be able to read the font file. Is that all still better than having no ability to specify family name and style?

LinusU:
@chearon I kind of like that approach actually :+1:, I don't think that the conflict part would be an actual problem in most cases...

veltman:
Hmmm... it seems like someone trying to add a font with a family name that already exists on the OS would be a somewhat common mistake, moreso than someone needing to add a custom font and being unable to use its actual family name, don't you think?

If there is a descriptive error message that explains the conflict, that seems OK, but if it just fails silently I could imagine that confusing a lot of people who don't have a perfect grasp of the font internals (basically everyone).

chearon:
Well, the example in this repo uses a renamed font, and it's really common for sites like FontSpring to rename the font in @font-face css that they bundle. I think it's mostly about matching CSS?

I dunno about it being a common mistake though, can't imagine why someone would be registering Helvetica or Arial

Also, registered fonts *might* take precedence over system fonts, hopefully they do on all systems. I'll check tonight

veltman:
Sure someone probably wouldn't register Arial, but isn't it easy to imagine a situation where several developers are working on the same codebase, and some, but not all of them, already have a particular font on their local machines, so some of them would experience a conflict if the font was being manually added from an included font file?

chearon:
Yeah that's fair, I realized that after my comment. I did get around to testing it, though, and neither Linux or OS X throw an error if you register a duplicate font family name. And even better than that: it chooses the one you registered over the OS font. So there's nothing to worry about. I don't know about Windows, though.

I should have the PangoFontDescription thing done very soon

veltman:
Hooray! Thank you again for working on this problem.

chearon:
> Hooray! Thank you again for working on this problem.

I'm glad I can help! Also, other than some needed testing and review, I think this is now done! If built with FreeType, `Canvas.registerFont` takes a second argument, an object with optional properties `weight`, `family`, and `style` which use CSS values (should add `stretch` in the future). Font selection is probably [very similar](https://github.com/chearon/node-canvas/blob/prefer-pango/src/Canvas.cc#L635) to the CSS3 spec, though it is done by Pango.

LinusU:
Looks good, going to give it a proper review as soon as I get some time. Could you update the documentation to show how to use the second argument?

LinusU:
Also, is freetype still used? Is it optional? What is it used for? and what differences is there between having it installed or not?

chearon:
Oh yeah, I can do that. Originally I took out FreeType since it was only needed for Cairo fonts. But since we're allowing users to describe the font, I needed a way to look at the font's metadata to get an accurate PangoFontDescription that will actually select the font. It is optional, but without it you cannot give custom names to fonts (the 2nd argument to `Canvas.registerFont` would have no meaning).

Another thing I should note: if people choose to describe the font, the old description will still work. In [the example](https://github.com/chearon/node-canvas/blob/prefer-pango/examples/font.js), either `ctx.font = 'Pfennig'` or `ctx.font = 'pfennigFont'` would work.

LinusU:
Hmm, I think that we should require FreeType then as well. Is there any downside to this?

chearon:
I guess we should yeah, I just thought since I can easily make it optional that I would. Also, just realized I need to re-expose the Freetype version. Also also now that I'm at work and have access to Windows, font renaming doesn't seem to be working right so I'll have to look into that

chearon:
Sorry @LinusU I just noticed the second part of your comment. The CSS3 spec for `@font-face` requires the SRC and a family name. weight/style are set to 'normal' regardless of what is in the file. Maybe we require the second argument, an object that must specify at least family? I don't have a strong opinion either way

LinusU:
I think that we should align ourselves with the spec and require at least the family name to be set :+1:

chearon:
So this is working great on Linux right now but there are some issues with Pango's OS X-specific code that was making the "get a font description that will definitely select this TTF" code not work well for a certain class of font weights.

I [just submitted one patch](https://bugzilla.gnome.org/show_bug.cgi?id=766148) today to Pango that makes it work a lot better. I think I need to submit one more to make it comparable to Linux, but regardless it may take a while to get upstream. So I think I've done everything I can here, aside from maybe adding some tests or something

christoph-buente:
Works like a charm with the provided font.js example on AWS Linux and Mac OS X.

veltman:
It looks like that Pango patch was accepted - is there any other blocker preventing this from getting merged?

veltman:
I recently tested the existing font behavior with the following scenarios on OS X and Ubuntu 14.04:

```js
var fs = require("fs")
    path = require("path")
    Canvas = require("canvas");

var canvas = new Canvas(320, 70),
    ctx = canvas.getContext("2d");

// Font name: SourceSansPro
// Filename: SourceSansPro
// Metadata font name: SourceSansPro
testFont("SourceSansPro", "SourceSansPro.otf", 1);

// Font name: SourceSansPro
// Filename: XYZ
// Metadata font name: SourceSansPro
testFont("SourceSansPro", "XYZ.otf", 2);

// Font name: XYZ
// Filename: SourceSansPro
// Metadata font name: SourceSansPro
testFont("XYZ", "SourceSansPro.otf", 3);

// Font name: SourceSansProX
// Filename: SourceSansProX
// Metadata font name: XYZ
testFont("SourceSansProX", "SourceSansProX.otf", 4);

function testFont(name, filename, i) {

  var ssp = new Canvas.Font(name, path.join(__dirname, filename));

  ctx.addFont(ssp);
  ctx.font = '50px ' + name + '';
  ctx.fillText("Quo Vaids?", 0, 70);

  canvas.createPNGStream().pipe(fs.createWriteStream(path.join(__dirname, "test-" + i + ".png")))

}
```

In all cases, the custom font failed with Pango but succeeded if I built without it. This was true whether or not I added other faces.  Under what circumstances _should_ it work with Pango enabled?  It seems like the metadata name and the supplied name matching should be enough, but even that doesn't seem to work.

@LinusU, is there anything we can do to help vet this PR further? Is it going to be delayed until 2.0?

okcoker:
@veltman did you also try replacing 

`ctx.font = '50px ' + name + '';`

 with 

`ctx.font = '50px "' + name + '"';`

On my machine (OSX), those double quotes, around the font name inside the single quotes, made a difference for me. Outside of OSX land so far, I tried using a fork with this PR on heroku using the https://github.com/mojodna/heroku-buildpack-cairo buildpack and ran into the error [here](https://github.com/Automattic/node-canvas/pull/715/files#diff-c76c7ea0c7b77ce0bd7e43541bf43e9dR490). Haven't really dug too far into that one though.

chearon:
It's probably because you are using OTF. The file path gets sent directly to an OS specific "register font" call, so it depends on the OS. I never tried anything other than TTF but underneath all of the major font formats are SFNT fonts and it's disappointing that all formats don't work

okcoker:
@chearon You'd be right. I've only tried one font and it did in fact happen to be otf. Will try with others soon.

LinusU:
> Is it going to be delayed until 2.0?

This will need to be released as 2.0 since we can't have breaking changes without a major bump. But there isn't much holding back the 2.0 release, it's not about features at all but just about getting some breaking changes in.

veltman:
@okcoker @chearon I'm not sure it's OTF's problem. See: https://github.com/veltman/canvas-font-test

If I run this test:

```js
var ssp = new Canvas.Font(name, path.join(__dirname, filename));

ctx.addFont(ssp);

// Weight
ctx.font = '400 50px ' + name + '';
ctx.fillText("Quo Vaids?", 0, 70);

// Weight + quotes
ctx.font = '400 50px "' + name + '"';
ctx.fillText("Quo Vaids?", 0, 140);

// No weight
ctx.font = '50px ' + name + '';
ctx.fillText("Quo Vaids?", 0, 210);

// No weight + quotes
ctx.font = '50px "' + name + '"';
ctx.fillText("Quo Vaids?", 0, 280);
```

on every conceivable combo of naming issues:

```sh
#!/bin/bash

# everything matches
node font-test.js "SourceSansPro" "otf/SourceSansPro.otf" > otf-matching.png
node font-test.js "SourceSansPro" "ttf/SourceSansPro.ttf" > ttf-matching.png

# filename is different, but font name and metadata in font file match
node font-test.js "SourceSansPro" "otf/XYZ.otf" > otf-different-filename.png
node font-test.js "SourceSansPro" "ttf/XYZ.ttf" > ttf-different-filename.png

# font name is different
node font-test.js "XYZ" "otf/SourceSansPro.otf" > otf-different-name.png
node font-test.js "XYZ" "ttf/SourceSansPro.ttf" > ttf-different-name.png

# metadata in font file is different
node font-test.js "SourceSansProX" "otf/SourceSansProX.otf" > otf-different-meta.png
node font-test.js "SourceSansProX" "ttf/SourceSansProX.ttf" > ttf-different-meta.png
```

It still all fails with Pango, and all succeeds if I rebuild with Pango disabled... is there something I'm missing?

LinusU:
@veltman I think it's broken currently, but if you try out this specific branch (which has another api) it should hopefully work.

okcoker:
@chearon I tried with a couple ttf fonts on a heroku linux box with no luck. Also looking at the code I noticed you mentioned trying a bunch of google fonts so i tried with Oswald and also had no luck. It would be a little tough for me to reproduce locally but I'll try if nobody has any ideas. 

I'm just using a free heroku cedar-14 dyno with the [cairo#pango](https://github.com/mojodna/heroku-buildpack-cairo.git#pango) buildpack as mentioned [here](https://github.com/mojodna/heroku-buildpack-cairo/issues/6#issuecomment-94102139).

chearon:
@okcoker Is it giving you the "could not register font with OS" error? Or is it printing the wrong font?

okcoker:
@chearon The same "could not register the font" as mentioned earlier. I just downloaded a vagrant box to hopefully mimic heroku so i can start debugging. If you have any time, you'll probably find it before me as I've never written C, so any suggestions/thoughts would be helpful

okcoker:
Ok so I started experimenting with the font that failed on the heroku box and was successful in using @veltman's script with a few edits within a similar vagrant environment. It even worked with an otf font.

Edited script:
```js
// Run with:
// $ node font-test.js "Whatever font name cuz not important" /path/to/font.ttf
const fs = require('fs');
const path = require('path');
const Canvas = require('canvas');

const out = fs.createWriteStream(__dirname + '/out.png');

const name = process.argv[2];
const filename = process.argv[3];

const canvas = new Canvas(320, 280);

Canvas.registerFont(path.join(__dirname, filename), {
    family: name
});

// Weight
ctx.font = '400 50px ' + name + '';
ctx.fillText('Quo Vaids?', 0, 70);

// Weight + quotes
ctx.font = '400 50px "' + name + '"';
ctx.fillText('Quo Vaids?', 0, 140);

// No weight
ctx.font = '50px ' + name + '';
ctx.fillText('Quo Vaids?', 0, 210);

// No weight + quotes
ctx.font = '50px "' + name + '"';
ctx.fillText('Quo Vaids?', 0, 280);


const stream = canvas.pngStream();
stream.on('data', function(chunk) {
    out.write(chunk);
});

stream.on('end', function() {
    console.log('saved png');
});

```
As mentioned before you need those double quotes or the font doesn't register correctly.
Screenshot:
out2

My only thoughts now would be differences in system packages are causing the error on heroku? Within vagrant I just installed the packages via `apt-get` as recommended in the readme where as heroku uses buildpacks. Will keep searching.

chearon:
@okcoker yes probably, unfortunately I don't know anything about Heroku. Is it just Ubuntu? If so do you get full access to the OS? It would be failing on the call to [FcConfigAppFontAddFile](https://www.freedesktop.org/software/fontconfig/fontconfig-devel/fcconfigappfontaddfile.html). I can't think of why it would fail other than maybe the process doesn't have permissions to the file

The quotes should only be necessary if the font family has spaces in it, is that true?

veltman:
@chearon question about this new API: it seems like if you `registerFont` on the same family + style + weight twice, you'll get an error. Is there some way to test whether a font is already registered, so that if, e.g., you have a script that might be doing various Canvas things with different configs/fonts, it doesn't kill itself by double-registering a font?

okcoker:
@chearon Yea just ubuntu. I don't have any access really. I only can see logs from stdout/stderr. Is there an easy way I can know if that's the specific problem?

Also you're right about the quotes. I should have known because that's the way fonts have worked forever :S

okcoker:
Ok so I made a [simple repo](https://github.com/okcoker/node-canvas-font-test) and [heroku](https://node-canvas-fonts.herokuapp.com/) app to demonstrate the problem. Works locally on OSX using otf fonts but I get the registerFont error on the heroku app. I don't even know where to begin debugging whether or not is is related to `FcConfigAppFontAddFile` but the more I look into it, the more I believe it might not be related to anything with node canvas.

I might just get in touch with someone at heroku and follow up here.

chearon:
It is definitely the FC call that fails, because the return value of that is when I throw the exception - see [here](https://github.com/chearon/node-canvas/blob/prefer-pango/src/Canvas.cc#L489) and [here](https://github.com/chearon/node-canvas/blob/prefer-pango/src/register_font.cc#L222). It would throw that error if the file doesn't exist or is not readable because permissions. I could probably check for that and add a second exception

Are you absolutely sure that the path to the TTF is right? I looked at your repo and it does look right but maybe add an `fs.stat` or something right before the `Canvas.registerFont` just in case

@veltman yeah, OS X throws an error if you register the same file twice. I think the other OSes might not. I could keep track of the full file path and if they register it twice, I would use the latest description they give. I'll look into it!

In other news, @FZFalzar pointed out on #761 that specifying multiple custom families (`ctx.font = '12px CustomA, CustomB';` was not supported - I just support for that so now you can specify a comma separated list of families and Pango will use the later ones if glyphs don't exist in the first few. So you can pick fallback fonts for different languages, wohoo! It's on a separate branch ATM but I'll add it to this PR soon

chearon:
I should add that the comma-separated families doesn't work on OS X yet because Pango just doesn't have that implemented on OS X. I [submitted another patch](https://bugzilla.gnome.org/show_bug.cgi?id=762873#c1) to them which fixes it and hopefully they merge it in, they did accept my last one for OS X weights

okcoker:
> Are you absolutely sure that the path to the TTF is right? I looked at your repo and it does look right but maybe add an fs.stat or something right before the Canvas.registerFont just in case

[Done](https://github.com/okcoker/node-canvas-font-test/commit/699f1030a46464b98d68b47f168ecd7fd4a10295)

> I could keep track of the full file path and if they register it twice, I would use the latest description they give. I'll look into it!

+1

okcoker:
Ok so from heroku support:

> If I then run node in the dyno's bash session and paste in the above I get:
```
Error: Could not load font to the system's font host
    at Error (native)
    at repl:1:-54
    at REPLServer.defaultEval (repl.js:272:27)
    at bound (domain.js:280:14)
    at REPLServer.runBound [as eval] (domain.js:293:12)
    at REPLServer.<anonymous> (repl.js:441:10)
    at emitOne (events.js:101:20)
    at REPLServer.emit (events.js:188:7)
    at REPLServer.Interface._onLine (readline.js:219:10)
    at REPLServer.Interface._line (readline.js:561:8)
```
Which at least has the proper stack trace - the interesting line here is at Error (native) which confirms that the issue is in the extension code. If you can get the underlying library to bubble up the actual error message, rather than just checking for false, you might be able to display it here.

> The other thing that might be useful to know is that the versions of the default packages we install on the dyno are listed here: https://devcenter.heroku.com/articles/cedar-ubuntu-packages

I tried to compile [some of the C code](https://gist.github.com/okcoker/1c9335f5659d36beca8506258ae5f9c9) on the heroku box to see what exactly was failing and I think I've narrowed it down to one of these method calls, [FcConfigSetFonts](https://cgit.freedesktop.org/fontconfig/tree/src/fccfg.c?id=2.11.0#n2150) or [FcFileScanConfig](FcFileScanConfig) simply by commenting out those certain method calls. Not sure if this helps anything but I'm starting to fall short of ideas. That was the first C code I've ever written/copied/compiled.

I can give you access to the box if you're interested.

chearon:
Nice work, thanks for debugging this! I think it would be `FcFileScanConfig` since `FcConfigSetFonts` doesn't seem to impact the return value of `FcConfigAppFontAddFile`. I'll try to debug it if you give me access, have not used Heroku though

I did notice that currently I'm not resolving relative paths correctly so there are now 2 things that can cause the "could not register font with os" exception:

* The font is already registered, or
* The path is relative, has `~`, etc

I'll get to work on those now, but there is no good way for me to handle registering fonts that were already on the OS

brussell98:
When `Canvas.registerFont()` is used but the font is already registered this error is displayed and the process exits
```
(node.exe:26496): Pango-CRITICAL **: pango_win32_font_map_get_font_cache: assertion `font_map != NULL' failed **
Pango:ERROR:pangowin32.c:831:pango_win32_font_finalize: assertion failed: (win32font->fontmap != NULL)

This application has requested the Runtime to terminate it in an unusual way.
Please contact the application's support team for more information.```

**Comment by chearon** [comment]
Hmmm, I tried reproducing that but couldn't - can you post the code that you used to cause that?

**Comment by brussell98** [comment]
I used
```js
const Canvas = require('canvas'),
	Image = Canvas.Image;

if (process.platform.startsWith('win')) {
	Canvas.registerFont(__dirname + '/../../special/weather/fonts/Roboto.ttf', {family: 'Roboto'});
} else {
	const Font = Canvas.Font,
		Roboto = new Font("Roboto", __dirname + "/../../special/weather/fonts/Roboto.ttf"); // eslint-disable-line no-unused-vars
}
```
On Windows 10 64bit

chearon:
I still wasn't able to reproduce that issue @brussell98 but I just added a commit that fixes 2 issues that have been mentioned above (which might solve your issue):

1. you can now double register the same font without getting any errors. it will simply update the description of the font you already registered accordingly

2. relative paths and symbolic links are now handled correctly

FGRibreau:
Any updates on this?

LinusU:
This is where we want to go so let's merge this and start testing it out. We can then iterate further in another PR if necessary 🚀

LinusU:
Also, a huge thank you @chearon for putting in the work on this, really great 👍

PR #717, opened 2016-02-15
require new in constructor

### LINKED ISSUE #294: Forgetting 'new' keyword causes node to crash
The following code causes my node repl to crash. While it's up to the developer to remember the 'new' keyword, it would be nice if the library threw a more intelligible error.

```
var cvs = Canvas(100, 200);
Assertion failed: (handle->InternalFieldCount() > 0), function Wrap, file ../src/node_object_wrap.h, line 71.
Abort trap: 6
```

LinusU:
Before this segfaulted :sob: 

The error message is exactly the same as v8 throws when you try to call a constructor...

Fixes #294


LinusU:
released as 1.3.12

PR #718, opened 2016-02-15
expose freetype version

### LINKED ISSUE #299: A way to determine if canvas was compiled with necessary features at runtime
I would request a way to test if canvas was compiled necessary features at runtime. Something like `canvas.features.JPEG` or `canvas.features.FREETYPE`. Because canvas can compile successfully even without some features it would be useful to be able to check for them and complain if they are required by the app.

LinusU:
Makes the freetype verison available as a string under `.freetypeVersion`.

Fixes #299

### DIFF
--- a/lib/canvas.js
+++ b/lib/canvas.js
@@ -57,6 +57,14 @@ if (canvas.gifVersion) {
 }
 
 /**
+ * freetype version.
+ */
+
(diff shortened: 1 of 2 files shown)


LinusU:
released as 1.3.12

PR #739, opened 2016-03-19
Clamp JPEG buffer size. Fixes #674

### LINKED ISSUE #674: jpegstream is broken since 1.2.6
After 1.2.6 jpeg stream makes for me really big files with some trash. Tested on mac and linux.

zbjornson:
NB: I was somewhat surprised that the default of 4096 doesn't likewise cause errors for tiny canvases...


zbjornson:
Also, if someone could confirm that the clamping criteria are correct, that would be great. I'm not positive that the buffer should never exceed the size of the canvas.w * canvas.h * 4. (Linus' test case passes though.)

LinusU:
Sorry for forgetting about this, merging now

LinusU:
Released as 1.3.14

Issue #232, opened 2012-11-29
[TypeError: Object #<PNGStream> has no method 'destroy']


kilianc:
Stream interface not fully implemented.

zbjornson:
From the [docs](https://nodejs.org/dist/latest-v4.x/docs/api/stream.html) and [discussion on SO](http://stackoverflow.com/questions/19277094/how-to-close-a-readable-stream-before-end) it seems like streams don't actually provide a `.destroy` method, although `fs.readStream` happens to.

See more notes in #740.

zbjornson:
Fixed in 8b5dc056b3045ffe6665c38c732dc7b37ce7c454 I think. (Definitely fixed, but not positive it was that commit.)

```js
> canvas.createPNGStream().destroy
[Function: destroy]
```

PR #740, opened 2016-03-19
Implement Readable instead of Stream

zbjornson:
Would benefit from some testers (@mitar?) and feedback...

Both JPEGStream and PNGStream currently inherit from `stream.Stream`, not `stream.Readable`. As a consequence, they are missing a bunch of methods. (See #674, #232.) They will also start flowing without a listener, but because the stream is wrapped in `process.nextTick` I think that would only be noticed from the node prompt.

It was straightforward to switch to `Readable`. I didn't attempt to make the C++ code wait to queue more data after the stream starts (first call to `.read`); after the first chunk it pushes data into the stream queue without waiting for the stream consumer to call `.read` (directly or indirectly). There should be no lost bytes as a result of flowing without a sink anymore, but consumers could see more memory consumption from the queuing of bytes that previously would have been lost into the ether.

I'm still confused about the "async" vs "sync" names of these APIs -- they all emit chunks of data within callbacks, which seems async to the consumer, if not internally.

### DIFF
--- a/lib/jpegstream.js
+++ b/lib/jpegstream.js
@@ -10,7 +10,8 @@
  * Module dependencies.
  */
 
-var Stream = require('stream').Stream;
+var Readable = require('stream').Readable;
+var util = require('util');
 
 /**
  * Initialize a `JPEGStream` with the given `canvas`.
@@ -30,33 +31,47 @@ var Stream = require('stream').Stream;
  */
 
 var JPEGStream = module.exports = function JPEGStream(canvas, options, sync) {
-  var self = this
-    , method = sync
+  if (!(this instanceof JPEGStream)) {
+    throw new TypeError("Class constructors cannot be invoked without 'new'");
+  }
+
+  Readable.call(this);
+
+  var self = this;
+  var method = sync
       ? 'streamJPEGSync'
       : 'streamJPEG';
   this.options = options;
   this.sync = sync;
   this.canvas = canvas;
-  this.readable = true;
+
   // TODO: implement async
   if ('streamJPEG' == method) method = 'streamJPEGSync';
+  this.method = method;
+};
+
+util.inherits(JPEGStream, Readable);
+
+function noop() {}
+
+JPEGStream.prototype._read = function _read() {
+  // For now we're not controlling the c++ code's data emission, so we only
+  // call canvas.streamJPEGSync once and let it emit data at will.
+  this._read = noop;
+  var self = this;
+  var method = this.method;
+  var bufsize = this.options.bufsize;
+  var quality = this.options.quality;
+  var progressive = this.options.progressive;
   process.nextTick(function(){
-    canvas[method](options.bufsize, options.quality, options.progressive, function(err, chunk){
+    self.canvas[method](bufsize, quality, progressive, function(err, chunk){
       if (err) {
         self.emit('error', err);
-        self.readable = false;
       } else if (chunk) {
-        self.emit('data', chunk);
+        self.push(chunk);
       } else {
-        self.emit('end');
-        self.readable = false;
+        self.push(null);
       }
     });
   });
 };
-
-/**
- * Inherit from `EventEmitter`.
- */
-
-JPEGStream.prototype.__proto__ = Stream.prototype;
--- a/lib/pngstream.js
+++ b/lib/pngstream.js
@@ -10,7 +10,8 @@
  * Module dependencies.
  */
 
-var Stream = require('stream').Stream;
+var Readable = require('stream').Readable;
+var util = require('util');
 
 /**
  * Initialize a `PNGStream` with the given `canvas`.
@@ -30,32 +31,59 @@ var Stream = require('stream').Stream;
  */
 
 var PNGStream = module.exports = function PNGStream(canvas, sync) {
-  var self = this
-    , method = sync
+  if (!(this instanceof PNGStream)) {
+    throw new TypeError("Class constructors cannot be invoked without 'new'");
+  }
+
+  Readable.call(this);
+
+  var self = this;
+  var method = sync
       ? 'streamPNGSync'
       : 'streamPNG';
   this.sync = sync;
   this.canvas = canvas;
-  this.readable = true;
+
   // TODO: implement async
-  if ('streamPNG' == method) method = 'streamPNGSync';
+  if ('streamPNG' === method) method = 'streamPNGSync';
+  this.method = method;
+};
+
+util.inherits(PNGStream, Readable);
+
+var PNGStream = module.exports = function PNGStream(canvas, sync) {
+  Readable.call(this);
+
+  var self = this;
+  var method = sync
+      ? 'streamPNGSync'
+      : 'streamPNG';
+  this.sync = sync;
+  this.canvas = canvas;
+
+  // TODO: implement async
+  if ('streamPNG' === method) method = 'streamPNGSync';
+  this.method = method;
+};
+
+util.inherits(PNGStream, Readable);
+
+function noop() {}
+
+PNGStream.prototype._read = function _read() {
+  // For now we're not controlling the c++ code's data emission, so we only
+  // call canvas.streamPNGSync once and let it emit data at will.
+  this._read = noop;
+  var self = this;
   process.nextTick(function(){
-    canvas[method](function(err, chunk, len){
+    self.canvas[self.method](function(err, chunk, len){
       if (err) {
         self.emit('error', err);
-        self.readable = false;
       } else if (len) {
-        self.emit('data', chunk, len);
+        self.push(chunk);
       } else {
-        self.emit('end');
-        self.readable = false;
+        self.push(null);
       }
     });
   });
 };
-
-/**
- * Inherit from `EventEmitter`.
- */
-
-PNGStream.prototype.__proto__ = Stream.prototype;
--- a/test/canvas.test.js
+++ b/test/canvas.test.js
@@ -6,7 +6,8 @@ var Canvas = require('../')
   , assert = require('assert')
   , parseFont = Canvas.Context2d.parseFont
   , fs = require('fs')
-  , os = require('os');
+  , os = require('os')
+  , Readable = require('stream').Readable;
 
 console.log();
 console.log('   canvas: %s', Canvas.version);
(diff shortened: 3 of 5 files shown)


test/canvas.test.js
```diff
@@ -749,6 +749,7 @@ describe('Canvas', function () {
   it('Canvas#createSyncPNGStream()', function (done) {
     var canvas = new Canvas(20, 20);
     var stream = canvas.createSyncPNGStream();
+    assert(stream instanceof require("stream").Readable);
```
LinusU: Single quotes please :)

LinusU:
Awesome :raised_hands: I've wanted to do this myself for quite some time, I think that this looks very good.

I think the point with "async" vs "sync" is that "sync" is being run on the main thread, although calling back to javascript land in chunks? Haven't looked at the code properly to say though.

In my opinion we shouldn't have sync streams at all since that seems like a very weird concept.

zbjornson:
Thanks!

1. I forgot that node 0.8 doesn't have stream.Readable to extend. @rvagg has a nice article about compatibility here: https://r.va.gg/2014/06/why-i-dont-use-nodes-core-stream-module.html. Drop 0.8 or make it compatible? 

2. Apparently 0.10 will miss the 'end' event if it is attached after the 'data' event is attached -- or at least, when I re-ordered the event listener attachments in the `toDataURL` method and tests, it worked. This could be breaking for consumers. I don't see anything about this in the 0.10 docs, and some quick googling didn't reveal any relevant bugs that were fixed in 0.12. https://nodejs.org/docs/v0.10.43/api/stream.html#stream_compatibility_with_older_node_versions.

test/canvas.test.js
```diff
@@ -749,26 +749,35 @@ describe('Canvas', function () {
   it('Canvas#createSyncPNGStream()', function (done) {
     var canvas = new Canvas(20, 20);
     var stream = canvas.createSyncPNGStream();
+    assert(stream instanceof require('stream').Readable);
```
LinusU: Could you move the `require` call to the top of the file?

LinusU:
Hmm

I don't have a problem with dropping 0.8 support, according to this tweet and conversation, almost no one is using 0.8 anymore. https://twitter.com/seldo/status/712386414902521861

screen shot 2016-03-25 at 13 40 20

It should be a major bump though.

The problems on 0.10 is a bit disturbing. Would it be possible for us to delay the first `data` event until the next tick?

calvinmetcalf:
Maybe the data event is causing the stream to synchronously drain before you get around to putting the end listener.

mcollina:
I agree with @calvinmetcalf. In streams 2, if all the data is available in JS-land and a `'data'` handler is added, it would be completely synch, and you will add the `'end'` event after `'end'` was already happened. Does `createSyncPNGStream` create all the data synchronously?

zbjornson:
@calvinmetcalf and @mcollina that appears to be what's happening -- thanks.

I was hoping to find documentation about this change between 0.10 and 0.12 just to confirm that I wasn't doing something wrong. I think this is the diff that caused the behavior change:
https://github.com/nodejs/node-v0.x-archive/commit/0f8de5e1f96a07fa6de837378d29ac5f2719ec60#diff-ba6a0df0f5212f5cba5ca5179e209a17L681
Notice `.resume` calls through a `nextTick`, whereas the removed `emitDataEvents` looks like it plowed straight through.

Anyway, wrapping it in a `nextTick` now...

zbjornson:
From https://github.com/nodejs/node-v0.x-archive/pull/5865#issue-16941454

> The other semantic change is that adding a readable event handler will not cause readable to be emitted until nextTick

Indeed.

zbjornson:
Passing in 0.10 and later. Undid the changes to `toDataUrl` and tests.

Node 5 doesn't seem to need to be an allowed failure anymore, btw.

lib/jpegstream.js
```diff
@@ -30,33 +31,39 @@ var Stream = require('stream').Stream;
...
+  // call canvas.streamPNGSync once and let it emit data at will.
+  var hasStarted = false;
+
+  self._read = function () {
+    if (!hasStarted) {
+      hasStarted = true;
+      process.nextTick(function(){
+        canvas[method](options.bufsize, options.quality, options.progressive, function(err, chunk){
+            if (err) {
+            self.emit('error', err);
```
LinusU: wrong indentation

lib/jpegstream.js
```diff
@@ -30,33 +31,39 @@ var Stream = require('stream').Stream;
...
+            } else if (chunk) {
+            self.push(chunk);
+            } else {
+            self.push(null);
+            }
+        });
```
mcollina: I would suggest you not to allocate another function for `nextTick`, but rather define it in the constructor, or outside. Or you can use http://npm.im/process-nextick-args.
Also, the callback for the canvas method can be recycled as well.

lib/pngstream.js
```diff
@@ -30,32 +31,37 @@ var Stream = require('stream').Stream;
  */
 
 var PNGStream = module.exports = function PNGStream(canvas, sync) {
```
mcollina: I would add a:

```js
if (!(this instanceof PNGStream)) {
  return new PNGStream(canvas, sync)
}
```

To avoid pollution of the global namespace if invoked without `new`.

lib/jpegstream.js
```diff
@@ -10,7 +10,8 @@
  * Module dependencies.
  */
 
-var Stream = require('stream').Stream;
+var Readable = require('stream').Readable;
```
mcollina: You should depend from `readable-stream`, it is better practice and yields more stability across node versions.

lib/jpegstream.js
LinusU: :+1:

lib/pngstream.js
LinusU: I would prefer throwing an error since this is the default behaviour with ES6 classes which we probably will want to upgrade the code base to at some point. Also, you can always stop throwing an exception without a major bump but going the other way is more complicated.

This is the exact error the Node throws:

```javascript
throw new TypeError("Class constructors cannot be invoked without 'new'")
```

lib/jpegstream.js
LinusU: It does require pulling in another dependency though...

mcollina: Considering the fact that most of the stream utilities depends on readable-stream, it's not really problematic (the users will do depend on it anyway). The only issue might be in the browser, but I don't think this can be used there.

calvinmetcalf: (in browserify the stream module is based on readable-stream so it's no big deal)

lib/pngstream.js
zbjornson: (Addressed. GitHub isn't hiding this because the diff is below the comment.)

lib/jpegstream.js
zbjornson: I'm confused what's going on here. Evidently... if I store a reference to a C++ function (i.e. `var x = canvas[method]; x();`) then I get an `Illegal invocation` error when I go to invoke it. A simple `canvas[method]()` works though.

In any case, @mcollina I'm curious what the motivation would be for pulling the two functions out into the constructor or outside. Because of the wrapping `if (!hasStarted)` block, I don't see any performance reasons.

lib/jpegstream.js
```diff
@@ -30,33 +31,43 @@ var Stream = require('stream').Stream;
...
+            self.emit('error', err);
+          } else if (chunk) {
+            self.push(chunk);
+          } else {
+            self.push(null);
+          }
+        });
+      });
+    }
+  };
```
mcollina: allocating `read` here or in the prototype has few differences, apart from style. I prefer to use `prototype` when dealing with a class an inheriting.

Regarding this block, I would consider removing the `process.nextTick`  call. Plus, I think it's possible to recycle (declare inside the constructor) the function passed to `canvas[method]`.

The golden rule is "try not to allocate functions inside `_read`, `_write` or `_transform`". That is usually a very hot path and all those function allocation will slow you down.

mcollina:
If you care about browser size, I suggest you using `inherits` from NPM instead of `util.inherits`.

lib/jpegstream.js
zbjornson: From earlier, we must use `process.nextTick`.

Sorry for not being more clear, or for still not understanding what you're saying, but this is not a hot path because it has the surrounding `if` block that causes evaluation of the function expression exactly once, thus there is nothing to reuse. I think then the only difference is style, but please let me know if I'm missing something.

As far as size of npm/inherits vs `util.inherits`, this is a node module only.

zbjornson:
Updated travis.yaml: remove node 0.8, add 6.x and no longer allow 5.x to fail. CI green!

zbjornson:
(Needs update since PDF stream was added)

LinusU:
@zbjornson sorry for the delay on this, I've now cut a 1.x branch from master and from now on master is for Canvas 2.x. I'm dropping 0.8 support in #825 so this should be good to merge 👍

zbjornson:
No problemo, will fix conflicts and add PDF stream this week. :)

zbjornson:
@LinusU updated :)

chearon:
I do think the comments by @mcollina deserve another look, can `_read` be defined on the prototype? If we drop 0.10, does that mean the `nextTick` stuff can be removed? It seems like it is not an issue in streams3

zbjornson:
If 0.10 is dropped, `nextTick` can be removed, yes.

I had `_read` in the ctor so that `hasStarted` is not visible, but that's a small reason. There's no difference in stream performance, only in stream construction:

* As it was: construct 1,430,136 ops/sec, end event 576 ops/sec
* As it was with guard clause reversed (`if (hasStarted) return`): 1,415,082 & 573
* On proto: 8,442,640 & 574
* On proto, reassign `_read = noop` on first `_read` call: 8,452,849 & 580
* `nextTick` removed: no change

The 4th option (reassign to noop) is the best of both worlds -- no internal properties exposed and fast construction. Updated PR to use that method.

zbjornson:
@LinusU gentle nudge, from my side this is ready for merge into 2.0. :)

LinusU:
Sorry for the delay, I hope to get some time this week 👍

LinusU:
Sorry for the delay, awesome work!

zbjornson:
TODO at myself (or anyone else who wants to do it): since 0.10 is dropped now, the nextTick stuff can be removed, per discussion above.

PR #750, opened 2016-04-08
Fix segfault in putImageData

### LINKED ISSUE #749: `putImageData(undefined, x, y)` segfaults
Tested using canvas@1.3.12 and Node.js v5.9.1.

Test case:

```js
'use strict';

const Canvas = require('canvas');

const canvas = new Canvas(1, 1);
context.putImageData(undefined, 0, 0);
```

Note that this is common when the first `putImageData` argument comes from a lookup dictionary, e.g. `someObject[foo]`, where the lookup fails. It should error, sure, but not segfault.

```sh
$ node whoops.js
Segmentation fault: 11
```

For comparison, Chromium/Blink throws the following exception:

```
Uncaught TypeError: Failed to execute 'putImageData' on 'CanvasRenderingContext2D': parameter 1 is not of type 'ImageData'.
```

LinusU:
*fixes #749*

ping @mathiasbynens, care to test this? :)

### DIFF
--- a/src/CanvasRenderingContext2d.cc
+++ b/src/CanvasRenderingContext2d.cc
@@ -571,6 +571,8 @@ NAN_METHOD(Context2d::AddPage) {
  */
 
 NAN_METHOD(Context2d::PutImageData) {
+  if (!info[0]->IsObject())
+    return Nan::ThrowTypeError("ImageData expected");
   Local<Object> obj = info[0]->ToObject();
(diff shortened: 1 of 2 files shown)


mathiasbynens:
I can confirm that the test case in #749 passes (i.e. throws but does not segfault) with this patch. Thanks, @LinusU!

LinusU:
Released as `1.3.15`

PR #756, opened 2016-04-15
Avoid permissions error in Vagrant

tranzium:
Vagrant + Windows can prevent file permission changes

### DIFF
--- a/binding.gyp
+++ b/binding.gyp
@@ -10,10 +10,10 @@
       }
     }, { # 'OS!="win"'
       'variables': {
-        'with_jpeg%': '<!(./util/has_lib.sh jpeg)',
-        'with_gif%': '<!(./util/has_lib.sh gif)',
-        'with_pango%': '<!(./util/has_lib.sh pango)',
-        'with_freetype%': '<!(./util/has_lib.sh freetype)'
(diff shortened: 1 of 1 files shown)


zbjornson:
This looks closely related to the first commit in #754 and subject to the same issues. But I don't understand how this is a Windows problem -- the change is in the non-Windows part of the binding file -- can you explain that please?

tranzium:
From what I could diagnose, changes to the file's permissions (including +x) could not be applied, thus the util script could not be executed for the install.

Using both chmod and chown result in the file being the exact same.

Although this could be a Vagrant issue, I made the changes in the assumption that one was not able to change their Vagrant configuration.

zbjornson:
I see, thanks for the info. I'm still not quite clear if this relates to Windows (mentioned in the commit message but not the PR title). Are you using a set of extensions on top of Windows to get `chmod` and `chown`, or are you actually running on a linux distro?

tranzium:
I am running Ubuntu in Vagrant (virtual machine).
I will test adjusting the sync_folder permissions and see if I can get the /util/has_lib.sh script running directly from a `npm install canvas` call. Will report any further findings.

LinusU:
This script has been migrated to node now so I don't think this should be a problem anymore...

PR #774, opened 2016-05-26
Fix leak of data when streaming JPEG

### LINKED ISSUE #773: jpegStream doesn't truncate last buffer?
I just noticed that all files created with `canvas.jpegStream` have a size that's multiple of 4096. Inside of those files after JPEG EOI (end of image marker, 0xFFD9) I found unrelated data from memory (something looking like internal V8 data structures).

It's almost like jpegStream allocates 4096 buffers and never truncates the last one when data size is less than 4096.

Here's minimal example:

```js
#!/usr/bin/env node

var fs     = require('fs')
var img    = new Canvas.Image()
var canvas = new Canvas(200, 133)

img.src = fs.readFileSync('./sample.jpeg')

canvas.getContext('2d').drawImage(img, 0, 0, 200, 133)

var buffers = []

canvas.jpegStream({ quality: 95 })
  .on('data', function (data) {
    console.log('rcvd', data.length) // <-- this always prints 4096, even for the last buffer!
    buffers.push(data)
  })
  .on('end', function () {
    fs.writeFileSync('./out.jpeg', Buffer.concat(buffers))
  })
  .on('error', console.log)
```

Saved JPEG image looks like this:

```
0000000: ffd8 ffe0 0010 4a46 4946 0001 0100 0001  ......JFIF...... // start of a usual JPEG
0000010: 0001 0000 ffdb 0043 0002 0101 0101 0102  .......C........
...
0003a20: e8a2 b65b 1cbf 68a8 46f6 795d 8955 7202  ...[..h.F.y].Ur.
0003a30: 67a9 1ce4 fad5 bd3d 8dc4 2b23 9e5c 039f  g......=..+#.\..
0003a40: 4a28 a9ea 69d0 b37f 0c36 ed1c 11c4 327a  J(..i....6....2z
0003a50: b1e4 9e94 5145 5b4a e4c5 be53 ffd9 0000  ....QE[J...S.... // end of actual image
0003a60: 3094 2e02 0000 0000 0100 0000 0000 0000  0...............
0003a70: 2aca 1389 1100 0000 9899 2e02 0000 0000  *...............
0003a80: 0400 0000 0100 0000 9894 2e02 0000 0000  ................
...
0003d50: 709c 2e02 0000 0000 0400 0000 0000 0000  p...............
0003d60: 726f 7065 7274 7900 d09f ee00 0000 0000  roperty......... // that's definitely not JPEG
0003d70: d051 2702 0000 0000 0100 ffff 0000 0000  .Q'.............
...
0003fe0: 0000 0000 4883 ec10 1110 0000 0000 0000  ....H...........
0003ff0: 50af 2e02 0000 0000 b877 0985 7d7f 0000  P........w..}... // end of file
```

LinusU:
Fixes #773

### DIFF
--- a/src/JPEGStream.h
+++ b/src/JPEGStream.h
@@ -53,7 +53,7 @@ term_closure_destination(j_compress_ptr cinfo){
   closure_destination_mgr *dest = (closure_destination_mgr *) cinfo->dest;
 
   /* emit remaining data */
-  Local<Object> buf = Nan::NewBuffer((char *)dest->buffer, dest->bufsize).ToLocalChecked();
+  Local<Object> buf = Nan::NewBuffer((char *)dest->buffer, dest->bufsize - dest->pub.free_in_buffer).ToLocalChecked();
 
   Local<Value> data_argv[2] = {
       Nan::Null()
--- a/test/canvas.test.js
+++ b/test/canvas.test.js
@@ -817,7 +817,7 @@ describe('Canvas', function () {
       bytes += chunk.length;
     });
     stream.on('end', function(){
-      assert.equal(bytes, 8192);
+      assert.equal(bytes, 5427);
       done();
     });
     stream.on('error', function(err) {


LinusU:
ping @rlidwka, would you mind reviewing/testing this for me? :)

I have verified that the image does indeed get smaller, and it's no longer a multiple of 4096. It can also open the image in the Preview app in OS X so it seems valid.

LinusU:
@TooTallNate @kangax @rvagg @kkoopa @zbjornson 

Any thoughts on how to handle this, we are leaking users memory, potentially in web apps which could be exploited.

I think that we should `npm deprecate` all the current versions with a message on what's happening and urging all users to update to a version that we'll publish when we merge this.

zbjornson:
That sounds reasonable and responsible. By "all current versions" do you mean all 86 published versions?

Nice job getting a fix submitted quickly. :+1:

LinusU:
Published as 1.3.16

Would still appreciate input on how to move forward with the deprecation...

kkoopa:
I guess whether to deprecate or not depends on how much JPEG streaming is used.

PR #777, opened 2016-06-03
CanvasRenderingContext2D class name

arusakov:
Rendering context called "CanvasRenderingContext2D" in the [specification].(https://html.spec.whatwg.org/#2dcontext) 
Also I saw [object CanvasRenderingContext2D] in Chrome, FF and Safari after this code:
```
document.createElement('canvas').getContext('2d').toString();
```


LinusU:
Thanks 🙌

PR #781, opened 2016-06-09
Crude PDF stream implementation

### LINKED ISSUE #778: Max 1GB PDF output size re 1GB nodejs buffer limit
We're creating PDFs with node-canvas and I've found the following error when generating particularly large files:

```
FATAL ERROR: v8::Object::SetIndexedPropertiesToExternalArrayData() length exceeds max acceptable value
Aborted (core dumped)
```
I think this is because the PDF buffer generated from `canvas.toBuffer()` exceeds 1GB as per:
http://stackoverflow.com/questions/8974375/whats-the-maximum-size-of-a-node-js-buffer

This seems quite likely, as it's a 120 page PDF with multiple high-resolution images in it!

If this is the problem, is there a way to stream the PDF output instead, increase node's 1GB limit, or any other potential solutions?

kkoopa:
A Minimial Working PDFStream.

Fixes #778

However, I would not advise merging it as such. It would be better to clean up all the streams code, avoiding the need to keep everything in memory and eliminating unnecessary copying of data. It would also be nice to finally have the asynchronous streams. This would entail a major rewrite of the library and its API, which I do not have time to do.

### DIFF
--- a/test/canvas.test.js
+++ b/test/canvas.test.js
@@ -802,6 +802,24 @@ describe('Canvas', function () {
     });
   });
 
+  it('Canvas#createSyncPDFStream()', function (done) {
+    var canvas = new Canvas(20, 20, 'pdf');
+    var stream = canvas.createSyncPDFStream();
+    var firstChunk = true;
+    stream.on('data', function (chunk) {
+      if (firstChunk) {
+        firstChunk = false;
+        assert.equal('PDF', chunk.slice(1, 4).toString());
+      }
+    });
+    stream.on('end', function () {
+      done();
+    });
+    stream.on('error', function (err) {
+      done(err);
+    });
+  });
+
   it('Canvas#jpegStream()', function (done) {
     var canvas = new Canvas(640, 480);
     var stream = canvas.jpegStream();
--- a/lib/canvas.js
+++ b/lib/canvas.js
@@ -16,6 +16,7 @@ var canvas = require('./bindings')
   , cairoVersion = canvas.cairoVersion
   , Context2d = require('./context2d')
   , PNGStream = require('./pngstream')
+  , PDFStream = require('./pdfstream')
   , JPEGStream = require('./jpegstream')
   , FontFace = canvas.FontFace
   , fs = require('fs')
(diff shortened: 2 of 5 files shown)


zbjornson:
Nice.

> "It would also be nice to finally have the asynchronous streams"

Could you explain how the current streams are synchronous please? This came up when I was working on revising streams in #740, but I'm not sure in what way a stream can be synchronous.

kkoopa:
Because they block the event loop by running on the same thread as V8. For example, [this](https://github.com/kkoopa/node-canvas/blob/53530454fae4df764362446c0b477c7d16424e3f/src/Canvas.cc#L467-473) keeps going (with intermittent calls to the JS stream callback) until all data in the Cairo surface has been processed.

To make them async, they should run in a separate thread. `Nan::AsyncProgressWorker` could be shoehorned in to do that.

However, what I would like to see is a way of operating directly on a stream, without storing an entire Cairo surface in memory. For example, create a Canvas bound to a stream which could then be redirected to a file (or a Buffer, or a socket, or anything else). Operations on the Canvas would stream out in chunks, never using more memory than the size of a chunk. A canvas created in this fashion could not conceivably support `toBuffer` or deletion, but would be highly useful for non-interactive work.

test/canvas.test.js
```diff
@@ -802,6 +802,24 @@ describe('Canvas', function () {
...
+    stream.on('data', function (chunk) {
+      if (firstChunk) {
+        firstChunk = false;
+        assert.equal('PDF', chunk.slice(1, 4).toString());
+      }
+    });
+    stream.on('end', function () {
+      done();
+    });
+    stream.on('error', function () {
```
LinusU: Missing `err` in the functions parameter list...

LinusU:
Sorry for the delay on merging this, just one tiny nit and then we're good to go 👍

test/canvas.test.js
kkoopa: Ah, yes. I must have missed it. However, the code always returns `CAIRO_STATUS_SUCCESS`, since nothing can fail in that sense, so there will actually never be an error event. But I guess it's better to leave it in case someone changes the code in the future.

yy1107:
在线合并

jakeg:
Any progress on this? Was the "tiny nit" fixed?

LinusU:
Sorry for the delay, merging now 👍

PR #793, opened 2016-07-15
fix lineDash browser tests never finishing

chearon:
*before*
before

*after*
after


LinusU:
Awesome! Man this has been bugging me 😆

LinusU:
Thanks!

PR #797, opened 2016-07-27
Update CI settings

LinusU:
Also keeps 0.8 since we don't want to drop that support until 2.x


LinusU:
The difference between GCC 4.8 and 4.9 was just 7 seconds which suggest that there isn't any difference at all in speed...

PR #808, opened 2016-09-10
Fix accessing prototypes of Canvas, Image, Context2d

### LINKED ISSUE #76: .prototype segfaults
I like to inspect prototype attributes to get a feel for what methods objects support, but this causes segfaults on most objects, like this one:

```
$ node
> require('canvas').prototype

rlwrap: warning: node killed by signal 11 - rlwrap will now kill itself with the same signal
Segmentation fault
```

This is on node v0.2.5 and it seems that most everything else works.

### LINKED ISSUE #568: Hard crash calling Context2d.prototype.filter
This line will crash node:

Context2d.prototype.filter

Perhaps worse, so do this:

Object.getOwnPropertyDescriptor(Context2d.prototype, "filter")

I think this is because GetFilter doesn't have a HasInstance guard. I think it should do something along the lines of

if (!HasInstace(this)) return default_value

(as opposed to throwing, since then getOwnPropertyDesctipor would throw. This would match Array.prototype.length for example which is 0).

### LINKED ISSUE #803: Issue: Crash when access Canvas.prototype
image

``` js
Canvas.prototype // Crash here
```

Node-canvas 1.4.0 on Node JS 5.12.0 on Windows 10 (1607)

### LINKED ISSUE #847: C++ Runtime Error when debuging
## Issue or Feature
If I debug my js with a debugger, and I set a breakpoint, then the node-canvas fail and let the whole node fail

## Steps to Reproduce
```js
var Image = Canvas.Image // Make a breakpoint on this line, I use Visual Studio Code to do this
```

## Your Environment
* Version of node-canvas: 1.6.2
* Environment (e.g. node 4.2.0 on Mac OS X 10.8): Windows 10, Node 6.9

image

Thanks

### LINKED ISSUE #885: for windows
sorry i meet the same question,the installation on windows seems not success,i have read the steps told,but it seems not very useful!

zbjornson:
Fixes #803 
Fixes #847
Fixes #76 
Fixes #885
Fixes nodejs/node#15099
Fixes #568

### DIFF
--- a/src/Canvas.cc
+++ b/src/Canvas.cc
@@ -13,6 +13,7 @@
 #include <cairo-pdf.h>
 #include <cairo-svg.h>
 
+#include "Util.h"
 #include "Canvas.h"
 #include "PNG.h"
 #include "CanvasRenderingContext2d.h"
@@ -57,10 +58,10 @@ Canvas::Initialize(Nan::ADDON_REGISTER_FUNCTION_ARGS_TYPE target) {
 #ifdef HAVE_JPEG
   Nan::SetPrototypeMethod(ctor, "streamJPEGSync", StreamJPEGSync);
 #endif
-  Nan::SetAccessor(proto, Nan::New("type").ToLocalChecked(), GetType);
-  Nan::SetAccessor(proto, Nan::New("stride").ToLocalChecked(), GetStride);
-  Nan::SetAccessor(proto, Nan::New("width").ToLocalChecked(), GetWidth, SetWidth);
-  Nan::SetAccessor(proto, Nan::New("height").ToLocalChecked(), GetHeight, SetHeight);
+  SetProtoAccessor(proto, Nan::New("type").ToLocalChecked(), GetType, NULL, ctor);
+  SetProtoAccessor(proto, Nan::New("stride").ToLocalChecked(), GetStride, NULL, ctor);
+  SetProtoAccessor(proto, Nan::New("width").ToLocalChecked(), GetWidth, SetWidth, ctor);
+  SetProtoAccessor(proto, Nan::New("height").ToLocalChecked(), GetHeight, SetHeight, ctor);
 
   Nan::SetTemplate(proto, "PNG_NO_FILTERS", Nan::New<Uint32>(PNG_NO_FILTERS));
   Nan::SetTemplate(proto, "PNG_FILTER_NONE", Nan::New<Uint32>(PNG_FILTER_NONE));
--- a/src/CanvasRenderingContext2d.cc
+++ b/src/CanvasRenderingContext2d.cc
@@ -12,6 +12,7 @@
 #include <algorithm>
 #include <string>
 
+#include "Util.h"
 #include "Canvas.h"
 #include "Point.h"
 #include "Image.h"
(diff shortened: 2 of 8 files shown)


zbjornson:
Urg, no real desire to debug node 0.10 and 0.8.

The Canvas 2.0 proposal drops 0.8, at least...

zbjornson:
@LinusU I just made a change to allow node v0.8 and v0.10 to skip the tests that I added in this PR. None of the existing tests fail in v0.10 so this seems harmless to 0.10 and beneficial to 0.12+. What do you think of merging this?

zbjornson:
I can reproduce that, but the access violation is in libcairo-2.dll, not node-canvas. I can't see enough of the call stack to tell if it's a bad usage of libcairo or an actual libcairo bug.

@chearon does your msys2 build use a newer version of cairo? I don't have the msys2 build chain setup, but it might be informative to test this branch with newer cairo.

chearon:
I think the GTK.zip on here is from 2010 so the MSYS version should almost certainly be much newer. If it helps I made a [wiki](https://github.com/Automattic/node-canvas/wiki/Installation---Windows-(with-MSYS2)) documenting how I got it working, although I need to update it for Pango

basvanmeurs:
Although the error message seems to disappear, nodejs still exits.
If the access violation is in cairo, I wonder why the debugger braiks even on just inspecting the module root? gles2 is my own C++ module root for comparison, which works just fine.

  7 var gl = gles2.init(options);
  8 
> 9 debugger;
 10 
 11 var img = new Image();
repl
Press Ctrl + C to leave debug repl
> gles2
{ init: [Function: val], blit: [Function: val] }
> Canvas
program terminated

zbjornson:
@basvanmeurs what OS are you running? gles2 doesn't link to Cairo, does it?

LinusU:
@zbjornson Would you mind rebasing this on latest master as soon as I have merged #915?

Also, I think it's better to fix `History.md` when merging so that we don't get conflicts in the PRs, what do you think?

zbjornson:
> Would you mind rebasing

Done!

>better to fix History.md when merging

Sure, dropped that change.

LinusU:
> Segmentation fault (core dumped)

Hmm 🤔 seems like something isn't quite right...

zbjornson:
Boo. Gonna take me a bit longer to look into that, won't get to it today...

LinusU:
No problem, I'll try and have a look if I have time as well, but the clock just turned midnight here and I need to get up to work tomorrow :)

mightyiam:
It doesn't work for me from this branch. I still get a crash when inspecting.

zbjornson:
I need to work on this more sometime. I think this PR is a correct change to make based on some feedback from a core node.js dev, but another change is definitely needed to fully fix the error. I came across this node.js core issue, which looks similar and might provide some hints: https://github.com/nodejs/node-v0.x-archive/issues/9028

zbjornson:
I think this is finally fixed, but I posed a [small question](https://github.com/nodejs/node/issues/15099#issuecomment-326835601) to the node.js devs that would be good to know before landing.

src/Canvas.cc
```diff
@@ -51,16 +52,17 @@ Canvas::Initialize(Nan::ADDON_REGISTER_FUNCTION_ARGS_TYPE target) {
 
   // Prototype
   Local<ObjectTemplate> proto = ctor->PrototypeTemplate();
+  proto->SetInternalFieldCount(1);
```
TimothyGu: Hey @zbjornson, thanks for linking to this PR.

While this is just fine as a Band-Aid for the problem, I'd like to note several potential problems that would stem from this implementation. First, philosophically, the prototype object should not be an instance of the class -- that is, it shouldn't have internal slots like instances do. JavaScript got this wrong with `Array.prototype` and `Function.prototype` (`Array.isArray(Array.prototype)` and `typeof Function.prototype === 'function'` both return true), but they have rectified this problem with future additions to the language with a note in the [spec](https://tc39.github.io/ecma262/#sec-properties-of-the-array-prototype-object):

> The Array prototype object is specified to be an Array exotic object to ensure compatibility with ECMAScript code that was created prior to the ECMAScript 2015 specification.

More concretely, although accessing `Canvas.prototype.type` will not throw, other code fragments e.g. `Object.create(Canvas.prototype).type` will still throw, I believe. Even worse (but more unlikely), `Object.setPrototypeOf(anotherObjectWithInternalFields, Canvas.prototype).type` may cause memory issues, as the type of the object pointer wrapped in V8 object is untyped.

If a Band-Aid is all that is sought here, then please disregard my suggestions below to make this more robust. Either suggestion will unfortunately incur quite some work, so it's totally up to you regarding what's best for this project.

1. Write a pure JS wrapper around the JS object created in C++, and implement the property getters/setters as ordinary functions (e.g. `getLength()`, `setLength()`). This means that you would do all necessary type checking in JS, so in C++ you can feel free to assert() on predicates. The C++ wrapper would also be hidden in a symbol, so we can tell any user that get curious and call the prototype methods w/o context, well, don't try to access things you are not supposed to like hidden properties. This is generally what we do in Node.js.

2. Check if `info.Holder()` is an instance of the `FunctionTemplate`, and throw an error if it is not. This is what Chrome [does](https://cs.chromium.org/chromium/src/third_party/WebKit/Source/platform/bindings/V8PerIsolateData.cpp?l=310&rcl=4f7fcf34302b3c09572d978bcd3e58fe6eb5c703) for DOM, but it would require holding a reference to the FunctionTemplate, always. If you were to go down this route, I would put the Persistent reference to the FunctionTemplate in a `std::unordered_map` keyed on `Isolate*`.

A lot of this is based on speculation, so please let me know if you found a better solution to all of this!

zbjornson: Thanks so much for the detailed and helpful info!

>  the prototype object should not be an instance of the class -- that is, it shouldn't have internal slots like instances do

Oops, `proto->SetInternalFieldCount(1)` was leftover from before I added the signatures to the `SetAccessor`s, and no value is set for the slot anyway. Will remove in a moment.

As far as the proto being an instance of the class: my understanding is that internal slots are usable for anything and have no effect on the behavior of the object (they're just C++-accessible storage). `Canvas.prototype instanceof Canvas === false`, so I think we're okay?

> `Object.create(Canvas.prototype).type` will still throw, I believe. ... `Object.setPrototypeOf(anotherObjectWithInternalFields, Canvas.prototype).type` may cause memory issues

Both of these throw the "incompatible receiver" error (as desired; same as `Canvas.prototype.type`), and neither have the assertion failure.

> Check if info.Holder() is an instance of the FunctionTemplate

I think this is what v8 does actually, when `SetAccessor` is used with an `AccessorSignature`:

> *The signature describes valid receivers for the accessor and is used to perform implicit instance checks against them. If the receiver is incompatible (i.e. is not an instance of the constructor as defined by FunctionTemplate::HasInstance()), an implicit TypeError is thrown and no callback is invoked.*

Am I reading that right? If so, then I think this PR is actually a sturdy route forward.

TimothyGu: I wasn't familiar with how AccessorSignature worked... and it looks like that's it! This looks good to me.

zbjornson: Thank you very much for reviewing!

zbjornson:
@LinusU this should be ready to go 🚀. bnoordhuis answered my question in the linked thread, no change needed.

LinusU:
Awesome!!

@zbjornson mind rebasing and fixing the conflicts? ❤️

zbjornson:
Done! :)

LinusU:
Super!

Jack-Works:
Wait for 1 year! 🤣🤣

LinusU:
...and one day 😆

PR #810, opened 2016-09-10
fix #809

### LINKED ISSUE #809: If you call ctx.restore while on the initial state, nothing will get drawn
To reproduce, open any of the examples and add a call to `ctx.restore()` before any other operations on `ctx`. The resulting image will be completely empty, which isn't what happens in the browser

Seems like `cairo_restore` does not check if you're on the initial state for some reason, or maybe it works differently than we expect

Cairo 1.14.6

chearon:
See #809

### DIFF
--- a/src/CanvasRenderingContext2d.cc
+++ b/src/CanvasRenderingContext2d.cc
@@ -191,8 +191,13 @@ Context2d::~Context2d() {
 
 void
 Context2d::save() {
-  cairo_save(_context);
-  saveState();
+  if (stateno < CANVAS_MAX_STATES) {
+    cairo_save(_context);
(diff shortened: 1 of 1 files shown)


zbjornson:
See also #316 ... looks like it has/had the same goal?

LinusU:
Hmm, would it make sense to merge `restore` and `restoreState`? That way we are sure that we don't introduce any new bugs by removing the check from `restoreState`

chearon:
oh yeah you're right, I didn't know about #316 

@LinusU makes sense I'll update in a sec

chearon:
I've merged those functions as requested and updated it to be compatible with #715

LinusU:
Thank you 👍

zbjornson:
Wanna close #316 since this makes it redundant? :)

PR #811, opened 2016-09-12
Render SVG img elements when librsvg is available

foysavas:
In the browser, canvas can render SVG img elements.

This commit enables node-canvas to do the same, if librsvg is installed.
Note that librsvg remains an optional dependency.

Use cases for embedded SVG images include the rendering of d3 charts,
which often blend SVG and canvas content.

Big thanks to petarov and scott113341 for much of the original code.


petarov:
Nice job! I'll have a look tomorrow. I got some comments as well.

src/Image.cc
```diff
@@ -420,6 +428,24 @@ Image::loadSurface() {
   if (isJPEG(buf)) return loadJPEG(stream);
 #endif
 
+// svg
+#ifdef HAVE_RSVG
+  // confirm svg using first 1000 chars
+  // if a very long comment precedes the root <svg> tag, isSVG returns false
+  uint8_t head[1000] = {0};
+  fseek(stream, 0 , SEEK_END);
```
petarov: Would it be better to first call `isSVG` on the 5 bytes already read in `buf` and, if no `svg` bytes are found, then proceed to do `fseek` and `rewind`? It might save some ops, if most svg files got no xml or other comments at the beginning.

src/Image.cc
```diff
@@ -218,6 +218,14 @@ Image::loadFromBuffer(uint8_t *buf, unsigned len) {
   }
 #endif
 #endif
+#ifdef HAVE_RSVG
+  // confirm svg using first 1000 chars
+  // if a very long comment precedes the root <svg> tag, isSVG returns false
+  uint8_t head[1000] = {0};
+  unsigned head_len = (len < 1000 ? len : 1000);
+  memcpy(head, buf, head_len * sizeof(uint8_t));
+  if (isSVG(head, head_len)) return loadSVGFromBuffer(buf, len);
```
petarov: Is it possible to directly pass `buf` to `isSVG` here? I mean `head_len` should already specify the max length, so one extra `memcpy` could be skipped. ;)

foysavas: Good catch. I'll revise to:

```
  unsigned head_len = (len < 1000 ? len : 1000);
  if (isSVG(buf, head_len)) return loadSVGFromBuffer(buf, len);
```

src/Image.cc
foysavas: Thanks @petarov.

I quickly surveyed SVGs in the wild and found only 12% starting with `<svg>` (not sure if my sample set was biased though). Other starts include some combination of `<?xml>`, `<!DOCTYPE>`, `<!-- -->`, and whitespace. For instance Inkscape SVGs start with this preamble:

```
<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<!-- Created with Inkscape (http://www.inkscape.org/) -->
```

What do you think? Should we first try the five bytes before trying the first 1000 since they're around or go straight to the 1000? Like you said, it's only a few ops, but since these lines may get a lot of use, I'm happiest doing it best. Please tell me either way.

petarov: That's a good point. Perhaps most SVGs do indeed start with xml processing instructions and/or preambles. I don't have a large SVG set to either confirm or discard that. What I have is SVGs for the app I used to work on, but that's meaningless in this context. 

Perhaps it's better to leave it this way then and don't do any premature optimizations at least until someone provides more info. 

Cheers 🍺

foysavas:
@LinusU should we modify the readme? Otherwise, I think this PR is near ready for merge.

LinusU:
This is awesome 🍾 sorry for the delay on this, I hope to have time to properly look at it soon 👍

Rush:
Thanks, your PR seems to be working but unfortunately the SVG is being immediately rasterized. What I'd love to do is import the SVG, draw it onto the canvas, draw some more stuff and export it as SVG. Will it be doable?

LinusU:
@Rush while probably technically possible with Cairo it's a bit outside the scope of this package, which aims to be a implementation of the web canvas api...

LinusU:
@Rush actually, seems like I'm wrong. It could potentially work if you initialise the canvas with the string `'svg'` as the format. e.g. `new Canvas(400, 400, 'svg')`

Rush:
@LinusU I did try that but it does not work in your current version of code.

LinusU:
Hmm, looking at the code in this pull request it seems like it's rasterizing the svg as soon as it's loaded.

@foysavas Do you think it would be possible to store the loaded svg as an `RsvgHandle` and then using `rsvg_handle_render_cairo` only when the image is to be drawn to a canvas?

foysavas:
Avoiding immediate rasterization should be possible. Though I've never personally used the `Image#dataMode` functionality of `node-canvas`, we should be able to use `Image#dataMode` settings [as is done with JPEGs](https://github.com/foysavas/node-canvas/blob/bdf2d1104cb8aa6c4ff60bcb149d20dc20946853/src/Image.cc#L212-L216). Here `Image::assignDataAsMime` calls [`cairo_surface_set_mime_data`](https://cairographics.org/manual/cairo-cairo-surface-t.html#cairo-surface-set-mime-data), which claims support for SVG.

brunobasto:
Hey guys,

Any plans for releasing this?

cammanderson:
Looking forward to this too. Is it far off? :-)

foysavas:
I have rebased to master resolving a few conflicts. As this is the second time, I've done so, please let's merge soon. I don't want this to go stale.

As far as SVG Image#dataMode support goes, I think it best left for a later PR likely by someone else, as I don't intend to make use of such a feature myself.

cammanderson:
Very cool, look forward to this merge!

asturur:
i like this too.
I think is gonna close a couple of issue on fabricjs issue tracker also about node differences.

For what concers direct svg ouput i can add my personal experience that not all canvas api are the same in SVG.
For example in canvas you can scale the context, fill a path, unscale the context, stroke the same path. This will give you a thinner border that is not reproducible in SVG directly.
Or shadows, canvas shadow do not get output in svg and you need to build some effect for them, i  do not think cairo has this kind logic under the hood.

afdev82:
Still no news about this PR?

kevinthr:
I want this

klaudia9:
Will this be released anytime soon?

marcusstenbeck:
Ping. :)

marcusstenbeck:
I just tested this, and it's working for me. Is there a hold-up for this not being merged?

geuis:
Hey guys, lets get this in.

LinusU:
There is currently conflicts with the master branch, but otherwise this should be fine to land 👍

foysavas:
@LinusU, conflicts resolved via rebase to master. Let's merge this!

LinusU:
Awesome, thank you!

martijn94:
Any idea on when this will be released?

martijn94:
Any idea when this will be released in a 2.x release?

petahz:
@kfoxder wondering the same thing, when is this going to be released?

LinusU:
This is out in the latest 2.x release

PR #815, opened 2016-09-24
Windows JPEG support

timknip:
This PR adds JPEG support on windows.

Basically a duplicate of #458, but this one needs explicit options to be be set on `npm install` and `node-gyp rebuild`. The explicit options make sure nothing breaks.

It needs libjpeg-turbo or libjpeg-turbo64 installed.

Added the `jpeg_root` variable to binding.gyp (default: `c:/libjpeg-turbo64`).
Also `jpeg62.dll` will be copied to `build/Release`

Install:
`node-gyp rebuild --jpeg_root=C:\libjpeg-turbo64 --with_jpeg` 
`npm install --jpeg_root=C:\libjpeg-turbo64 --with_jpeg` 

Tested on win8.1 x64 node 6.2.1, Visual Studio 2015

### DIFF
--- a/binding.gyp
+++ b/binding.gyp
@@ -149,7 +149,7 @@
                 '<(jpeg_root)/include'
               ],
               'libraries': [
-                '-l<(GTK_Root)/lib/jpeg.lib'
+                '-l<(jpeg_root)/lib/jpeg.lib',
               ]
             }, {
               'libraries': [
--- a/binding.gyp
+++ b/binding.gyp
@@ -3,11 +3,21 @@
     ['OS=="win"', {
       'variables': {
         'GTK_Root%': 'C:/GTK', # Set the location of GTK all-in-one bundle
-        'jpeg_root%': 'c:/libjpeg-turbo64',
         'with_jpeg%': 'false',
         'with_gif%': 'false',
         'with_pango%': 'false',
-        'with_freetype%': 'false'
+        'with_freetype%': 'false',
+        'variables': { # Nest jpeg_root to evaluate it before with_jpeg
+          'jpeg_root%': '<!(node ./util/win_jpeg_lookup)'
+        },
+        'jpeg_root%': '<(jpeg_root)', # Take value of nested variable
+        'conditions': [
+          ['jpeg_root==""', {
+            'with_jpeg%': 'false'
+          }, {
+            'with_jpeg%': 'true'
+          }]
+        ]
       }
     }, { # 'OS!="win"'
       'variables': {
--- a/binding.gyp
+++ b/binding.gyp
@@ -3,6 +3,7 @@
     ['OS=="win"', {
       'variables': {
         'GTK_Root%': 'C:/GTK', # Set the location of GTK all-in-one bundle
+        'jpeg_root%': 'c:/libjpeg-turbo64',
         'with_jpeg%': 'false',
         'with_gif%': 'false',
         'with_pango%': 'false',
@@ -138,6 +139,15 @@
           ],
           'conditions': [
             ['OS=="win"', {
+              'copies': [{
+                'destination': '<(PRODUCT_DIR)',
+                'files': [
+                  '<(jpeg_root)/bin/jpeg62.dll',
+                ]
+              }],
+              'include_dirs': [
+                '<(jpeg_root)/include'
+              ],
               'libraries': [
                 '-l<(GTK_Root)/lib/jpeg.lib'
               ]
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "canvas",
   "description": "Canvas graphics API backed by Cairo",
-  "version": "1.5.0",
+  "version": "1.5.1",
   "author": "TJ Holowaychuk <tj@learnboost.com>",
   "contributors": [
     "Nathan Rajlich <nathan@tootallnate.net>",
--- a/test/image.test.js
+++ b/test/image.test.js
@@ -9,14 +9,40 @@ var Canvas = require('../')
 
 var png_checkers = __dirname + '/fixtures/checkers.png';
 var png_clock = __dirname + '/fixtures/clock.png';
+var jpg_face = __dirname + '/public/face.jpeg';
 
 describe('Image', function () {
+  this.timeout(5000);
   it('should require new', function () {
     assert.throws(function () { Image(); }, TypeError);
   });
 
   it('Image', function () {
     assert.ok(Image instanceof Function);
+  });
+
+  it('Image set src to JPEG', function () {
+    var img = new Image
+      , onloadCalled = 0;
+
+    assert.strictEqual(null, img.onload);
+    assert.strictEqual(false, img.complete);
+
+    img.onload = function () {
+      onloadCalled += 1;
+      assert.strictEqual(img.src, jpg_face);
+      assert.strictEqual(485, img.width);
+      assert.strictEqual(401, img.height);
+      assert.strictEqual(true, img.complete);
+    };
+
+    img.onerror = function (e) {
+      console.error("ERROR", e); // temporary...
+    };
+
+    img.src = jpg_face;
+    assert.strictEqual(1, onloadCalled);
+    assert.strictEqual(img.src, jpg_face);
   });
 
   it('Image#onload', function () {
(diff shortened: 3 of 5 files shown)


binding.gyp
```diff
@@ -138,6 +139,15 @@
...
+                'destination': '<(PRODUCT_DIR)',
+                'files': [
+                  '<(jpeg_root)/bin/jpeg62.dll',
+                ]
+              }],
+              'include_dirs': [
+                '<(jpeg_root)/include'
+              ],
               'libraries': [
                 '-l<(GTK_Root)/lib/jpeg.lib'
```
zbjornson: Maybe I've botched my GTK dir, but I don't have a (GTK_Root)/lib/jpeg.lib, so I get a linker error. Did you add this file manually?

#458 changes it to `'-l<(jpeg_root)/lib/jpeg-static.lib'`, but that does not build.

If I change it to `'-l<(jpeg_root)/lib/jpeg.lib'`, it builds, but see comment in main discussion.

zbjornson:
Glad to see someone working on jpeg support for Windows, thanks!

I'm not sure this fully works. Per the comment [here](https://github.com/libjpeg-turbo/libjpeg-turbo/issues/45#issuecomment-181690889), apparently `jpeg_stdio_src` is not available from jpeg62 (you're copying jpeg62.dll to the destination, but linking against jpeg.lib). Currently there's no test for the code that uses `jpeg_stdio_src`, and if I add one that I think should work, node hangs.

Could you please add this test to see if the CI builds it on linux?

```diff
--- a/test/image.test.js
+++ b/test/image.test.js
@@ -9,6 +9,7 @@ var Canvas = require('../')

 var png_checkers = __dirname + '/fixtures/checkers.png';
 var png_clock = __dirname + '/fixtures/clock.png';
+var jpg_face = __dirname + '/public/face.jpeg';

 describe('Image', function () {
   it('should require new', function () {
@@ -40,6 +41,31 @@ describe('Image', function () {
     assert.strictEqual(320, img.height);
   });

+  it('Image set src to JPEG', function () {
+    var img = new Image
+      , onloadCalled = 0;
+
+    assert.strictEqual(null, img.onload);
+    assert.strictEqual(false, img.complete);
+
+    img.onload = function () {
+      onloadCalled += 1;
+      assert.strictEqual(img.src, jpg_face);
+    };
+
+    img.onerror = function (e) {
+      console.error("ERROR", e); // temporary...
+    };
+
+    img.src = jpg_face;
+    assert.strictEqual(1, onloadCalled);
+    assert.strictEqual(img.src, jpg_face);
+
+    assert.strictEqual(true, img.complete);
+    assert.strictEqual(320, img.width);
+    assert.strictEqual(320, img.height);
+  });
+
   it('test Image#onload multiple times', function() {
     var img = new Image
       , onloadCalled = 0;
```

zbjornson:
Also:

> The explicit options make sure nothing breaks.

Hmm, is there a case when having libjpegturbo installed and findable would break? It's nice to not have to pass a flag to add jpeg support (a la the script in #458). If anyone is installing canvas as something other than a primary dependency (e.g. they depend on package that depends on canvas), it'll be hard to pass the flag along.

binding.gyp
timknip: @zbjornson You're right, `jpeg.lib` isn't in GTK, so need the one from `libjpeg-turbo`.

That compiles. But... there's a nasty issue.

`img.src = fs.readFileSync('foo.jpg')` works
`img.src = foo.jpg` doesn't work

The problem is that libjpeg-turbo was compiled against `msvcrt.lib`, whilst Visual Studio uses `msvcrt{VS_VERSION}.lib`... Apparently this makes `jpeg_read_header` hang / crash... See http://stackoverflow.com/questions/4092251/alternatives-to-jpeg-read-header-libjpeg

I'll try fix `Image.cc` ... but in any case: it seems more involved then previously thought ;-)

timknip:
@zbjornson 

Latest commit includes your patch for `image.test.js`
`npm test --with_jpeg=true` passes. 

> apparently jpeg_stdio_src is not available from jpeg62 

I edited `Image.cc` in a way that `jpeg_stdio_src` is never used on windows. When loading a jpeg from file `Image::loadJPEGFromBuffer` is used. The hit is that `CAIRO_VERSION_MINOR>=10` must be used. Luckily the Cairo shipped with `GTK2` is version 10.

> it'll be hard to pass the flag along.

Agree... Do you have a suggestion? Its possible of course to check for `libjpeg-turbo` using some node script.

binding.gyp
timknip: Worked around the `img.src ="foo.jpg"` issue... See below.

timknip:
@zbjornson Ok reworked `Image.cc` a bit. See https://github.com/floorplanner/node-canvas/blob/master/src/Image.cc#L878-L920 Its simply using `Image::loadJPEGFromBuffer` when loading from file.

Furthermore I used the `win_jpeg_lookup.js` in `binding.gyp` from #458 to check for `c:\libjpeg-turbo` or `c:\libjpeg-turbo64`.

Seems the only difference with #458 is usage of `jpeg.lib` instead of `jpeglib-static.lib` and my changes to `Image.cc` to not use `jpeg_stdio_src ` on windows.

zbjornson:
:+1: Good solution. Compiles for me on Win 7 / VS2015 and tests pass. Browser JPEG tests have a black background but, if I remember, that's normal... Thanks!

zbjornson:
@LinusU when you're free would you mind taking a look at this please? It looks good to me, is semver-minor and would be super nice to finally get settled. ;)

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "canvas",
   "description": "Canvas graphics API backed by Cairo",
-  "version": "1.5.0",
+  "version": "1.5.1",
```
LinusU: Please don't bump the version number, I'll do that as part of the release process

test/image.test.js
```diff
@@ -19,6 +21,30 @@ describe('Image', function () {
...
+    img.onload = function () {
+      onloadCalled += 1;
+      assert.strictEqual(img.src, jpg_face);
+      assert.strictEqual(485, img.width);
+      assert.strictEqual(401, img.height);
+      assert.strictEqual(true, img.complete);
+    };
+
+    img.onerror = function (e) {
+      console.error("ERROR", e); // temporary...
```
LinusU: How about `onerrorCalled = 0`, `onerrorCalled += 1` and `assert.strictEqual(onerrorCalled, 0)`?

test/image.test.js
```diff
@@ -9,8 +9,10 @@ var Canvas = require('../')
 
 var png_checkers = __dirname + '/fixtures/checkers.png';
 var png_clock = __dirname + '/fixtures/clock.png';
+var jpg_face = __dirname + '/public/face.jpeg';
 
 describe('Image', function () {
+  this.timeout(5000);
```
LinusU: This is quite a long time, is it really needed?

test/image.test.js
timknip: @LinusU Missed your post ;-) I'll prep a fresh PR based on latest master when I find some time. Probably somewhere later this week.

LinusU: Thank you 💌

chearon:
just curious, how come there need to be changes to the way JPEGs are loaded from a file on Windows? I was able to get JPEG support on Windows by only [modifying](https://github.com/chearon/node-canvas-prebuilt/blob/master/ci/win/binding.gyp) the build and not the source, using MSYS2

zbjornson:
@chearon see reason in my first comment and code comment. To me, the minor change to the source is preferable to changing the build chain to use msys, which is a much less common tool set. (Although your precompiled binaries make that invisible for users of them.)

chearon:
Oh sorry, did not realize that was the entire discussion...interesting problem, and it seems like the MSYS2 build would have the same problem since it links with `msvcrt`, but I can't figure out [how to tell] which C library node is using (is it statically built in?)

zbjornson:
No worries :). MSVCRT is statically linked to node, yes. (I think this is the same problem that fs-ext is facing ([a](https://github.com/nodejs/node-gyp/issues/416), [b](https://github.com/nodejs/node/issues/6369), [c](https://github.com/baudehlo/node-fs-ext/issues/57)).)

timknip:
@LinusU @zbjornson Please check #841. I'm closing this one.

PR #817, opened 2016-10-02
Remove unused member variables warned by compiler

rhysd:
Hi, thank you for nice library.  I'm testing my `<canvas>` application with this and this works great.

I found that some fields were reported as unused by Clang.  So I removed them safely and confirmed all tests passed.  I also confirmed my application tests pass with my fork.


LinusU:
Nice and clean, thank you!

PR #820, opened 2016-10-07
Convert has_lib.sh to sh so it also works on BSD OSes

hakkiplaten:
Passes http://www.shellcheck.net/

Ref. #754

### DIFF
--- a/util/has_lib.sh
+++ b/util/has_lib.sh
@@ -1,18 +1,19 @@
-#!/bin/bash
+#!/bin/sh
 
 has_ldconfig() {
   hash ldconfig 2>/dev/null
 }
 
 has_system_lib() {
-  local regex="lib$1.+(so|dylib)"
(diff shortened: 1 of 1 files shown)


util/has_lib.sh
```diff
@@ -52,7 +53,7 @@ case "$1" in
     result=$?
     ;;
   *)
-    >&2 echo "Unknown library: $1"
+    echo "Unknown library: $1"
```
LinusU: Is there no way to print to stderr in `sh`?

hakkiplaten: Was a bug at ShellCheck. That line is valid. I've fixed the PR.

hakkiplaten:
@LinusU all good to merge?

LinusU:
Thank you for this

LinusU:
Backporting this to 1.x in #826

LinusU:
Published as 1.6.1

PR #571, opened 2015-06-28
Support for FBDev, statically linked library and musl

piranna:
This pull-request is work in progress, but it works :-) I've open early to discuss about how to integrate it. It has the next features:
- updated dependencies, both npm and libraries. This last ones are only done on their current version branch to prevent incompatibilies on newer ones that would require modify the code.
- allow to use FBDev as backend, so node-canvas can be used to draw on the console. This is done thanks to the work of @ReneHollander at https://github.com/ReneHollander/node-canvas, that also added some basic support for a selectable backend support.
- compile it as a statically linked library without dependencies on any external dynamic library. This is based on the work of @magicode at https://github.com/magicode/node-canvas, although I needed to do some fix-ups due to undefined references.
- musl support as C lib instead of glibc, that since don't support lazy linking is the reason why I need to fix the undefined references.

All of this makes it possible to use node-canvas to draw on the console on [NodeOS](https://github.com/NodeOS/NodeOS).

First step will to update the code with current master, and later improve the FbDev support and the alternatives backends, probably making them selectable at compile time.

PR #829, opened 2016-10-26
Backends support

piranna:
This is a stripped-down version of pull-request #571 focused only on the
backends support, but adding also support for PDF and SVG backends and
passing the tests. it also retain backwards compatibility with the old API,
so this would be easy to merge since it's mostly to move around the PDF and
SVG code and checks to independent classes. There's some code for
`pdfStream` and similar methods that I think they should be moved too since
they are mostly backend specific, and make more sense there. I didn't do
that since it would break the API, but maybe it would be added some
functions on the Canvas object that proxy the petitions to the backend to
retain compatibility.

I know it's a big pull-request, so I'm open for comments over it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ node_modules
 # Vim cruft
 *.swp
 *.un~
+npm-debug.log
--- a/binding.gyp
+++ b/binding.gyp
@@ -2,11 +2,11 @@
   'conditions': [
     ['OS=="win"', {
       'variables': {
-        'GTK_Root%': 'C:/GTK', # Set the location of GTK all-in-one bundle
+        'GTK_Root%': 'C:/GTK',  # Set the location of GTK all-in-one bundle
         'with_jpeg%': 'false',
         'with_gif%': 'false'
       }
-    }, { # 'OS!="win"'
+    }, {  # 'OS!="win"'
       'variables': {
         'with_jpeg%': '<!(./util/has_lib.sh jpeg)',
         'with_gif%': '<!(./util/has_lib.sh gif)'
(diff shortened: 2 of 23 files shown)


piranna:
Any update on this one?

cammanderson:
This seems promising! :+1:

piranna:
> This seems promising!

I'm glad you liked it, let's hope if it gets merged soon :-)

ianmetcalf:
@piranna I couldn't get the example to work, `Canvas.backends` was undefined. I got it to work by adding the initialization of `backends` on the target object in init.cc

```js
NAN_MODULE_INIT(init) {
  Backends::Initialize(target);
  Canvas::Initialize(target);
  Image::Initialize(target);
```

https://github.com/ianmetcalf/node-canvas/blob/27a48c66f8ad1cd642377d5afd74d230efbab0aa/src/init.cc#L28

piranna:
You are right, I missed it :-P It doesn't matter too much since this
pull-request just prepare the backends mechanism and only offer the current
internal ones, so the current API doesn't change. When this got merged then
I'll move forward other ones from my previous pull-request to add support
for fbdev and X11, and in that case the Backends namespace will be needed.

ianmetcalf:
I just pointed it out since this PR includes `examples/backends.js` which uses this new API. It would probably be good to either update `init.cc` or change the example to use the exiting API.

I really appreciate the work you did abstracting the backends. I'm currently working with the FBDev backend that I extracted from your master branch.

piranna:
> I just pointed it out since this PR includes examples/backends.js which uses this new API. It would probably be good to either update init.cc or change the example to use the exiting API.

You are right. Since it's the new behaviour, I've added the missing namespace.


> I really appreciate the work you did abstracting the backends. I'm currently working with the FBDev backend that I extracted from your master branch.

Well, I have just only updated some previous code to use it with static builds for NodeOS :-P So why are you using it? Are you using it on production?

piranna:
Any update on this? Can we merge it?

piranna:
Ping?

LinusU:
Awesome work! Sorry for the long delay 🎉

piranna:
Thank you for merging :-)

ReneHollander:
I am a bit saddened by the fact, that there is no mention of my work on this. Just found out through coincidence, that this got merged two years after I made my first commit to my fork.

chearon:
You were mentioned in the original PR #571

piranna:
@ReneHollander you are right, the foundational code of this pull-request is based on your work, without it I would have not been able to know how to do it, thanks :-)

piranna:
I hope now that this got merged, if static builds support gets merged too, since now Github better support pull-requests merges without need to do rebases that we could re-use the previous pull-request as a container for several of the backends, so your work would have a direct repercusion on the node-canvas code :-)

Issue #832, opened 2016-10-30
Is it time to drop Node.js 0.10 support (in 2.x)


LinusU:
Official support for Node.js 0.10 will cease after tomorrow. We currently have a few issues with 0.10 as well, which currently hinders the merging of #740 and #808.

Would anyone be sad by seeing 0.10 support go?

Is it, as module authors, even good to continue to support an end of life version of Node?

Any input is appreciated!

_(note that this will only happen in a major version, v 1.x will continue to work with 0.8 and 0.10 forever)_

zbjornson:
#740 works in 0.10 actually. I think #808 is the only one that is borked in 0.10, but I also think #808 does not add any *new* failures; I think the error happens only if you access the prototype, which is already broken.

In any case, my :+1: indicates support because I think it's good to discourage use of EOL runtimes, but not sure if it's necessary to drop support given the above.

zbjornson:
In any case, think this part of the readme can be removed :laughing:: 

> Tested with and designed for:

> node 0.4.2
> cairo 1.8.6
> For node 0.2.x node-canvas <= 0.4.3 may be used, 0.5.0 and above are designed for node 0.4.x only.

chearon:
I found [this tweet](https://twitter.com/seldo/status/781258069678391296) from the NPMer who posts stats sometimes. Looks like maybe it's a bit early, I haven't used 0.10 in a long long time though

> In any case, think this part of the readme can be removed

I noticed that the other day too!😆

jdalton:
There's a more recent version [here](https://twitter.com/seldo/status/793264271417184256). There's also some other [things to consider](https://twitter.com/thealphanerd/status/794253960374788096) on the numbers. Accounting for that it looks like 0.10 use is < 10% now.

LinusU:
Some interesting discussion going on here related to that: travis-ci/travis-ci#6659

LinusU:
Here are some more recent data, I feel that it's time to drop both Node.js 0.10 and also Node.js 0.12 since 0.12 has a lower adoption rate then 0.10. Both is end of life and there doesn't seem to be that many people using it...

https://docs.google.com/spreadsheets/d/1J6TxXIfGOqV9cx1qEFGSXHCma4JHfKPzlsMWfbGmDbE/edit#gid=0

jdalton:
@LinusU Where did you get the Node usage data from?

LinusU:
[...]

PR #849, opened 2016-11-25
update supported node versions

chearon:
discussion in #832 

Seems doubtful there'd ever be Cairo version issues, or issues with new node versions, so I removed those. And just put "minimum 0.10" to the top near where it says to use Pango/Cairo

Or it could be done any other way, just wanted to initiate this


LinusU:
Thanks

Issue #873, opened 2017-02-22
Canvas#toDataURL(fn) isn't async when there's no canvas data.


lukechilds:
I'm writing a module on top of node-canvas and it expects the image to be returned async in the callback. This works fine unless the user hasn't written any data to the canvas. In that case the function returns `data:` instantly and never executes the callback.

e.g:
```js
> canvas.toDataURL((err, png) => console.log('foo'))
'data:,'
```

I think the image should always be returned async if a function is passed in.

PR #874, opened 2017-02-22
Make sure Canvas#toDataURL is always async if a callback is passed in

lukechilds:
Resolves https://github.com/Automattic/node-canvas/issues/873

With unit test :)

Old behaviour: returns the string as the function return value and never executes the callback

```js
> new Canvas().toDataURL(function(err, str){
...   console.log('hi');
...   console.log(str);
... });
'data:,'
```

New behaviour: function instantly returns undefined and callback is executed async.

```js
> new Canvas().toDataURL(function(err, str){
...   console.log('hi');
...   console.log(str);
... });
undefined
hi
data:,
```

It might be worth returning `'data:,'` both sync (instead of undefined) and async for backwards compatibility incase someone is handling this edge case.

### DIFF
--- a/lib/canvas.js
+++ b/lib/canvas.js
@@ -231,11 +231,6 @@ Canvas.prototype.toDataURL = function(a1, a2, a3){
   // ['image/jpeg', qual, fn] -> ['image/jpeg', {quality: qual}, fn]
   // ['image/jpeg', undefined, fn] -> ['image/jpeg', null, fn]
 
-  if (this.width === 0 || this.height === 0) {
-    // Per spec, if the bitmap has no pixels, return this string:
-    return "data:,";
-  }
-
   var type = 'image/png';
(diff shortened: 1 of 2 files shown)


lukechilds:
Yeah actually, changing return values probably isn't a good idea, I've updated it so the function now returns `'data:,'` like before this PR instead of `undefined`. However, now the callback is properly executed as well.

lukechilds:
@LinusU Bumpety bump bump :)

I'm not sure who maintains this, I think you?

I'm building a module on top of node-canvas, it would be great if this could be merged so I can continue working on it.

Let me know if you need me to do anything else :)

LinusU:
Looks great, thank you!

LinusU:
Released as 1.6.4

lukechilds:
@LinusU Thanks!

Issue #566, opened 2015-06-17
the property font is not compatiable with HTML5 canvas font 


zwzw911:
HTML5 canvas font accept: font-style/font-variant/font-weight/font-size/font-family, for example, context.font="italic small-caps bold 12px arial", but node-canvas only accept  font-style/font-size/font-family, it set font-variant or font-weight, the property font change to default value "10px sans-serfi"

keverw:
Interesting. I posted about this here https://github.com/Automattic/node-canvas/issues/53#issuecomment-151760247

but noticed bold isn't working in my case. I guess it's not supported yet?

kaivi:
Yep, all options other than font-size and font-family are being ignored by node-canvas. When one specifies both font-style and font-variant, context.font gets reset to a default one:

## Chrome 48.0.2564.116 on OS X
![chrome-osx](https://cloud.githubusercontent.com/assets/3091192/13350427/4858a70a-dc80-11e5-9c4a-5913e4a9684d.png)

## canvas v1.3.10 on OS X
![node-canvas-osx](https://cloud.githubusercontent.com/assets/3091192/13350426/48566364-dc80-11e5-9fb9-fc294ffd40ac.png)

Here is [an in-browser JSFiddle](https://jsfiddle.net/rggzfyzf/3/), this code attempts to auto-fit font size to fill a rectangle.

yever:
@zwzw911, the issue with not parsing the font style correctly (especially when it includes the variant) was fixed in #891. In the meantime, I can't get `pango` to actually render small-caps. Most fonts don't support it and I'm not sure that `pango` is capable of using this feature when it does exist.  I opened an issue for this: #894.

I think that the "font reset" problem is fixed, so should this issue be closed now?

jleclanche:
I'm not sure if [line-height](https://developer.mozilla.org/en/docs/Web/CSS/line-height) (`16px/1em`) is being taken into account right now, I don't think it is, but it doesn't error anymore in the 2.0 alpha, so yay. :)

chearon:
There is no line height on canvas since there's no layout functionality in the font APIs, it just draws one line of text. It might parse OK but would be ignored.

zbjornson:
I think since #944 landed, this is as good as we can get. Please comment if you find any other discrepancies.

PR #891, opened 2017-03-17
Parse font using parse-css-font and units-css

yever:
This pull request address issue #566 

My issue was in using `node-canvas` to test code that uses `pixi.js`, which always specifies fully the style, variant and weight in ``ctx.font``, and currently it falls back to the default size.

The problem seems to be in the parsing of the font style. I found the modules [parse-css-font](https://www.npmjs.com/package/parse-css-font) and [units-css](https://www.npmjs.com/package/units-css) that provide functions to parse the font style and could be used as a drop in replacement to the current code. Both modules are shipped with the MIT license, so I don't think that this dependency should be an issue.

The `font-variant` property is currently not passed to the `c` code and not processed by `pango`. I managed to implement this, but still couldn't get `pango` to render "small-caps" (see my branch [font-variant](https://github.com/yever/node-canvas/tree/font-variant)). In the meantime, I am only sharing in this pull-request the solution for the font style parsing issue.

### DIFF
--- a/lib/context2d.js
+++ b/lib/context2d.js
@@ -16,6 +16,10 @@ var canvas = require('./bindings')
   , CanvasPattern = canvas.CanvasPattern
   , ImageData = canvas.ImageData;
 
+var parseCssFont = require('parse-css-font');
+
+var unitsCss = require('units-css');
+
 /**
  * Export `Context2d` as the module.
  */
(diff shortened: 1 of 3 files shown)


chearon:
Always hated that regex, appreciate this! Have you tried creating a test program in C which calls Pango and asks for a small-caps font? The example code [here](https://developer.gnome.org/pango/stable/pango-Cairo-Rendering.html) is easy to modify to test things like that... I have found a lot of holes in support for less-used features like this in OSX and Windows. Or it might only work on TTFs that identify as a small caps variant, I don't know if Pango will do the "fake" small caps that browsers do

LinusU:
>  I wasn't sure whether there's some policy in this project against using external packages.

I'm all for using external packages 👍 the less code we have the better 🎉

LinusU:
Released as v1.6.5

yever:
Thank you! :)

I'll create a new issue for the problem with small-caps in pango

PR #906, opened 2017-04-22
Support Chocolaty deps install instructions

nicekiwi:


Readme.md
```diff
@@ -39,7 +39,7 @@ OS X | `brew install pkg-config cairo pango libpng jpeg giflib`
 Ubuntu | `sudo apt-get install libcairo2-dev libjpeg8-dev libpango1.0-dev libgif-dev build-essential g++`
 Fedora | `sudo yum install cairo cairo-devel cairomm-devel libjpeg-turbo-devel pango pango-devel pangomm pangomm-devel giflib-devel`
 Solaris | `pkgin install cairo pango pkg-config xproto renderproto kbproto xextproto`
-Windows | [Instructions on our wiki](https://github.com/Automattic/node-canvas/wiki/Installation---Windows)
+Windows | `choco install -y python2 gtk-runtime libjpeg-turbo microsoft-build-tools`  [More instructions on our wiki](https://github.com/Automattic/node-canvas/wiki/Installation---Windows)
```
zbjornson: I'm not sure how many people have chocolaty installed, so this might create more confusion. You might make this "Windows (with Cholcolaty) | `choco ...`" and "Windows | [More instructions on our wiki](...)", or put this info in the wiki only.

nicekiwi: Hmm, I would argue the same for HomeBrew, but good point. Maybe just leave it in the Wiki. :)

PR #907, opened 2017-04-24
Image.getRawData method

basvanmeurs:
Background info: WPE UI Framework depends on node-canvas for creating text bitmaps, rounded edges and image parsing.

When you need to read parsed raw ARGB image data, you need to load the image, then draw it on a canvas and convert it to a buffer:

```javascript
        var img = new Canvas.Image();
        img.src = data;
        var canvas = new Canvas(img.width, img.height);
        ctx.drawImage(img, 0, 0, img.width, img.height);
        var buffer = canvas.toBuffer('raw');
```

This is an unnecessary double copy of the data. This pull request contains a new method: Image.getRawData, which returns the buffer directly:
```javascript
        var img = new Canvas.Image();
        img.src = data;
        var buffer = img.getRawData();
```
This allows for a more efficient usage of Canvas.Image.

We'd appreciate it if you'd merge this into the node-canvas module!

### DIFF
--- a/src/Image.cc
+++ b/src/Image.cc
@@ -45,7 +45,7 @@ Image::Initialize(Nan::ADDON_REGISTER_FUNCTION_ARGS_TYPE target) {
 
   // Prototype
   Local<ObjectTemplate> proto = ctor->PrototypeTemplate();
-  Nan::SetPrototypeMethod(ctor, "getRawData", GetRawData);
+  Nan::SetAccessor(proto, Nan::New("rawData").ToLocalChecked(), GetRawData);
   Nan::SetAccessor(proto, Nan::New("source").ToLocalChecked(), GetSource, SetSource);
   Nan::SetAccessor(proto, Nan::New("complete").ToLocalChecked(), GetComplete);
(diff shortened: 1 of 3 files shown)


lib/context2d.js
```diff
@@ -16,10 +16,6 @@ var canvas = require('./bindings')
   , CanvasPattern = canvas.CanvasPattern
   , ImageData = canvas.ImageData;
 
-var parseCssFont = require('parse-css-font');
```
LinusU: The removal of this should be it's own PR, it also should remove the dependencies from the package.json...

basvanmeurs: I didn't realise that it would be part of the pull request.. that was not intentional.

zbjornson:
@basvanmeurs if you still want this to be merged, can you clean up the branch so it contains only the relevant changes please per @LinusU's feedback?

zbjornson:
(I guess this is replaced by #1256)

basvanmeurs:
Correct, it can be closed ;-) Thanks zbjornson!

PR #917, opened 2017-05-04
Fix 2.0 regression, warnings, v8 memory dealloc hint

zbjornson:
Handful of little fixes:

1. `nBytes` was wrong as of 1a6dffe, but the tests didn't catch it because they used a square canvas.
1. Allow alpha and beta version in the test assertions.
1. Minor: fix shadowing warnings during compilation.
1. Fix v8 memory dealloc hint. When this function was run, it would have actually hinted to v8 that a massive amount of external memory was being used.


LinusU:
Awesome work 👍

LinusU:
Published as 2.0.0-alpha.2

PR #926, opened 2017-06-20
Fix memory leak in Canvas constructor

### LINKED ISSUE #922: Memory leak in version 2.0
Hi!

I have come across what seems to be a memory leak in `node-canvas` version 2. I've put together a very basic repro case that you can find in [this repo](https://github.com/azangru/node-canvas-testing).

Briefly, the problem manifests like this:

- Start an app that uses node-canvas v.2
- Send multiple requests to it
- Observe the memory usage. Specifically, the `rss` portion of `process.memoryUsage` is quickly increasing despite garbage collection.

For example, in my repro case, the memory usage that started with:

```
{"rss":72290304,"heapTotal":21008384,"heapUsed":11065680}
```

rose after a 1000 requests to:

```
{"rss":643870720,"heapTotal":17862656,"heapUsed":11048840}
```

(notice the rise in `rss`).

This memory leak is not present in the 1.x branch of `node-canvas`.

I tried both `.pngStream()` and `.toBuffer()` methods in my repro case (since there is an issue regarding memory leaks with `pngStream`). None of this helped.

If you could suggest the place in the version 2 branch that is free from this leak, I would be super grateful. I've used the master branch of `node-canvas` in production this winter, and it seems like there was no such bug then.

**My environment:**
MacOS v. 10.11.6
node v6.7.0

**P.S.:** I have not bisected the git history to find the exact commit that has introduced the leak, but at least at the time of [this March commit](https://github.com/Automattic/node-canvas/tree/3d545056e2bb61b25f2d8bb1445c2e424a615163) memory seems to be managed fine.

### LINKED ISSUE #925: External memory usage hinting is wrong
Found while working on #922... 

Before https://github.com/Automattic/node-canvas/commit/198a61df1bd090a79cb63d6be2f0b1d27c13cf8e, external memory usage (`process.memoryUsage().external`) would become negative after drawing a large number of canvases.

With that commit where I fixed that hint, it now supposedly grows unboundedly. That's also not true.

image

Somewhere, something isn't calling `AdjustExternalMemory` where it should.

(AFAIK this is just a hint to v8 and only affects when the GC runs.)

zbjornson:
Fixes #922 
Fixes #925

@azangru this is simple enough to repro the leak that you reported (run with `node --expose-gc`):

```js
var Canvas = require('./index.js');
var result;
for (var k = 0; k < 2000; k++) {
	result = new Canvas(1200, 600 /*, "pdf" */);
	if (k % 50 === 0) console.log(k, process.memoryUsage().rss);
	gc();
}
```

https://github.com/Automattic/node-canvas/commit/48eb938b70ce593a225a54e49c6e307694cb19f9 created orphaned Cairo surfaces for all three types of backends (PDF, SVG, image).

This also removes some redundant work because both the backend constructors and Canvas constructor were invoking `createSurface`.

### DIFF
--- a/src/Canvas.cc
+++ b/src/Canvas.cc
@@ -617,7 +617,6 @@ NAN_METHOD(Canvas::RegisterFont) {
 
 Canvas::Canvas(Backend* backend) : ObjectWrap() {
   _backend = backend;
-  this->backend()->createSurface();
 }
 
 /*
--- a/src/backend/ImageBackend.cc
+++ b/src/backend/ImageBackend.cc
@@ -17,9 +17,10 @@ ImageBackend::~ImageBackend()
 
 cairo_surface_t* ImageBackend::createSurface()
 {
+	assert(!this->surface);
 	this->surface = cairo_image_surface_create(CAIRO_FORMAT_ARGB32, width, height);
 	assert(this->surface);
-  Nan::AdjustExternalMemory(4 * width * height);
+	Nan::AdjustExternalMemory(4 * width * height);
 
 	return this->surface;
 }
@@ -34,7 +35,6 @@ cairo_surface_t* ImageBackend::recreateSurface()
 
 	return createSurface();
 }
-
 
 Nan::Persistent<FunctionTemplate> ImageBackend::constructor;
 
--- a/src/backend/PdfBackend.cc
+++ b/src/backend/PdfBackend.cc
@@ -13,6 +13,8 @@ using namespace v8;
 PdfBackend::PdfBackend(int width, int height)
 	: Backend("pdf", width, height)
 {
+	_closure = malloc(sizeof(closure_t));
+	assert(_closure);
 	createSurface();
 }
 
@@ -28,8 +30,6 @@ PdfBackend::~PdfBackend()
 
 cairo_surface_t* PdfBackend::createSurface()
 {
-	_closure = malloc(sizeof(closure_t));
-	assert(_closure);
 	cairo_status_t status = closure_init((closure_t*)_closure, this->canvas, 0, PNG_NO_FILTERS);
 	assert(status == CAIRO_STATUS_SUCCESS);
 
--- a/src/backend/SvgBackend.cc
+++ b/src/backend/SvgBackend.cc
@@ -15,7 +15,6 @@ SvgBackend::SvgBackend(int width, int height)
 {
 	_closure = malloc(sizeof(closure_t));
 	assert(_closure);
-
 	createSurface();
 }
 


zbjornson:
@piranna question for you on the new backend constructors. They're not documented right now and I don't fully understand the use cases. In particular, I'm not sure what appropriate behavior would be in this scenario:

```js
var ib = new Canvas.backends.ImageBackend(50,50);
var c1 = new Canvas(ib);
var c2 = new Canvas(ib);
```

Before this PR, the behavior is rather undefined because a new surface is created in the Backend when each Canvas instance is created. With this PR, there will be an assertion failure for ImageBackend and the same old behavior (including a memory leak) for Svg and PdfBackend. (I could add the same assertion to those backends to protect against the leak but I assume it's very unusual for a user to do this.) Is there a desired behavior?

piranna:
> var ib = new Canvas.backends.ImageBackend(50,50);
> var c1 = new Canvas(ib);
> var c2 = new Canvas(ib);

Regarding to that code sample, I would expect to have two canvas painting on the same backend, in this case the same underlying Cairo surface of the image. Is this what you were talking about?

zbjornson:
OK, thanks. I'll see if I can change this fix to support that behavior then (unless that's not a real use case).

LinusU:
> OK, thanks. I'll see if I can change this fix to support that behavior then (unless that's not a real use case).

@zbjornson do you want me to merge this now or should I hold off?

piranna:
> unless that's not a real use case

It's more like a corner case, people should not be doing this kind of things, but I would not prevent them to do it so. Ey, maybe they would like to render several screen regions from different threads, for example on a multiplayer game with split screen :-)

zbjornson:
@LinusU I'll make the change on Monday, so please wait 'till then :)

zbjornson:
Alright, should be good to go. Tests pass and I don't see the memory leak.

> maybe they would like to render several screen regions from different threads,

@piranna I don't know anything about thread safety in cairo, so no guarantees, but theoretically this change will prevent clobbering when multiple Canvas instances use the same Backend instance.

piranna:
What do you mean by clobbering?

LinusU:
Thanks!

zbjornson:
> What do you mean by clobbering?

@piranna I mean that in this case:

```js
var ib = new Canvas.backends.ImageBackend(50,50);
var c1 = new Canvas(ib);
// draw stuff
var c2 = new Canvas(ib);
```

before this PR and with the initial version of this PR, whatever was put on the surface by `c1` would have been destroyed when `c2 = new Canvas(ib)` was executed. The version of this PR that got merged doesn't destroy the surface when the second Canvas is constructed.

piranna:
Oh, that's cool! :-D

Issue #833, opened 2016-10-30
Move non-standard API out of the Canvas class


LinusU:
I would like to move all of the non-standard APIs out from the `Canvas` class, and make the `Canvas` class follow the specification to the letter.

The benefit of this would be to be able to provide a `browser.js` file with a small shim for out specific APIs, which we could also make in the browser. Then you would be able to write truly isomorphic code that works both on Node.js and in the browser, which I think would be awesome 🙌 

It would also allow us to ship some better image loading apis out of the box. The `onload` and `onerror` properties seems quite dated.

I'm thinking something like this, but are very open to suggestions:

`createCanvas([width[, height[, format]]]) => Canvas`

Creates and return a `Canvas` with the specified width and height. `format` is for pdf generation which would not work in the browser.

`loadImage(path[, options]) => Promise<Image>`

Creates and returns a `Promise` of an `Image`. This would be a shorthand for creating an `Image` manually, setting the `src` and hooking into the `load` and `error` events.

The `options` object could be used for setting the `dataMode`, which would be a no-op in the browser.

`imageFromBuffer(buffer) => Image`

Create an image from a buffer. This should create the image immediately since all the data is available and no IO needs to be performed.

`registerFont(fontPath, fontSettings) => Void`

Register a new font. This could potentially work in the browser by constructing a stylesheet that we append `@font-face`s to. Good thing we followed the `@font-face` definition as closely as we could 👍 

`createPngStream(canvas[, options]) => ReadableStream` <br>
`createJpegStream(canvas[, options]) => ReadableStream`

Not exactly sure how, or even if, these should work in the browser...

`getBuffer(canvas) => Buffer` <br>
`getPngBuffer(canvas[, options]) => Buffer` <br>
`getJpegBuffer(canvas[, options]) => Buffer`

Not exactly sure how, or even if, these should work in the browser...

Apart from these functions, we should probably export the `Image` class and maybe the `Canvas` class as well. Although the `Canvas` class is actually called `HTMLCanvasElement` in the browser and there is no way to call `new` on it.

As always, any input is appreciated 🙌

zbjornson:
> `createCanvas([width[, height[, format]]]) => Canvas`

This doesn't seem like much of a change; the usage change is simply `s/new Canvas/createCanvas`, right? What about making the parameters optional, since you cannot pass params to `document.createElement`, and allowing them to be set as properties instead?

```js
var canvas = new Canvas(); // or createCanvas
[...]

PR #929, opened 2017-06-24
Add browser compatible API

LinusU:
This PR modifies the API to a few wrapper functions that are provided both for Node.js and the browser. Using a tool like [browserify](http://browserify.org) it should be easy to write code that works both on the backend and the frontend 🙌 

ref: #833

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -1,0 +1,35 @@
+/* globals document, ImageData */
+
+const parseFont = require('./lib/parse-font')
+
+exports.parseFont = parseFont
+
+exports.createCanvas = function (width, height) {
+  return Object.assign(document.createElement('canvas'), { width, height })
+}
+
+exports.createImageData = function (array, width, height) {
+  // Browser implementation of ImageData looks at the number of arguments passed
+  switch (arguments.length) {
+    case 0: return new ImageData()
+    case 1: return new ImageData(array)
+    case 2: return new ImageData(array, width)
+    default: return new ImageData(array, width, height)
+  }
+}
+
+exports.loadImage = function (src) {
+  return new Promise((resolve, reject) => {
+    const image = document.createElement('img')
+
+    function cleanup () {
+      image.onload = null
+      image.onerror = null
+    }
+
+    image.onload = () => { cleanup(); resolve(image) }
+    image.onerror = () => { cleanup(); reject(new Error(`Failed to load the image "${src}"`)) }
+
+    image.src = src
+  })
+}
--- a/package.json
+++ b/package.json
@@ -30,12 +30,12 @@
     "test-server": "node test/server.js"
   },
   "dependencies": {
-    "assert-rejects": "^0.1.1",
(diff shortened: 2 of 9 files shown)


package.json
```diff
@@ -29,6 +30,7 @@
     "test-server": "node test/server.js"
   },
   "dependencies": {
+    "assert-rejects": "^0.1.1",
```
zbjornson: meant to be a devDependency?

LinusU: Great catch!

browser.js
```diff
@@ -0,0 +1,35 @@
...
+  return Object.assign(document.createElement('canvas'), { width, height })
+}
+
+exports.createImageData = function (array, width, height) {
+  // Browser implementation of ImageData looks at the number of arguments passed
+  switch (arguments.length) {
+    case 0: return new ImageData()
+    case 1: return new ImageData(array)
+    case 2: return new ImageData(array, width)
+    default: return new ImageData(array, width, height)
```
zbjornson: Do cases 0 and 1 exist? [MDN only documents](https://developer.mozilla.org/en-US/docs/Web/API/ImageData/ImageData) `new ImageData(arr, w, h)` and `ImageData(w, h)`.

Also, would this be the same as `return new ImageData(...arguments)`?

LinusU: `0` and `1` doesn't exist, I added them to get the errors reported back to be accurate.

screen shot 2017-06-24 at 21 31 52

`return new ImageData(...arguments)` would be perfect, but unfortunately Node.js 4.x doesn't support it... It's doing this since passing `(10, 10, undefined)` doesn't do what you want it to in the browsers:

screen shot 2017-06-24 at 21 33 28

Not 100% sure that this is the best approach though 🤔

zbjornson: Ah, clever :) and now I understand what your comment means.

zbjornson:
@LinusU I started playing more with this and realized one thing to consider changing. This PR hid the `ImageData` constructor, so you can't do `xx instanceof ImageData` or `new ImageData(...)` anymore, for example. Browsers provide `ImageData`, so I think it should still be exposed.

zbjornson:
Actually, a lot of stuff got hidden...

```js
> /* 2.0 */ Canvas.
[inherited from Object, snip]

Canvas.createCanvas          Canvas.createImageData       Canvas.loadImage             Canvas.parseFont
```

```js
> /* 1.6 */ Canvas.
[inherited from Object, snip]

[inherited from Function, snip]

Canvas.Context2d             Canvas.Image                 Canvas.ImageData             Canvas.JPEGStream
Canvas.PNGStream             Canvas.cairoVersion          Canvas.prototype             Canvas.version
```

The constructors are useful at least for `instanceof` (although `Context2d` should be `CanvasRenderingContext2d`), and `cairoVersion`/`version` have their merits as well. Maybe go back to exposing all of those, and just additionally export the `create*` and `loadImage` functions?

Issue #934, opened 2017-06-30
Move PNG_* constants to constructor, clean up toBuffer API


zbjornson:
Proposal for a breaking change in 2.0: move all the `PNG_*` constants from the Canvas prototype to be static members of the Canvas constructor. That is, this doc example:

```js
var buf2 = canvas.toBuffer(undefined, 3, canvas.PNG_FILTER_NONE);
```

would become

```js
var buf2 = canvas.toBuffer(undefined, 3, Canvas.PNG_FILTER_NONE);
```

The constants were added in https://github.com/Automattic/node-canvas/commit/14d9f587b6106dd900cabe0d6bc31117acb596c6. @kkoopa you added these -- do they need to be on the prototype?

kkoopa:
I do not believe that they need to be on the prototype. The third argument was added in a strange backwards-compatible way, and you pass undefined first only in this case, so you can consider cleaning it up now.

zbjornson:
I think this would be the wrong direction after all. If the Canvas spec ever gets an equivalent property, it won't be on the Canvas constructor, because there is no Canvas constructor. :)

Fine as-is methinks.

PR #935, opened 2017-07-02
Support canvas.getContext("2d", {alpha: boolean, pixelFormat: string})

### LINKED ISSUE #332: Any possible way to compress PNG or use indexed RGB on node-canvas?
I'm not sure is it possible or not? by node-canvas api after we use "createPNGStream" image will be in indexed rgb format or PNG-8

Thank you!

zbjornson:
This is an amped-up version of most of #678. ("Most" because it doesn't add the `flush` function or allow passing a buffer as that PR adds. "Amped-up" because all of the APIs work AFAICT, with the caveats noted in the docs.) My priority was to support indexed PNG encoding, but a lot of stuff came with it. I'm happy to split the 1st/2nd commits from the 3rd and 4th. Indexed PNG encoding is maybe out of scope and is something that could be extracted into a separate C++ addon, but the first two support the standard.

Docs for the major addition: https://github.com/zbjornson/node-canvas/tree/formats#image-pixel-formats-experimental

* Adds support for the [standard](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/getContext) `alpha` attribute: `canvas.getContext("2d", {alpha: false})`. In the browser tests, node-canvas matches Firefox as well as ever, and Chrome seems to have a few bugs in which canvases randomly *do* get an alpha channel.

* Adds support for the [proposed](https://github.com/WICG/canvas-color-space/blob/master/CanvasColorSpaceProposal.md) `pixelFormat` attribute: `canvas.getContext("2d", {pixelFormat: <format>})`. The color space proposal is still fairly early, and in particular the format names look like they'll change (see the open issues), but adding aliases for other formats for backwards compatibility in the future will be trivial. Even if that proposal doesn't move forward, this is nice functionality to have and a clean API (i.e. it adds options to the standard `contextOptions` object instead of positional args to the Canvas constructor).
    * `RGBA32`, `RGB24` and `A8` have full support: `getImageData`/`putImageData`/`createImageData` all work.
    * `RGB16_565` also has full support, but note: The `ImageData.data` property is a `Uint16Array` instead of a `Uint8ClampedArray` because it doesn't make sense as a uint8. This follows the linked proposal AFAICT -- `ImageData` has multiple storage types there, but [it's not specified how bit widths like 5-6-5 map to integral types](https://github.com/WICG/canvas-color-space/issues/18). This commit is easy to drop if there's any disagreement.
    * `A1` and `RGB30` have partial support: `getImageData`/`putImageData`/`createImageData` don't work because I wasn't sure how the data should be represented. Cairo packs pixels together but also has a padded stride (e.g. an A1 canvas with width 10 has a stride of 4 bytes). We could either pack all pixels together or keep them padded. This is a ton of bit manipulation given that these APIs can work on regions of a canvas. :sweat:

* Adds support (non-standard) for indexed PNG encoding. :tada: See the two new files in `examples/`. That particular image fits losslessly into 474 bytes when indexed (left), vs 1.92 kB when RGBA32 (right).

    indexed rgb
    Note that it's sorta hard to specify an exact palette index using the 0-1 value in `rgba()` syntax. Adding [`#rgba` and `#rrggbbaa` support](https://www.chromestatus.com/feature/5685348285808640) will allow integers to be provided.

    Right now only `pngStream` supports this, but if this PR lands, then `toBuffer` and maybe `toDataURL` should support it also. For the former, I'd propose moving all of the `toBuffer` options into an object since the current API is awkward (see #934).

---

No breaking API changes.
Fixes #332

### DIFF
--- a/History.md
+++ b/History.md
@@ -4,6 +4,7 @@ Unreleased / patch
  * Port has_lib.sh to javascript (#872)
  * Support canvas.getContext("2d", {alpha: boolean}) and
    canvas.getContext("2d", {pixelFormat: "..."})
+ * Support indexed PNG encoding.
 
 1.6.0 / 2016-10-16
 ==================
--- a/History.md
+++ b/History.md
@@ -2,6 +2,8 @@ Unreleased / patch
 ==================
 
  * Port has_lib.sh to javascript (#872)
+ * Support canvas.getContext("2d", {alpha: boolean}) and
+   canvas.getContext("2d", {pixelFormat: "..."})
 
 1.6.0 / 2016-10-16
 ==================
--- a/Readme.md
+++ b/Readme.md
@@ -312,6 +312,73 @@ var canvas = new Canvas(200, 500, 'svg');
 fs.writeFile('out.svg', canvas.toBuffer());
 ```
 
+## Image pixel formats (experimental)
+
+node-canvas has experimental support for additional pixel formats, roughly
+following the [Canvas color space proposal](https://github.com/WICG/canvas-color-space/blob/master/CanvasColorSpaceProposal.md).
+
+```js
+var canvas = new Canvas(200, 200);
+var ctx = canvas.getContext('2d', {pixelFormat: 'A8'});
+```
+
+By default, canvases are created in the `RGBA32` format, which corresponds to
+the native HTML Canvas behavior. Each pixel is 32 bits. The JavaScript APIs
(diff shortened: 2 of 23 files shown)


LinusU:
This looks amazing, great work! 🎉 

I think I introduced some conflict, sorry about that, would you mind fixing them? 😀

LinusU:
Awesome work as always!

PR #943, opened 2017-07-19
Restore original exports

zbjornson:
Restores all of the previously exported functions and properties. See https://github.com/Automattic/node-canvas/pull/929#issuecomment-314305548.

NB: `parseFont` is a new export added in #929. I think `parseFont` is a private, internal API. Should it be exported?

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -15,7 +15,7 @@ Still having problems, found a bug or want a feature? Fill out the form below.
 <!--- For bugs, provide a short, complete code example to reproduce the issue. -->
 ```js
 var Canvas = require('canvas');
-var canvas = new Canvas(200, 200);
+var canvas = Canvas.createCanvas(200, 200);
 var ctx = canvas.getContext('2d');
 // etc.
 ```
--- a/Readme.md
+++ b/Readme.md
@@ -84,6 +84,7 @@ loadImage('examples/images/lime-cat.jpg').then((image) => {
  node-canvas adds `Image#src=Buffer` support, allowing you to read images from disc, redis, etc and apply them via `ctx.drawImage()`. Below we draw scaled down squid png by reading it from the disk with node's I/O.
 
 ```javascript
+const { Image } = require('canvas');
 fs.readFile(__dirname + '/images/squid.png', function(err, squid){
   if (err) throw err;
   img = new Image;
@@ -95,6 +96,7 @@ fs.readFile(__dirname + '/images/squid.png', function(err, squid){
  Below is an example of a canvas drawing it-self as the source several time:
 
 ```javascript
+const { Image } = require('canvas');
 var img = new Image;
 img.src = canvas.toBuffer();
 ctx.drawImage(img, 0, 0, 50, 50);
@@ -109,7 +111,8 @@ node-canvas adds `Image#dataMode` support, which can be used to opt-in to mime d
 When mime data is tracked, in PDF mode JPEGs can be embedded directly into the output, rather than being re-encoded into PNG. This can drastically reduce filesize, and speed up rendering.
 
 ```javascript
-var img = new Image;
+const { Image } = require('canvas');
+var img = new Image();
 img.dataMode = Image.MODE_IMAGE; // Only image data tracked
 img.dataMode = Image.MODE_MIME; // Only mime data tracked
 img.dataMode = Image.MODE_MIME | Image.MODE_IMAGE; // Both are tracked
@@ -214,18 +217,19 @@ canvas.toDataURL('image/jpeg', {opts...}, function(err, jpeg){ }); // see Canvas
 canvas.toDataURL('image/jpeg', quality, function(err, jpeg){ }); // spec-following; quality from 0 to 1
 ```
 
-### Canvas.registerFont for bundled fonts
+### `registerFont` for bundled fonts
 
 It can be useful to use a custom font file if you are distributing code that uses node-canvas and a specific font. Or perhaps you are using it to do automated tests and you want the renderings to be the same across operating systems regardless of what fonts are installed.
 
-To do that, you should use `Canvas.registerFont`.
+To do that, you should use `registerFont()`.
 
 **You need to call it before the Canvas is created**
 
 ```javascript
-Canvas.registerFont('comicsans.ttf', {family: 'Comic Sans'});
+const { registerFont, createCanvas } = require('canvas');
+registerFont('comicsans.ttf', {family: 'Comic Sans'});
 
-var canvas = new Canvas(500, 500),
+var canvas = createCanvas(500, 500),
   ctx = canvas.getContext('2d');
 
 ctx.font = '12px "Comic Sans"';
@@ -297,7 +301,7 @@ ctx.antialias = 'none';
   a PDF on initialization, using the "pdf" string:
 
 ```js
-var canvas = new Canvas(200, 500, 'pdf');
+var canvas = createCanvas(200, 500, 'pdf');
 ```
 
  An additional method `.addPage()` is then available to create
@@ -323,7 +327,7 @@ ctx.addPage();
  You also need to tell node-canvas that it is working on SVG upon its initialization:
 
 ```js
-var canvas = new Canvas(200, 500, 'svg');
+var canvas = createCanvas(200, 500, 'svg');
 // Use the normal primitives.
 fs.writeFile('out.svg', canvas.toBuffer());
 ```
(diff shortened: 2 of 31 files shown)


zbjornson:
@LinusU thanks for merging all the PRs recently :tada: 

This one remains an important one that un-breaks a lot of existing code. Could you take a look sometime soon please?

Regarding my question about `parseFont`, it looks like you made it a top-level export for the sake of exposing it for unit tests. Originally it was just under `context2d` I think. Doesn't seem to matter much one way or another.

LinusU:
Sorry for not getting around to this one, I wanted to write a proper response and sort my thoughts out properly...

There is something about this that rubs me the wrong way. I'm not really a fan of adding all of the methods to the canvas itself, since that will be something that would be different between the browser version and the node version.

Another thing I like about not exposing the `Canvas` class at all is that it discourage the use of `instanceof` to check if something is a `Canvas`. This should be avoided because it will fail if there is multiple canvas classes loaded. We already have some issues reported related to this and it can happen for any number of reasons, be it different evaluation context, different versions of node-canvas, or just that the require cache was cleared (e.g. to be mocked during a test).

I'm not saying that we should do it one way or the other, I just really want to think thru the implications of this, and what the different ways to go are, and their pros and cons. I'm very open to feedback.

regarding `parseFont`, it was for the tests initially, but I was also thinking that it might be useful to expose it since it can be used both in Node.js and in the browser. We should probably move that to another module though in that case so maybe we should remove it for now :)

zbjornson:
No problem, thanks for thinking through it, all good points.

I agree that it's weird that those constructors are properties of the object returned by `require('canvas')`, but only because that top-level export is the `Canvas` constructor.

Discouraging `instanceof` checks seems unrealistic -- it's something that has to be done in some cases. In some cases where `instanceof` would return false, it also only defers the issue if you skip that check: for example, if you try to call `drawImage` with an `Image` from a different instance of node-canvas, the error will still happen, but not until it reaches the C++ layer. -- The issues with `instanceof` are real, but they're not unique to node-canvas and are just something devs have to be aware of, I think.

One possibility is making the top-level export a plain object like this, so that the other properties are at least not properties of the `Canvas` constructor:

```js
{
    // cross-environment method:
    createCanvas(...): NodeCanvas|HTMLCanvasElement,

    // standard constructors
    ImageData(...): ImageData, // don't need to provide a createImageData method since this is standard
    Image(...): Image,
    CanvasRenderingContext2D: CanvasRenderingContext2D,

    // non-standard stuff:
    Canvas(): NodeCanvas, // for `instanceof`
    Backends: { ... },
    cairoVersion: string,
    version: string,
    ...
}
```

So the properties of the exported object will be exactly the same, except it's not the ctor and there's a `Canvas` property so that you can still do `canvas instanceof nodecanvas.Canvas`. <-- note that the obvious problem becomes *what do you call the object returned by `require('canvas')`, since a really convenient name for a Canvas instance used throughout the docs and examples is `canvas`, and `nodecanvas` is sorta ugly*.

I'm sort of wondering if making the top-level export not the ctor is being too pedantic though. The `Canvas` constructor is not standard, so its static properties are also not standardized. We're not really violating anything, right?

zbjornson:
@LinusU gentle nudge on this... any thoughts?

saschagehlich:
I like to see node-canvas as kind of a drop-in replacement for the browser side `Canvas` and `Image` classes - pre-2.0 it was very easy to write code that works on both the browser and the server side. The wrapping methods `loadImage` and `createCanvas` now introduce a totally different API that is nowhere close to the way Canvas and Image are being used in the browser.

Broken `instanceof` checks are (imho) problems that can appear with any other package as well, so I don't see why that would be a reason to introduce such breaking changes.

chearon:
To be fair, I don't think `new Canvas` is any more "standard" than `createCanvas` is.

saschagehlich:
Neither `createCanvas` nor `loadImage` exist in any browser by default. `new Canvas` and `new Image` works fine in all browsers.

chearon:
`Image` does yeah, but `Canvas`:

```
> new Canvas
Uncaught ReferenceError: Canvas is not defined
> window.Canvas
undefined
```

saschagehlich:
Oh yeah, of course, you're right. It'd be `document.createElement('canvas')` then.

zbjornson:
There is, however, [`new OffscreenCanvas(w, h)`](https://developer.mozilla.org/en-US/docs/Web/API/OffscreenCanvas/OffscreenCanvas) (which is arguably closer to what node-canvas is).

Normal canvases also work with `instanceof HTMLCanvasElement`.

LinusU:
> Discouraging `instanceof` checks seems unrealistic -- it's something that has to be done in some cases. In some cases where `instanceof` would return false, it also only defers the issue if you skip that check: for example, if you try to call `drawImage` with an `Image` from a different instance of node-canvas, the error will still happen, but not until it reaches the C++ layer. -- The issues with `instanceof` are real, but they're not unique to node-canvas and are just something devs have to be aware of, I think. (The problem happens in browsers even: `foo instanceof ImageData ` fails across iframes.)

This could be fixed by having the data accessible via a normal property. We could also give a `isCanvas` and `isImage` function, like `Buffer.isBuffer`. But maybe this isn't that important 🤔

> One possibility is making the top-level export a plain object like this, so that the other properties are at least not properties of the `Canvas` constructor:

Yeah, that feels quite nice. It works very good when importing as well:

```js
const { Canvas, Image } = require('canvas')

import { Canvas, Image } from 'canvas'
```

vs.

```js
const Canvas = require('canvas')
const { Image } = Canvas

import Canvas from 'canvas'
const { Image } = Canvas
```

> So the properties of the exported object will be exactly the same, except it's not the ctor and there's a `Canvas` property so that you can still do `canvas instanceof nodecanvas.Canvas`. <-- note that the obvious problem becomes *what do you call the object returned by `require('canvas')`, since a really convenient name for a Canvas instance used throughout the docs and examples is `canvas`, and `nodecanvas` is sorta ugly*.

I think the best would be to use the destructuring interface and just avoid calling it anything

> I'm sort of wondering if making the top-level export not the ctor is being too pedantic though, compared to the small hassle it places on users to change their code. The `Canvas` constructor is not standard, so its static properties are also not standardized. We're not really violating anything, right?

Possibly, I really want to try and make the API as idiomatic as possible, since this is basically the only chance we get to change it for quite a while 😄 

> I like to see node-canvas as kind of a drop-in replacement for the browser side `Canvas` and `Image` classes - pre-2.0 it was very easy to write code that works on both the browser and the server side. The wrapping methods `loadImage` and `createCanvas` now introduce a totally different API that is nowhere close to the way Canvas and Image are being used in the browser.

The plan with the new interface was to have it work in the browser without any modifications at all, via the `browser.js` and `browser` field in the package.json. I feel that it would be very weird to expose a `document` object with a `createElement` function that returns the `Canvas`.

> Broken `instanceof` checks are (imho) problems that can appear with any other package as well, so I don't see why that would be a reason to introduce such breaking changes.

Yeah probably true, probably a good idea to export `Image` and `Canvas` as properties on the exported objects...

> There is, however, [`new OffscreenCanvas(w, h)`](https://developer.mozilla.org/en-US/docs/Web/API/OffscreenCanvas/OffscreenCanvas) (which is arguably closer to what node-canvas is).

Wow, awesome, haven't seen that one :)

> Normal canvases also work with `instanceof HTMLCanvasElement`.

👍

zbjornson:
Cool, solution is on the horizon then! :slightly_smiling_face: 

One final variant I can think of...

```js
module.exports = Canvas; // top-level export is still the constructor

// All other needed exports are still on Canvas ctor:
Canvas.ImageData = ImageData;
Canvas.CanvasGradient = CanvasGradient;
Canvas.BACKENDS.(...) = ...;
Canvas.createCanvas = createCanvas;
// etc.

Canvas.Canvas = Canvas; // Canvas is also a property of the exported object so destructuring works
```

This supports both destructuring (`let {Canvas, CanvasGradient} = require("canvas");`) *and* the 1.x syntax (`let Canvas = require("canvas")`). @LinusU what do you think of that? Maybe I'm being too timid about introducing breaking changes.

LinusU:
I would personally prefer breaking and not exporting the canvas as the `module.export`, but if I'm the only one I will of course budge 😄

zbjornson:
Alright, I think I got everything... Also updated docs and examples so they work with the 2.0 syntax.

LinusU:
Awesome work!!

Thank you for having patience with me and sorry for the slow response 😆

LinusU:
Released as `v2.0.0-alpha.3` 🎉

zbjornson:
No problem, thank you! I like the outcome.

PR #944, opened 2017-07-24
Faster, more compliant parseFont

### LINKED ISSUE #920: Text rendering becomes horribly slow after upgrading from 1.6.4 to 1.6.5
Hi all, after upgrading from 1.6.4 to 1.6.5 I started to get the feeling that text rendering had become quite a bit slower. Now I've used the nodejs inspector to investigate the issue and can confirm that there it has become up to 3x as slow for (small) text renderings, caused by the introduction of the parse-css-font module. It has horrible performance, 2x slower than actually rendering the text:

1.6.5:
node-canvas-1 6 5

1.6.4:
node-canvas-1 6 4

The test script:
```javascript

var draw = function() {
    var canvas = new Canvas(200, 50);
    ctx.font = "bold 20px Verdana";
    ctx.fillText("text " + Math.random(), 0, 0);
};
setInterval(draw, 10);
```

zbjornson:
* Improves perf by ~23x. This is the fastest method that I can come up with, short of dropping into C++, and is faster than parse-css-font and css-font-parser.

* Fixes caching -- should be done *before* parsing the string.

* Supports more of the spec:
    * `font-style`, `font-variant`, `font-weight`, `font-stretch` may appear in any order
    * supports `font-stretch` (although it will be ignored)
    * supports more units: `em`, `rem` and `q` work; `ch` and `ex` will parse but not work. Supporting `ch` and `ex` would be fastest if the API was reworked so that the actual number of device units was calculated in the same call as `fill/strokeText`.

Also, I think handling of `%` is wrong but I didn't fix it because there's an existing assertion for it. As I read it, that should be percent of the parent (default/root) font height, and with this commit I've standardized that to 16 pt (common in browsers). Right now `20%` parses to `20 px`. (@chearon?)

Fixes #920 
Fixes or at least improves #566

### DIFF
--- a/lib/parse-font.js
+++ b/lib/parse-font.js
@@ -1,64 +1,101 @@
 'use strict'
 
-const parseCssFont = require('parse-css-font')
-const unitsCss = require('units-css')
+/**
+ * Font RegExp helpers.
+ */
+
+const weights = 'bold|bolder|lighter|[1-9]00'
+  , styles = 'italic|oblique'
+  , variants = 'small-caps'
+  , stretches = 'ultra-condensed|extra-condensed|condensed|semi-condensed|semi-expanded|expanded|extra-expanded|ultra-expanded'
+  , units = 'px|pt|pc|in|cm|mm|%|em|ex|ch|rem|q'
+  , string = '\'([^\']+)\'|"([^"]+)"|[\\w-]+'
+
+// [ [ <‘font-style’> || <font-variant-css21> || <‘font-weight’> || <‘font-stretch’> ]?
+//    <‘font-size’> [ / <‘line-height’> ]? <‘font-family’> ]
+// https://drafts.csswg.org/css-fonts-3/#font-prop
+const weightRe = new RegExp(`(${weights}) +`, 'i')
+const styleRe = new RegExp(`(${styles}) +`, 'i')
+const variantRe = new RegExp(`(${variants}) +`, 'i')
+const stretchRe = new RegExp(`(${stretches}) +`, 'i')
+const sizeFamilyRe = new RegExp(
+  '([\\d\\.]+)(' + units + ') *'
+  + '((?:' + string + ')( *, *(?:' + string + '))*)')
 
 /**
- * Cache color string RGBA values.
+ * Cache font parsing.
  */
 
 const cache = {}
+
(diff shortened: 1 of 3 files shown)


lib/parse-font.js
```diff
@@ -1,64 +1,101 @@
...
-      size.value *= 96.0 / 25.4
+      font.size *= 96.0 / 25.4
       break
-    case 'cm':
-      size.value *= 96.0 / 2.54
+    case '%':
+      // TODO disabled because existing unit tests assume 100
+      // font.size *= defaultHeight / 100 / 0.75
+      break
+    case 'em':
```
chearon: Nice 👍

lib/parse-font.js
```diff
@@ -1,64 +1,101 @@
...
+    case 'cm':
+      font.size *= 96.0 / 2.54
       break
     case 'mm':
-      size.value *= 96.0 / 25.4
+      font.size *= 96.0 / 25.4
       break
-    case 'cm':
-      size.value *= 96.0 / 2.54
+    case '%':
```
chearon: So yeah, I don't think the test is right... the units do get sent to C++, but [aren't used](https://github.com/Automattic/node-canvas/blob/25b03f3e34bf25ca62f9069ebddc9b9335d9b6df/src/CanvasRenderingContext2d.cc#L1998). We probably don't need to send the units, just the normalized `px` value.

It looks like the current version would be doing percentages and `mm` wrong, I'll have to look at if the non-Pango build was ever right about that.

test/canvas.test.js
```diff
@@ -59,17 +59,27 @@ describe('Canvas', function () {
...
+      , '20px "new century schoolbook", serif'
+      , { size: 20, unit: 'px', family: 'new century schoolbook,serif' }
+      , '20px "Arial bold 300"' // synthetic case with weight keyword inside family
+      , { size: 20, unit: 'px', family: 'Arial bold 300', variant: 'normal' }
     ];
 
     for (var i = 0, len = tests.length; i < len; ++i) {
       var str = tests[i++]
-        , obj = tests[i]
+        , expected = tests[i]
```
LinusU: 👍

LinusU:
Awesome!

Issue #903, opened 2017-04-12
drawText renders text in wrong location on scaled Canvas


mvitale:
## Issue

After calling scale(scaleFactor, scaleFactor) on a Canvas instance that has been resized accordingly, text is drawn in the wrong location.
## Steps to Reproduce

```js
var fs = require('fs')
  , Canvas = require('canvas')
  , origCanvas = new Canvas(100, 100)
  , doubleCanvas = new Canvas(200, 200) // 2x size of origCanvas
  , origCtx = origCanvas.getContext('2d')
  , doubleCtx = doubleCanvas.getContext('2d')
  , origOut = null
  , origStream = null
  , doubleOut = null
  , doubleStream = null
  ;

doubleCtx.scale(2, 2); // Scale by factor of 2

draw(origCtx);
draw(doubleCtx);

origOut = fs.createWriteStream(__dirname + '/orig.png');
origStream = origCanvas.pngStream();
doubleOut = fs.createWriteStream(__dirname + '/double.png');
doubleStream = doubleCanvas.pngStream();

origStream.on('data', function(chunk) { origOut.write(chunk) });
origStream.on('end', function() { console.log('saved orig') });
doubleStream.on('data', function(chunk) { doubleOut.write(chunk) });
doubleStream.on('end', function() { console.log('saved double') });


function draw(ctx) {
  ctx.fillStyle = '#ddd';
  ctx.fillRect(0, 0, 100, 100);

  ctx.beginPath();
  ctx.strokeStyle = '#000';
  ctx.moveTo(0, 50);
  ctx.lineTo(100, 50);
  ctx.stroke();

  ctx.fillStyle = '#000';
  ctx.font = '18px Helvetica';
  ctx.fillText('test', 10, 48);
}
```

### orig.png
![orig](https://cloud.githubusercontent.com/assets/296803/24979604/da8a06f8-1fa2-11e7-8fe0-3aeae99adfec.png)

### double.png
![double](https://cloud.githubusercontent.com/assets/296803/24979616/e34dec8c-1fa2-11e7-985d-aa4fbc4df355.png)

I have verified that this is not the behavior in my browser (tested on Chrome ).

## Environment
* Version of node-canvas: 1.6.5
* Environment: Node v6.9.2 on macOS 10.12.3

LinusU:
This should be fixed in version 1.6.6 🙌

mvitale:
Awesome!

mvitale:
[...]

PR #950, opened 2017-07-31
Fix: textBaseline consider current scale.

cho45:
textBaseline is odd with scale() context.

Added visual test:

invalid place

Fix: consider current context scale and apply it to font metrics.


LinusU:
Thank you!

zbjornson:
I think this fixes https://github.com/Automattic/node-canvas/issues/903 👍

LinusU:
Released in 1.6.6

PR #951, opened 2017-07-31
Fix: textBaseline consider current scale.

LinusU:


LinusU:
Travis seems to be having some trouble at the moment 😕

LinusU:
Released in 1.6.6

Issue #962, opened 2017-08-15
`putImageData(imageData, -1, -1)` segfaults 


zixia:
Tested using canvas@2.0.0-alpha.2 (with the latest master branch) and Node.js v8.2.1.

Test case:

```js
const Canvas = require('canvas');

const canvas = new Canvas(1, 1);
const context = canvas.getContext('2d');
const imageData = new ImageData(2, 2);
context.putImageData(imageData, -1, -1);
```

What I want to do is to crop an image using a canvas. It should draw the image start from point [1, 1] but not segfault.

$ node test.js
```shell
*** Error in `/usr/bin/nodejs': free(): invalid next size (fast): 0x0000562dcf9deb60 ***
======= Backtrace: =========
/lib/x86_64-linux-gnu/libc.so.6(+0x7908b)[0x7fdf92e8d08b]
/lib/x86_64-linux-gnu/libc.so.6(+0x82c3a)[0x7fdf92e96c3a]
/lib/x86_64-linux-gnu/libc.so.6(cfree+0x4c)[0x7fdf92e9ad2c]
/usr/bin/nodejs(_ZN2v88internal18ArrayBufferTracker7FreeAllEPNS0_4PageE+0xd3)[0x562dcd92aec3]
/usr/bin/nodejs(_ZN2v88internal9SemiSpace8TearDownEv+0x58)[0x562dcd99db18]
/usr/bin/nodejs(_ZN2v88internal8NewSpace8TearDownEv+0x6e)[0x562dcd99db9e]
/usr/bin/nodejs(_ZN2v88internal4Heap8TearDownEv+0x2f9)[0x562dcd949089]
/usr/bin/nodejs(_ZN2v88internal7Isolate6DeinitEv+0x1d9)[0x562dcda26509]
/usr/bin/nodejs(_ZN2v88internal7Isolate8TearDownEv+0x53)[0x562dcda26773]
/usr/bin/nodejs(_ZN4node5StartEP9uv_loop_siPKPKciS5_+0xeee)[0x562dcde9695e]
/usr/bin/nodejs(_ZN4node5StartEiPPc+0x111)[0x562dcde90251]
/lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xf1)[0x7fdf92e343f1]
/usr/bin/nodejs(_start+0x2a)[0x562dcd2bd7ea]
======= Memory map: ========
3ca2d00000-3ca2d80000 rw-p 00000000 00:00 0 
[...]

PR #963, opened 2017-08-15
Allow negative x/y for putImageData() (#962)

zixia:
I had tested this PR in my ubuntu 17.04, it works. (#962)

### DIFF
--- a/src/CanvasRenderingContext2d.cc
+++ b/src/CanvasRenderingContext2d.cc
@@ -593,6 +593,7 @@ NAN_METHOD(Context2d::PutImageData) {
 
   uint8_t *src = imageData->data();
   uint8_t *dst = context->canvas()->data();
+  uint8_t *dstStart = context->canvas()->data();
 
   int dstStride = context->canvas()->stride();
   int Bpp = dstStride / context->canvas()->getWidth();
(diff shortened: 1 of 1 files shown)


src/CanvasRenderingContext2d.cc
```diff
@@ -662,10 +663,13 @@ NAN_METHOD(Context2d::PutImageData) {
         uint8_t b = *srcRow++;
         uint8_t a = *srcRow++;
 
-        // argb
-        // performance optimization: fully transparent/opaque pixels can be
-        // processed more efficiently.
-        if (a == 0) {
+        if (dstRow < dstStart) {
```
zbjornson: It would be nice to avoid more branches inside the innermost loop. I'd look at modifying the `rows`, `cols`, `dst` and `src` values around lines 615-616 instead. Lines 624-639 should provide good guidance; they handle negative values for the 7-argument form.

zixia: Yes, I agree with you that we should modify the outer code if we can. 

HoweverI'm not very familiar with the algorithm with the strike/row/col, so I just modify to prevent the pointer to write below the memory boundary.

And I think this PR has a bug that will cause the negative x/y not work properly: the other boundary is not considered, and the row/col calculation is not right(sub a negative value). In my code, this caused the crop function does not work as expected.

So finally I drop this method for crop my picture, but it would be nice to prevent the segfault from the code.

BTW: Do you know when will we publish the new version to NPM? 2.0 alpha in the NPM is outdated(so I have to npm install from github master).

zbjornson: No problem, I'm happy to work on a fix for this. Can probably do it this weekend. Thanks again for finding the issue.

Re: publishing, I don't have the ability to make releases, I just help out with issues where I can, and I'm not sure if there are more breaking changes that should go in before 2.0 is officially released.

zixia: The current npm version is `2.0-alpha.2`, I would like to suggest we publish a `2.0-alpha.3` with the latest code because the exported interface is all changed(BREAKING).

And because it is `alpha` version, I believe it's not a problem.

PR #964, opened 2017-08-15
Re-rasterize SVG on draw

saschagehlich:
This PR includes multiple things:

* Introducing `naturalWidth` and `naturalHeight` properties for `Image` (readonly)
* `width` and `height` are no longer readonly
* `Image#surface` checks whether `width` or `height` has changed since the last SVG rasterization. If they changed, it re-rasterizes the SVG with the correct scale.


piranna:
> Image#surface checks whether width or height has changed since the last SVG rasterization. If they changed, it re-rasterizes the SVG with the correct scale.

Why don't change them for some properties, so when they get setted the SVG are re-rasterized automatically? Probably it would be easier to do...

saschagehlich:
Because I usually change width and height separately, which would cause in
unnecessary rasterization

piranna:
> Because I usually change width and height separately, which would cause in
> unnecessary rasterization

It doesn't need to be inmediately, it can be done using `setInmediate()`...

saschagehlich:
It would still be done twice, though it should only be done once

zbjornson:
Would it even make sense to try to remove the `renderSVGToSurface` call from `loadSVGFromBuffer` to totally minimize rasterizations?

Side note: Ideally I think the SVG would render directly to the Canvas's surface instead of using a second surface within the Image. I haven't tried it, but I think that would allow SVGs rendered to SVG and PDF canvases (`new Canvas(w, h, "SVG")`) to remain as SVG elements.

saschagehlich:
@zbjornson Removing the `renderSVGToSurface` call from `loadSVGFromBuffer` didn't work, because in Image#onLoaded it says:

```c
width = naturalWidth = cairo_image_surface_get_width(_surface);
height = naturalHeight = cairo_image_surface_get_height(_surface);
```

So we could replace the `renderSVGToSurface` call with a `_surface = cairo_image_surface_create(CAIRO_FORMAT_ARGB32, width, height);` to make sure the width and height are still correct after the onLoaded event handler.

zbjornson:
@saschagehlich Ah, you're saying that the rendering step is necessary in order to determine the `width`/`height`/`naturalWidth`/`naturalHeight`? Wonder if there's a shorter way to do that (either via rsvg or by parsing the SVG width/height/viewport attributes). I guess figuring that out would also be necessary for directly rendering to the target Canvas instance's surface.

saschagehlich:
That would be a fundamental change to the Image class that I'm not capable of doing I think :/ Not that much of a C++ guy

asturur:
I have a question:
Does this code ( i tried to read it, but i could not figure out) take in consideration the current scale factor of the context?
It looks like the svg is rendered bigger and better also when the scale is bigger than 1 and not just when the width and height changes.

https://jsfiddle.net/5sdg666d/2/

Also regarding the initial

saschagehlich:
@LinusU Any feedback on this?

zbjornson:
(@LinusU if you're waiting based on my previous comment about rendering directly to the cairo_surface_t without rasterizing ever, merging this won't affect my ability to work on that in the future, and I won't be able to get to that for at least a week or two.)

LinusU:
LGTM 👍 thanks!

asturur:
Anyone knows if this code will rescale the SVG also with the scaling of the context destination?

dungmv:
Anyone can help me review this issue #1474
When I set scale canvas, svg image was lost quality

PR #970, opened 2017-08-27
support maxWidth arg to fill/strokeText

### LINKED ISSUE #119: fillText/strokeText MaxWidth Not Implemented?
It appears that when using fillText and strokeText the optional maxWidth parameter does nothing.

http://www.whatwg.org/specs/web-apps/current-work/multipage/the-canvas-element.html#text-0

chearon:
Had this sitting around if we want it. Fixes #119


LinusU:
Thanks 🍻

Issue #977, opened 2017-08-30
Using font family multiple times results in default font


saschagehlich:
## Issue or Feature
When registering multiple font files with the same font family, but different styles (like in `examples/font.js`), the default font is used instead of the given font files. When using different font family names for every font file, it works.

Right now, `examples/font.js` creates this image:

![font2](https://user-images.githubusercontent.com/119684/29859100-da6b0ac4-8d60-11e7-8280-b70cb8b641de.png)

As you can see, the `Pfennig` font is not used at all.

Changing the font definitions like this (and changing the `font` property on the rendering context accordingly):

```js
Canvas.registerFont(fontFile('Pfennig.ttf'), {family: 'Pfennig'})
Canvas.registerFont(fontFile('PfennigBold.ttf'), {family: 'PfennigBold', weight: 'bold'})
Canvas.registerFont(fontFile('PfennigItalic.ttf'), {family: 'PfennigItalic', style: 'italic'})
Canvas.registerFont(fontFile('PfennigBoldItalic.ttf'), {family: 'PfennigBoldItalic', weight: 'bold', style: 'italic'})
```

This results in:

![font](https://user-images.githubusercontent.com/119684/29859081-c6c651cc-8d60-11e7-91ba-203c612cd19f.png)

## Your Environment
* Version of node-canvas (e.g. 1.4.0): 2.0.0-alpha.3
* Environment (e.g. node 4.2.0 on Mac OS X 10.8): node 7.9.0 on macOS 10.12.6

chearon:
Spent an hour trying to figure out why, and then I remembered that I already know why. 

The three consecutive `registerFont` calls basically add 3 `PangoFontDescription`s to an internal list held in node-canvas. When you set `ctx.font = "pfennigFont"` it finds 3 matches, and [joins the real font names with commas](https://github.com/Automattic/node-canvas/blob/master/src/Canvas.cc#L753) before it sends the font query to Pango. So what gets sent to the OS in this case would be `{family: "Pfennig, Pfennig, Pfennig"}`. 

[...]

PR #984, opened 2017-09-01
macOS font selection improvement (see #977)

chearon:
This makes the font example work on macOS.

People can still run into the Pango bug on Mac if they do something like `ctx.font = "10px Arial, Helvetica".`

Behdad said that the patch looked good but hasn't merged it yet: https://bugzilla.gnome.org/show_bug.cgi?id=762873

### DIFF
--- a/src/Canvas.cc
+++ b/src/Canvas.cc
@@ -742,6 +742,7 @@ Canvas::ResolveFontDescription(const PangoFontDescription *desc) {
   // if someone registered two different fonts under the same family name.
   // https://drafts.csswg.org/css-fonts-3/#font-style-matching
   char **families = g_strsplit(pango_font_description_get_family(desc), ",", -1);
+  GHashTable *seen_families = g_hash_table_new(g_str_hash, g_str_equal);
   GString *resolved_families = g_string_new("");
 
   for (int i = 0; families[i]; ++i) {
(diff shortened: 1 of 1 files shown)


LinusU:
:beers:

src/Canvas.cc
```diff
@@ -750,8 +751,14 @@ Canvas::ResolveFontDescription(const PangoFontDescription *desc) {
 
     for (; it != _font_face_list.end(); ++it) {
       if (g_ascii_strcasecmp(families[i], pango_font_description_get_family(it->user_desc)) == 0) {
-        if (renamed_families->len) g_string_append(renamed_families, ",");
-        g_string_append(renamed_families, pango_font_description_get_family(it->sys_desc));
+        char *name = g_strdup(pango_font_description_get_family(it->sys_desc));
+
+        // Avoid sending duplicate SFNT font names due to a bug in Pango for macOS:
+        // https://bugzilla.gnome.org/show_bug.cgi?id=762873
+        if (g_hash_table_add(seen_families, name)) {
```
zbjornson: @chearon `g_hash_table_add` was added in 2.32, and the latest Windows GTK bundle (the one suggested in the install instructions) is 2.22.

Apparently it's equivalent to `g_hash_table_replace(seen_families, name, name)`, but in 2.22 `g_hash_table_replace` returns `void`. Also 2.22 has no `_contains` function, so maybe this:

```cpp
        bool existed = g_hash_table_lookup(seen_families, name) == NULL;
        g_hash_table_replace(seen_families, name, name);
        if (existed) {
```

That fixes the Windows build, but I don't have a Mac to test if it fixes the font resolution bug.

(2.22 source: https://github.com/GNOME/glib/blob/glib-2-22/glib/ghash.c)

chearon: Ugh, 2.22 was released in 2009. 2.32 in 2012

I've also thought of maybe switching to c++ libraries, but the GLib stuff plays nicely with Pango.

I'll switch it to `g_hash_table_replace` in a sec

LinusU:
Published in `2.0.0-alpha.5`

PR #985, opened 2017-09-02
Fix text baselines and measureText

### LINKED ISSUE #983: Text baseline for rotated text is wrong in 2.0
Since c927c809, this (new) test fails:

```js
tests['rotated baseline'] = function (ctx) {
  ctx.font = '12px Arial';
  ctx.fillStyle = 'black';
  ctx.textAlign = 'center';
  ctx.translate(50, 82.5);
  ctx.rotate(-Math.PI / 2);
  ctx.textBaseline = "bottom";
  ctx.fillText("Hello world!", 0, -18);
}
```

pasted_image_at_2017_08_31_01_51_pm
(I drew the line for clarity)

In this case, `yy` from the cairo matrix is `cos(pi/2)` ~= 0, so the baseline shift is effectively doing nothing.

zbjornson:
* Fixes text baseline adjustment (`fillText`, `strokeText`, `measureText`), including when rotation and scale transforms are active. (Fixes #983)
* Fixes `measureText` results. Some had the wrong signs and/or weren't adjusting for the baseline properly.

tl;dr: Fixes stuff, no regressions, still not perfect.

---

### Baselines
*The good* - these are fixed:
image

*The bad* - this still doesn't work:
image
Known issue: all we have is the current transformation matrix, and it's impossible to uniquely decompose that matrix into the scale, rotation and skew matrices. (We need the scale matrix only to adjust the pango metrics.) See https://math.stackexchange.com/questions/1415854/decompose-affine-transformation-including-shear-in-x-and-y and the linked questions. Given that the Canvas spec defines `context.transform(a, b, c, d, e, f)` which thus lets you define a skew, scale and rotation simultaneously, I don't know what to do ... but I think skew is relatively rare, and this PR is an improvement over what it was. Haven't looked what browsers do; they might have tighter integration of their text and graphics libraries that avoids this problem.

---

### `measureText`
Also tricky because there's nothing to compare against. Chrome is the only browser with any support right now, but I think some of their values and [test assertions](https://chromium.googlesource.com/chromium/src/+/master/third_party/WebKit/LayoutTests/fast/canvas/canvas-textMetrics-all.html) are wrong as I noted in [my comment](https://bugs.chromium.org/p/chromium/issues/detail?id=277215). (Definitions in the spec are [here](https://html.spec.whatwg.org/multipage/canvas.html#dom-textmetrics-width-2). There's also an [older polyfill](https://github.com/motiz88/canvas-text-metrics-polyfill) with [visual tests](https://rawgit.com/motiz88/canvas-text-metrics-polyfill/master/tests/visual-test.html) -- the logic for drawing the bounding box in the new `tests['measureText()']` test matches [the logic from the polyfill's tests](https://github.com/motiz88/canvas-text-metrics-polyfill/blob/master/tests/visual-test.js#L39-L42) so I feel reasonably comfortable that the metrics are now correct.)

image
(left is node-canvas, right is Chrome with the experimental canvas flag enabled)

### DIFF
--- a/src/CanvasRenderingContext2d.cc
+++ b/src/CanvasRenderingContext2d.cc
@@ -1928,13 +1928,31 @@ NAN_METHOD(Context2d::StrokeText) {
 }
 
 /*
+ * Gets the baseline adjustment in device pixels, taking into account the
+ * transformation matrix. TODO This does not handle skew (which cannot easily
+ * be extracted from the matrix separately from rotation).
+ */
+inline double getBaselineAdjustment(PangoFontMetrics* metrics, cairo_matrix_t matrix, short baseline) {
+  double yScale = sqrt(matrix.yx * matrix.yx + matrix.yy * matrix.yy);
+  switch (baseline) {
+  case TEXT_BASELINE_ALPHABETIC:
+    return (pango_font_metrics_get_ascent(metrics) / PANGO_SCALE) * yScale;
+  case TEXT_BASELINE_MIDDLE:
+    return ((pango_font_metrics_get_ascent(metrics) + pango_font_metrics_get_descent(metrics)) / (2.0 * PANGO_SCALE)) * yScale;
+  case TEXT_BASELINE_BOTTOM:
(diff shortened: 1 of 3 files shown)


LinusU:
Published in `2.0.0-alpha.5`

PR #988, opened 2017-09-04
Support #RGBA, #RRGGBBAA hex colors

zbjornson:
From CSS Color Module Level 4 https://drafts.csswg.org/css-color/#hex-notation

https://developer.mozilla.org/en-US/docs/Web/CSS/color_value

Firefox has had this since 49, Safari since 9.1, Chrome in 62 (52 behind flag).


LinusU:
Great :+1:

Issue #990, opened 2017-09-06
Export CanvasPattern and CanvasGradient


cmdelatorre:
It'd be great if you exported CanvasPattern and CanvasGradient.

I have OpenLayers as a dependency. Internally, it's using CanvasPattern and CanvasGradient. I installed `node-canvas` for testing: I'm using `mocha` and `jsdom` to set a suitable DOM environment for the tests to run. So, in my tests configuration file I have something like:

```
const canvas = require('canvas');
// ...
global['CanvasRenderingContext2D'] = canvas.Context2d;
```
Now, some tests that use OpenLayers are failing because it is expecting CanvasPattern and CanvasGradient to be in the global scope.

I can see that you have such classes within your project but they are not public. Or maybe I don't know how to import/use them?

## Environment
* canvas@1.6.6
* node v6.11.1 on Ubuntu 16.04.3 LTS

zbjornson:
In 2.x, `CanvasPattern` is exposed. `CanvasGradient` isn't, but should be since it is in browsers.

2.x also exports "Context2d" with the correct name, so you could do something like this:
```js
["CanvasRenderingContext2D", "CanvasPattern" /*, "CanvasGradient" */].forEach(obj => {
  global[obj] = canvas[obj]
})
```

Full list of exports in 2.x: https://github.com/Automattic/node-canvas/blob/master/index.js#L49
This is an easy fix. I can do it today, or if you want to open a PR, `CanvasGradient: bindings.CanvasGradient` just needs to be added to that `exports` object.

cmdelatorre:
Awesome. Do you have an ETA for v2 to be in NPM?
I'll see if I can open that PR...

zbjornson:
Linus has been publishing alpha versions regularly. The latest is 2.0.0-alpha.5 (`npm i canvas@2.0.0-alpha.5`).

cmdelatorre:
Unlucky for me, I cannot move to v2 yet because of this: https://github.com/Automattic/node-canvas/issues/978#issuecomment-327566491

zbjornson:
Here's a hack for 1.x if you need:

```js
const Canvas = require("canvas");
const canvasBindings = require("canvas/lib/bindings");
Object.assign(Canvas, {
  CanvasGradient: canvasBindings.CanvasGradient,
  CanvasPattern: canvasBindings.CanvasPattern
});
```

cmdelatorre:
[...]

PR #991, opened 2017-09-06
Export CanvasGradient

cmdelatorre:
As suggested in https://github.com/Automattic/node-canvas/issues/990
(more specifically [here](https://github.com/Automattic/node-canvas/issues/990#issuecomment-327553455) )


LinusU:
Cheers 🍻

Issue #930, opened 2017-06-27
Error: The specified procedure could not be found. With sharp js.


SeanSobey:
## Issue or Feature

When using this with [sharp js](https://github.com/lovell/sharp) canvas fails to load.

## Steps to Reproduce
```js
// Both work fine
require('canvas');
require("sharp");

// Throws error
require("sharp");
require('canvas');
```

## Environment
* node-canvas (1.6.5)
* sharp (0.18.1)
* node 7.10.0 x64
* Windows 10

zbjornson:
I can reproduce this. Actual loader output:

```
3cf0:0e20 @ 45833265 - LdrLoadDll - ENTER: DLL name: \\?\D:\git\temp\node_modules\canvas\build\Release\canvas.node
3cf0:0e20 @ 45833265 - LdrpLoadDllInternal - ENTER: DLL name: \\?\D:\git\temp\node_modules\canvas\build\Release\canvas.node
3cf0:0e20 @ 45833265 - LdrpResolveDllName - ENTER: DLL name: \\?\D:\git\temp\node_modules\canvas\build\Release\canvas.node
3cf0:0e20 @ 45833265 - LdrpResolveDllName - RETURN: Status: 0x00000000
3cf0:0e20 @ 45833265 - LdrpMinimalMapModule - ENTER: DLL name: \\?\D:\git\temp\node_modules\canvas\build\Release\canvas.node
'node.exe' (Win32): Loaded 'D:\git\temp\node_modules\canvas\build\Release\canvas.node'. Symbols loaded.
3cf0:0e20 @ 45833265 - LdrpMinimalMapModule - RETURN: Status: 0x00000000
3cf0:0e20 @ 45833265 - LdrpFindKnownDll - ENTER: DLL name: libpng14-14.dll
3cf0:0e20 @ 45833265 - LdrpFindKnownDll - RETURN: Status: 0xc0000135
3cf0:1118 @ 45833265 - LdrpSearchPath - ENTER: DLL name: libpng14-14.dll
[...]

Issue #909, opened 2017-04-26
getImageData and putImageData could be much faster


pcordes:
`NAN_METHOD(Context2d::GetImageData)` and `PutImageData` could be much faster (at least a factor of 2, or much more with SIMD).

gcc doesn't merge those byte-stores or byte-loads into 32-bit stores (even on x86 where unaligned loads/stores never fault and are fast), so that's a major bottleneck.  (If you look at the asm, it's not pretty).

Transforming ARGB to/from RGBA can be done with a single 32-bit rotate by 8 bits.  See [this SO post](http://stackoverflow.com/questions/776508/best-practices-for-circular-shift-rotate-operations-in-c) for a portable + safe rotate idiom that reliably compiles to a single rotate instruction on most compilers.

To handle the portable-to-native endian conversion, use something like [`uint32_t ntohl(uint32_t)`](https://linux.die.net/man/3/htonl) from `<arpa/inet.h>` if available.  (On any decent implementation, these functions will inline.  e.g. on x86 gcc, it inlines to a single [`bswap` instruction](https://hjlebbink.github.io/x86doc/html/BSWAP.html)).  I'm not sure about Windows, or if we'd need an `#ifdef` to get a different endian-conversion function / intrinsic.

In summary, the get and put image data functions should both be doing 32-bit loads and stores, and doing an endian-conversion + a rotate.  The current GetImageData should run at about 1 byte per clock cycle on recent Intel CPUs (as compiled by gcc6.3.1 for x86-64), for ranges of pixels that have alpha == 0 or 255.  Maybe even slower on AMD Bulldozer-family CPUs, since they have fewer integer execution units.

A 32-bit-at-a-time bswap+rotate version should be at least twice as fast.  It probably can't manage 1 pixel per clock since it still has to check alpha.  (Although you can do that without unpacking the alpha with a shift, by checking if `pix > (0xffUL << 24) || pix < (1UL << 24)` in a format where we know alpha is in the MSB 8 bits.)

The floating-point stuff for alpha looks ridiculously slow.  Probably an integer multiply and shift would be a win.  (Or integer division by 255, which compilers can implement with multiply and shift.)  At the very least, GetImageData should do `float alpha = 255.0f / (float)a` and then three multiplies, instead of the other way around and three divides.  (Multiply is much faster than divide).

I guess I should just write the code and submit a pull request.

---

[...]

PR #992, opened 2017-09-07
Add prebuilds

chearon:
`canvas-prebuilt` has been working for a year now, and the latest version of the Linux build has broad compatibility. This would try to download the binaries from the prebuilt repo before `node-gyp` and hopefully cut down on a lot of installation issues.

You can check out this branch and `npm install` to try it out.

If we do go forward with this I can give access to @LinusU and @zbjornson and maybe others to the prebuilt repo, and document the release process better. It basically involves setting environment variables to indicate which version you want to build and release, and then you re-build the latest commit on AppVeyor/Travis.

(We could also use `prebuild-install`, it has half the dependencies which is nice. But (1) `node-pre-gyp` is more popular and thus more likely to already be installed and (2) `prebuild-install` is not compatible with `canvas-prebuilt`'s tarball structure (fixable for later versions)).

edit: ~~Forgot this affects CI too, will fix that soon~~ fixed, ~~but CI is stalled now~~ CI fixed

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,6 @@
 language: node_js
+install:
+  - npm install --build-from-source
 node_js:
   - '8'
   - '6'
--- a/Readme.md
+++ b/Readme.md
@@ -27,9 +27,13 @@
 $ npm install canvas
 ```
 
-Unless previously installed you'll _need_ __Cairo__ and __Pango__. For system-specific installation view the [Wiki](https://github.com/Automattic/node-canvas/wiki/_pages).
+By default, binaries for macOS, Linux and Windows will be downloaded. If you want to build from source, use `npm install --build-from-source`.
 
 Currently the minimum version of node required is __4.0.0__
+
+### Compiling
+
+If you don't have a supported OS or processor architecture, or you use `--build-from-source`, the module will be compiled on your system. Unless previously installed you'll _need_ __Cairo__ and __Pango__. For system-specific installation view the [Wiki](https://github.com/Automattic/node-canvas/wiki/_pages).
 
 You can quickly install the dependencies by using the command for your OS:
 
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "description": "Canvas graphics API backed by Cairo",
   "version": "2.0.0-alpha.5",
   "author": "TJ Holowaychuk <tj@learnboost.com>",
+  "main": "index.js",
   "browser": "browser.js",
   "contributors": [
     "Nathan Rajlich <nathan@tootallnate.net>",
@@ -27,9 +28,17 @@
     "pretest": "node-gyp build",
     "test": "standard examples/*.js test/server.js test/public/*.js benchmark/run.js util/has_lib.js browser.js index.js && mocha test/*.test.js",
     "pretest-server": "node-gyp build",
-    "test-server": "node test/server.js"
+    "test-server": "node test/server.js",
+    "install": "node-pre-gyp install"
+  },
+  "binary": {
+    "module_name": "canvas-prebuilt",
+    "module_path": "build/Release",
+    "host": "https://github.com/node-gfx/node-canvas-prebuilt/releases/download/",
+    "remote_path": "v{version}"
   },
   "dependencies": {
+    "node-pre-gyp": "^0.6.36",
     "nan": "^2.4.0"
   },
   "devDependencies": {


zbjornson:
(Would be nifty to setup a release hook that kicks off the prebuild process sometime.)

Does the readme need to be updated to say something like "prebuilt will be used by default; if you want to build from source, use `--build-from-source` and install all these dependencies"?

chearon:
> (Would be nifty to setup a release hook that kicks off the prebuild process sometime.)

Definitely, I might have to set something up that builds the latest version if triggered from a URL. Then just add a web hook to automattic/node-canvas. The environment variables are useful for going back and updating binaries for old versions (e.g. node 9 released).

I'll update the readme.

LinusU:
🙌 

Awesome, would you mind writing down some quick notes on how the releases process works?

I created a `node-gfx` organization on Github where we could potentially put stuff like that repo? If you think it's a good idea... We could also start breaking out some other parts of canvas there (eg. font string parsing, the matrix transform classes, etc.) 🤔

chearon:
Sure! I'll document it in a wiki. I'm okay with moving the repo as long as GH redirects old release tarball links correctly. I'm going to test that out now. If it doesn't, I'll have to fork a new repo to `node-gfx` instead.

LinusU:
Great :)

I added both you and @zbjornson as owners, I'd be happy to add anyone one else that's been contributing to node-canvas as well!

Also, I just made the name up on the spot, if anyone have some better ideas we could of course roll with that 🙌

chearon:
It works! Good on you, GitHub. I'm moving it now and then I'll update the PR.

```
wget https://github.com/chearon/test-repo/releases/download/v1.0.0/canvas-prebuilt-v2.0.0-alpha.5-node-v57-darwin-x64.tar.gz
--2017-09-11 11:19:36--  https://github.com/chearon/test-repo/releases/download/v1.0.0/canvas-prebuilt-v2.0.0-alpha.5-node-v57-darwin-x64.tar.gz
Resolving github.com... 192.30.253.113, 192.30.253.112
Connecting to github.com|192.30.253.113|:443... connected.
HTTP request sent, awaiting response... 301 Moved Permanently
Location: https://github.com/node-gfx/test-repo/releases/download/v1.0.0/canvas-prebuilt-v2.0.0-alpha.5-node-v57-darwin-x64.tar.gz [following]
--2017-09-11 11:19:37--  https://github.com/node-gfx/test-repo/releases/download/v1.0.0/canvas-prebuilt-v2.0.0-alpha.5-node-v57-darwin-x64.tar.gz
Reusing existing connection to github.com:443.
HTTP request sent, awaiting response... 302 Found
```

zbjornson:
> I created a node-gfx organization on Github where we could potentially put stuff like that repo? If you think it's a good idea... We could also start breaking out some other parts of canvas there (eg. font string parsing, the matrix transform classes, etc.)

Good idea! If we end up with pluggable encoders/decoders (e.g. WebP and anything else beyond the Canvas spec requirements), that could be a good place to put them as well. (Could even be a new home for node-canvas if no one at automattic wants to "own" this repo anymore and deal with admin stuff like authorizing appveyor.)

chearon:
Finally created a wiki [here](https://github.com/node-gfx/node-canvas-prebuilt/wiki) to document the release process. If we were to start using this repo to build releases, you could ignore the **Releasing to NPM** section. Let me know if anything is unclear, I think I covered everything.

zbjornson:
@LinusU are you cool with merging this? :)

LinusU:
Absolutely 👍 🚀 🎉

LinusU:
Published as `canvas@2.0.0-alpha.10` 🎉

piranna:
Can this allow to move forward to static builds? I'm busy lately, but can be able to upgrade my pull-requests if needed...

chearon:
Static builds might still be a good idea to avoid issues like node-gfx/node-canvas-prebuilt#30 and #930. I think I've convinced myself that the node-canvas repo should build against system libraries, though, because that + Gyp (CMake in the future?) gives end-users a standardized, easier to understand build.

LinusU:
Yeah, static builds are nice 👍 

Personally, I'm super interested in shipping the entire module as a `wasm` file 😍 but that's probably a bit into the future...

piranna:
> Static builds might still be a good idea to avoid issues like node-gfx/node-canvas-prebuilt#30 and #930

Yeah, and also the problems regarding to OSX and Windows builds due the need to install libraries and so.


> Personally, I'm super interested in shipping the entire module as a wasm file  but that's probably a bit into the future...

That would be freaking cool! :-D I'm also looking about distribution of Node.js compiled modules as wasm files, no need of build environment for the `npm install` command :-) Does anybody know if wasm files can be able to load `.so` files themselves? Since they are linked against custom C libs, maybe it's just a matter of calling `dlopen()`...

chearon:
Heh, we're all dreaming of no native dependencies.

Someone made an implementation of Canvas in [pure JS](https://github.com/joshmarinacci/node-pureimage) but without GPU it's just not practical.

If Node gave us a way to communicate with the GPU someone could compile Cairo to WASM. HarfBuzz (font shaping) is already probably not that hard to compile to WASM, but then you have to do FreeType (glyph rendering) too and something for font selection, which could be done in JS since I think the spec is loose on that.

> Does anybody know if wasm files can be able to load .so files themselves?

That would be cool, but how do we know the system has the library or has the right version? Isn't it just offloading the problem from build to runtime?

LinusU:
> Does anybody know if wasm files can be able to load .so files themselves?

Not directly, although you could probably expose `dlopen` to the wasm module if you wanted to.

> If Node gave us a way to communicate with the GPU someone could compile Cairo to WASM.

Is Cairo GPU accelerated? I didn't think it was 🤔 

I think that WebGL integration is planned for WASM, but that is probably quite a while away, and it's also not clear if Node.js would choose to have that support... But if Cairo is just running on the CPU, we probably would be able to get comparable speeds with pure WASM.

zbjornson:
> Is Cairo GPU accelerated? I didn't think it was

It isn't by default, but can use OpenGL.

#909 intends to add some x86 intrinsics that WASM doesn't support, btw.

---

My main issue with static builds is that they increase memory usage and occupy more of the CPU cache when the module is loaded by multiple node.js processes -- common because of node's native `cluster` module. I'd really want it to be an optional linking mode.

LinusU:
> [...] increase memory usage and occupy more of the CPU cache when the module is loaded by multiple node.js processes -- common because of node's native cluster module. 

Very good point 👍

piranna:
> Someone made an implementation of Canvas in pure JS but without GPU it's just not practical.

I don't think so, JIT can greatly improve performance, so for headless code I find it a good alternative. GPU would improve performance, but I don't find it something needed except for intensive calculations or drawing to an screen.


> If Node gave us a way to communicate with the GPU someone could compile Cairo to WASM

Node.js bindings for `libdrm`, anyone? :-)


> Isn't it just offloading the problem from build to runtime?

I think this already is happening on load (aka. runtime) time with dynamic libraries...


> Not directly, although you could probably expose dlopen to the wasm module if you wanted to.

That would be a cool project :-) Sort of like FFI, but having WASM just only floats and integers, it would be fairly easier.


> Is Cairo GPU accelerated? I didn't think it was

Maybe just only optionally for graphic backends, not for in-memory rendering... Or use the GPU and later fetch the rendered image... Definitely I don't think it's working by default but instead is an opt-in, so probably we are not using it.


> I think that WebGL integration is planned for WASM, but that is probably quite a while away, and it's also not clear if Node.js would choose to have that support...

If they add WebGL support it would be API level, not in the core, so Node.js will not have anything of this and would need to be done in an external module.


> But if Cairo is just running on the CPU, we probably would be able to get comparable speeds with pure WASM.

I think so.


> My main issue with static builds is that they increase memory usage and occupy more of the CPU cache when the module is loaded by multiple node.js processes -- common because of node's native cluster module.

That's not really a problem because the operating system already have the compiled modules in the cache, so the only situation where it would be duplicated it's if you are using statically linked `node-canvas` and at the same time `libcairo` in another process, and if so it would only be that they are duplicated, no more. Maybe an increase of about 7 MB in RAM in total due to the duplicated libraries, independently of the Node.js instances running due to the `cluster` module.


> I'd really want it to be an optional linking mode.

Yes, that has been my idea from the begin, that static linking would only be done when dynamic libraries are not installed on the system or when enabled explicitly by hand.

PR #994, opened 2017-09-07
Minimal backport of #985 (rotated text baselines) to v1.x

zbjornson:
#985 fixes a regression that was introduced in 1.6.6, but that patch won't apply cleanly to v1.x. If you want to backport it, here's the minimal patch.

(https://github.com/Automattic/node-canvas/issues/983#issuecomment-327783266)


LinusU:
Restarting the build after merging #996

LinusU:
Green!

LinusU:
Released as 1.6.7, thanks!

PR #995, opened 2017-09-08
Support currentTransform

### LINKED ISSUE #658: Support for CanvasRenderingContext2D.currentTransform
I think it would be nice to support this feature, it allows you to either set, or read back the current transformation that is being applied to the canvas.

More info at MDN: [CanvasRenderingContext2D.currentTransform](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/currentTransform)

zbjornson:
Fixes #658

---

I maybe went overboard by (almost) fully implementing `DOMPoint` and `DOMMatrix` (almost: skipped `invertSelf`/`inverse` and most of the CSS transforms that should be accepted in the constructor). If you want to keep the code base small, I could replace it with just an object exposing the matrix values, with none of the methods. (Would put the rest of the implementation in a separate module or donate it to jsdom if they want it.)

Also, this isn't super optimized since I don't expect the methods will get used much.

**Edit** Option 3 could be putting `DOMMatrix` in a separate optional module, and if it's present, `currentTransform` will return an instance of that; otherwise it'll be a plain object or a minimal implementation with just the a/b/c/d/e/f/mXY properties. Sort of like how jsdom integrates with canvas.

### DIFF
--- a/index.js
+++ b/index.js
@@ -8,6 +8,8 @@ const fs = require('fs')
 const PNGStream = require('./lib/pngstream')
 const PDFStream = require('./lib/pdfstream')
 const JPEGStream = require('./lib/jpegstream')
+const DOMMatrix = require('./lib/DOMMatrix').DOMMatrix
+const DOMPoint = require('./lib/DOMMatrix').DOMPoint
 
 function createCanvas (width, height, type) {
   return new Canvas(width, height, type)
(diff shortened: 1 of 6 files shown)


LinusU:
Looks great!

Issue #1001, opened 2017-09-13
`ctx.getImageData` reports `Floating point exception: 8`


angelathewebdev:
## Issue

I'm using the latest version `2.0.0-alpha3`, and I keep running into `Floating point exception: 8`

I tried different images, but I still got the same result

## Steps to Reproduce
```js
fs.readFile(__dirname + '/static/t1.jpeg', (err, squid) => {
  if (err) throw err;
  const img = new Image;
  const canvas = createCanvas(img.width, img.height);
  const ctx = canvas.getContext('2d');
  img.src = squid;
  ctx.drawImage(img, 0, 0, img.width, img.height); //this is all good
  const imageData = ctx.getImageData(0, 0, img.width, img.height); // return floating point exception: 8
});
```

## Your Environment
* Version of node-canvas (e.g. 1.4.0): 2.0.0-alpha3
* Environment (e.g. node 4.2.0 on Mac OS X 10.8):  Mac OS X 10.12.6

zbjornson:
I will fix that crash, but it's because your image isn't loaded yet and the canvas has zero width. Try something like this:

```js
fs.readFile(__dirname + '/static/t1.jpeg', (err, squid) => {
  if (err) throw err;
  const img = new Image;
  img.onload = function () {
	const canvas = createCanvas(img.width, img.height);
	const ctx = canvas.getContext('2d');
	ctx.drawImage(img, 0, 0, img.width, img.height); //this is all good
	const imageData = ctx.getImageData(0, 0, img.width, img.height);
  }
  img.onerror = function (e) { console.log("Error", e); }
  img.src = squid;
});
```

You could also use `loadImage`:

```
const { loadImage } = require("canvas");
loadImage(squid).then(() => {
  // image is loaded; safe to do stuff
})
```

zbjornson:
Gah, sorry. You're creating the Canvas before `img.src = squid`, so `img.width` and `img.height` are 0.

Issue #966, opened 2017-08-18
Jpeg load error


qertis:
## Issue or Feature

```
ctx.drawImage(img, 0, 0, img.width, img.height);
        ^

Error: Image given has not completed loading
    at Error (native)
    at C:\Users\iavto\WebstormProjects\CanvasLevels\index.js:22:9
    at FSReqWrap.readFileAfterClose [as oncomplete] (fs.js:445:3)
```

## Steps to Reproduce

```js
const Canvas = require('canvas');
const fs = require('fs')
var p = __dirname + '\\images\\noplan.jpg';
fs.readFile(p, function(err, data) {
    if (err) throw err;
    var img = new Canvas.Image;
    img.src = data;
    var canvas = new Canvas(img.width, img.height);
    var ctx = canvas.getContext('2d');
    ctx.drawImage(img, 0, 0, img.width, img.height);
})
```

## Your Environment
* Version of node-canvas (e.g. 1.4.0): 1.6.6
* Environment (e.g. node 4.2.0 on Mac OS X 10.8): Windows 10

qertis:
When I load *.png file all works.

LinusU:
Can you attach an `onerror` handler to the image, and log any potential error?

```js
img.onerror = (err) => console.error(err.stack)
```

zbjornson:
I think this is the same as #1001, and I just posted a solution there: The error doesn't happen (for me at least) if you wait for the `img.onload` event.

It's odd that it only happens for jpeg; it suggests there's something async but I'm not seeing it. Plus, apparently the `onerror` and `onload` listeners *must* be registered *before* the `img.src = ...` line, which also suggests that it's sync.

(Aside: that the order matters is sort of unusual for node.js. Usually events are triggered after a `process.nextTick` so that you can register them after the `img.src = ...`, but this is also a bit of an unusual scenario.)

cvalente56:
I was having the same issue, and spent the last couple of days trying find out why.

For me the issue was due the [this change](https://github.com/Automattic/node-canvas/commit/5702c4745a72956a34d635344937384a1b3bcd2e#diff-830233ab0726a31d9a16fa4f237b58ac) to the /util/has_lib script. 


I’m running my application from within a custom docker container with all of the native decencies compiled to /my-libs (in which node-canvas v1.6.9 runs just fine)

The new javascript based script fails the hasLdconfig() call and then fails  to manually find the presence of libjpeg.so in the "common library locations”

[...]

Issue #959, opened 2017-08-13
Error: Image given has not completed loading


kricha:
I'm trying to add remove image to canvas, but there is some troubles.

i'm getting this:
```
        ctx.drawImage(avatar, 50, 50 );
            ^

Error: Image given has not completed loading
    at Request.request.get
```

I want to do this on server side (using express):
```js
app.get('/test', (req, res) => {
    let Canvas = require('canvas')
        , Image = Canvas.Image
        , canvas = new Canvas(500, 500)
        , ctx = canvas.getContext('2d');

    let request = require('request');
    let avatar = new Image;
    request.get('http://s.4pda.to/pKugVJ7TmgG1Tg5z0SsVLfRCUqTAz2oqz1lz2z2co.jpg', (err, res) => {
        avatar.src = res;
        ctx.drawImage(avatar, 50, 50 );
    });

    ctx.font = '30px Impact';
    ctx.rotate(.1);
    ctx.fillText("Awesome!", 50, 100);
    let te = ctx.measureText('Awesome!');
    ctx.strokeStyle = 'rgba(0,0,0,0.5)';
    ctx.beginPath();
    
    ctx.stroke();
    
    res.setHeader('Content-Type', 'image/png');
    canvas.pngStream().pipe(res)
});
// etc.
```

## Your Environment
* Version of node-canvas (e.g. 1.6.6):
* Environment (e.g. node 8.3.0 on Mac OS X 10.12.6):

I have already tried to reinstall, rebuilt and do all from manual.

LinusU:
Could you try waiting for the `onload` to fire on the image?

Actually, that might not even be to problem, is `res` really a `Buffer` instance? It should probably be something like `avatar.src = res.body`.

```diff
     request.get('http://s.4pda.to/pKugVJ7TmgG1Tg5z0SsVLfRCUqTAz2oqz1lz2z2co.jpg', (err, res) => {
-        avatar.src = res;
-        ctx.drawImage(avatar, 50, 50 );
+        avatar.onload = () => ctx.drawImage(avatar, 50, 50)
+        avatar.src = res.body
     });
```

I think another problem might be that you are sending the canvas to the browser before the image have been drawn to it, you need to wait with sending the response until all operations are done.

Maybe something like this would be better:

```js
const Canvas = require('canvas')
const request = require('request')

function loadImage (url) {
  return new Promise((resolve, reject) => {
    const img = new Canvas.Image()

    img.onload = () => resolve(img)
[...]

Issue #945, opened 2017-07-24
Support img.src = <url>


Stomp1998:
Give me a example i stuck in pleaaz

jhuckaby:
I assume you are asking how to load an image into node-canvas from a URL?  You first have to use an HTTP client library to fetch the URL, and get yourself a `Buffer` object containing the binary image data in the HTTP response, then pass that to the node-canvas Image `src` property.

Here is an example using my own [pixl-request](https://www.npmjs.com/package/pixl-request) HTTP library, but there are tons of others available which can do this as well ([request](https://www.npmjs.com/package/request), [got](https://www.npmjs.com/package/got), [superagent](https://www.npmjs.com/package/superagent), etc.).

```js
var fs = require('fs');
var Request = require('pixl-request');
var Canvas = require('canvas');
var Image = Canvas.Image;

var url = "http://placehold.it/256x256/0000ff/ffffff&text=node-canvas";

var request = new Request();
request.get( url, function(err, resp, data) {
	if (err) throw err;
	
	var img = new Image();
	img.src = data;
	
	var canvas = new Canvas(256, 256);
	var ctx = canvas.getContext('2d');
	
	ctx.drawImage(img, 0, 0, 256, 256);
	canvas.createPNGStream().pipe(fs.createWriteStream('image-url.png'));
});
```

This should produce `image-url.png` in the current directory looking like this:

![image-url](https://user-images.githubusercontent.com/700765/28801418-49cf7994-7606-11e7-8863-faddc51df038.png)

Good luck to you!

\- Joe

LinusU:
Also, this really should be supported directly by node-canvas to mimic how it's working in the browser. I'll try and add it to the 2.x release as soon as possible...

Something like this should work (although it doesn't at the moment)

```js
const { createCanvas, loadImage } = require('canvas')
const canvas = createCanvas(200, 200)
const ctx = canvas.getContext('2d')

loadImage('http://placehold.it/256x256/0000ff/ffffff&text=node-canvas').then((image) => {
  ctx.drawImage(image, 50, 0, 70, 70)

  console.log('<img src="' + canvas.toDataURL() + '" />')
})
```

ryuukibeat:
Following @jhuckaby method, I got this error.
```js
Error: Image given has not completed loading
```
It seems because I call it in asynchronous function. How can I resolve this? I add `await` in front of `request.get()` still not working.
___
node-canvas version: 2.0.0-alpha.1

[...]

PR #1003, opened 2017-09-16
Fix memory leak when img onerror/onload callbacks reference img

### LINKED ISSUE #150: Image Memory Leak
This simple little loop chews through all the memory on my Debian box. I'm nulling the Image instance every time through.
Running with node --trace_gc tst6.js shows that garbage collection is happening but this script will use 2GB of memory in short order then get killed. In real life the interval would be 10s

...

setInterval(timestampImage, 100);

function timestampImage() {
    var cameraImage = new Image;

```
cameraImage.onload = onLoad;
cameraImage.onerror = function(err) {
                          console.log(err);
                      };

cameraImage.src = blackJpeg;

cameraImage = null;
```

}

function onLoad() {

...

Put the image in a canvas, add a text timestamp, save

...

}

https://gist.github.com/1720778#file_timestamp.js

### LINKED ISSUE #785: Memory leak in Image
When constructing an Image, memory for that image seems to never be released:

``` js
var fs = require('fs');
var Image = Canvas.Image;

var filename = 'image.png';
var data = fs.readFileSync(filename);

function load() {
    var img = new Image;
    img.onload = function() {
        process.nextTick(load);
        global.gc();
    };
    img.src = data;
}

load();
```

This code uses `global.gc()` to force garbage collection, but even then, memory rises very quickly. When moving the `var img = new Image;` to outside the `load` function (and effectively reusing it) changes memory usage behavior to near constant.

zbjornson:
Storing the `onerror`/`onload` callback functions in Persistent handles (here, as `Nan::Callback`s) means they will never be GC'ed automatically; only when the Image destructor runs, which explicitly resets them. But, if those functions have references to the image, then the image will never be GC'ed. --> leak

I'm annoyed that I can't figure out how to do this from C++ after thinking about it all day. I have an [open message on the v8-users list](https://groups.google.com/forum/#!topic/v8-users/QtTwdok_MeQ) to see if there's a way, but this method also removes 87 lines of code. The only benefit to doing it in C++ is that we can type check the value and only set it if it's a function. Up to you if you want to wait a bit to see if someone responds.

Fixes #785
Fixes #150 (for real)

---

Code that repros the leak:

```js
const fs = require("fs");
const Canvas = require(".");

const imgsrc = fs.readFileSync("./test/fixtures/clock.png");

const canvas = new Canvas.Canvas(15000, 15000);
const ctx = canvas.getContext("2d");

var invocations = 0;

// Causes leak:
for (var k = 0; k < 5; k++) {
  gc();
  console.log(process.memoryUsage().rss);
  for (var i = 0; i < 1000; i++) {
    const img = new Canvas.Image();
    img.onload = _ => {
      if (img.width > 0) invocations++;
      // img.src= null; // frees a bunch of memory but doesn't actually invoke ~Image
    }
    img.src = imgsrc;
  }
}

// Doesn't cause leak:
// for (var k = 0; k < 5; k++) {
//   gc();
//   console.log(process.memoryUsage().rss);
//   for (var i = 0; i < 10000; i++) {
//     const img = new Canvas.Image();
//     img.onload = _ => invocations++;
//     img.src = imgsrc;
//   }
// }
```

In master:
24535040
466857984
908058624
1347592192

This PR:
24571904
26157056
26300416
25919488
24920064

### DIFF
--- a/src/Image.cc
+++ b/src/Image.cc
@@ -52,8 +52,6 @@ Image::Initialize(Nan::ADDON_REGISTER_FUNCTION_ARGS_TYPE target) {
   SetProtoAccessor(proto, Nan::New("height").ToLocalChecked(), GetHeight, NULL, ctor);
   SetProtoAccessor(proto, Nan::New("naturalWidth").ToLocalChecked(), GetNaturalWidth, NULL, ctor);
   SetProtoAccessor(proto, Nan::New("naturalHeight").ToLocalChecked(), GetNaturalHeight, NULL, ctor);
-  SetProtoAccessor(proto, Nan::New("onload").ToLocalChecked(), GetOnload, SetOnload, ctor);
-  SetProtoAccessor(proto, Nan::New("onerror").ToLocalChecked(), GetOnerror, SetOnerror, ctor);
 #if CAIRO_VERSION_MINOR >= 10
   SetProtoAccessor(proto, Nan::New("dataMode").ToLocalChecked(), GetDataMode, SetDataMode, ctor);
   ctor->Set(Nan::New("MODE_IMAGE").ToLocalChecked(), Nan::New<Number>(DATA_IMAGE));
@@ -74,6 +72,8 @@ NAN_METHOD(Image::New) {
   Image *img = new Image;
   img->data_mode = DATA_IMAGE;
   img->Wrap(info.This());
+  info.This()->Set(Nan::New("onload").ToLocalChecked(), Nan::Null());
+  info.This()->Set(Nan::New("onerror").ToLocalChecked(), Nan::Null());
   info.GetReturnValue().Set(info.This());
 }
 
@@ -233,9 +233,17 @@ NAN_SETTER(Image::SetSource) {
 
   // check status
   if (status) {
-    img->error(Canvas::Error(status));
+    Local<Value> onerrorFn = info.This()->Get(Nan::New("onerror").ToLocalChecked());
+    if (onerrorFn->IsFunction()) {
+      Local<Value> argv[1] = { Canvas::Error(status) };
+      onerrorFn.As<Function>()->Call(Isolate::GetCurrent()->GetCurrentContext()->Global(), 1, argv);
+    }
   } else {
     img->loaded();
+    Local<Value> onloadFn = info.This()->Get(Nan::New("onload").ToLocalChecked());
+    if (onloadFn->IsFunction()) {
+      onloadFn.As<Function>()->Call(Isolate::GetCurrent()->GetCurrentContext()->Global(), 0, NULL);
+    }
   }
 }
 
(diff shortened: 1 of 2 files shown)


zbjornson:
I think I understand this better now (or I'm even more confused and this is all wrong, but at the moment I've convinced myself). The first paragraph of my OP is correct and explains the leak. In addition...

node-canvas's `img.src=` leads to *synchronous* invocation of `onerror` or `onload`. In browsers it's async, even if you assign a `data:` URL. Two consequences of this are that (a) `onload`/`onerror` must be assigned before `img.src=`, and (b) using `onload` is actually unnecessary because when `img.src=` returns, the image is loaded.

A third consequence is that `img` will still be in scope and not eligible for GC after the src is loaded. (That's why this PR works.) If the callbacks were changed to be invoked asynchronously, it would go out of scope. The solution there that avoids the leak would be to store a persistent handle containing the `img.onload` value when `img.src = ...` is invoked, and then reset the handle when `onload` is invoked.

Summary:
* **This PR is good to go.** Even if we decide to make the callbacks async sometime, the old code was wrong. The only slightly gross thing about this PR is that the property is set directly on the object instead of through a setter that validates that the argument is a function, and instead I'm checking that it's a function right before invoking it. A JS setter would end up exposing some other property like `img._onload`; only way around this is a C++ setter AFAIK.

* We should decide before 2.0 lands if `img.src=` should become async to align with browsers.
    * I think this would break a lot of existing code based on how many github issues show code that assumes it's sync (#1001, #966, #959, #945, #848, #787, ... it's like a 5:1 ratio of code that assumes sync to not).
    * There's also the performance aspect: the async turn adds about a millisecond for an idle node.js process, longer for active processes. node-canvas supports `img.src = <filepath>|Buffer|dataURI`; the `<filename>` case would unblock the CPU during reading, but for the others, since the data is in memory and requires no IO, we're not necessarily letting any more work happen. Worse, we're potentially retaining large amounts of memory for longer periods of time. If we make the `img.src=` async, I'd be a strong advocate for adding a non-standard sync API like `image.load(Buffer|dataUriString)`. (See also the discussion starting here with @adamhooper about async performance and retained memory: https://github.com/Automattic/node-canvas/issues/743#issuecomment-324532996)
    * I have a [sloppy branch](https://github.com/zbjornson/node-canvas/tree/async-cb) that makes the handlers async that I was using for testing. It's currently in no man's land ... the loading is sync, but the handlers fire async.

LinusU:
I really think that it should be asynchronous, since in most cases it is performing blocking io as it is now. It will also be unacceptable to be synchronous when/if we support `.src = 'http://...'`

I think that the minimum we should to before releasing 2.0 is to just add `process.nextTick` or similar to delay the callbacks, that way it will at least behave asynchronous for the user; and we would be able to land proper async without breaking something...

> I think this would break a lot of existing code based on how many github issues show code that assumes it's sync

This is a bit unfortunate but I think 2.x is a good time to break things. In the end, I hope it will lead to better practices...

> There's also the performance aspect: the async turn adds about a millisecond for an idle node.js process, longer for active processes.

I feel that this comes down to io vs cpu. My opinion is that io work should be async and cpu work should be sync. That callbacks adds a bit of overhead is just how Node.js works, the same can be said for e.g. `fs.readFile`, `fs.readFileSync` is faster, but it's not concurrent which can yield more throughput when looking at the bigger picture...

> I'd be a strong advocate for adding a non-standard sync API like [...]

Sounds good 👍

zbjornson:
Yeah, that's most logical, I agree. Will open another ticket for it.

fuzhenn:
Thx for the excellent works!
We met this memleak problem in our production env
@LinusU not in a hurry, any plan to publish a new version to fix this recently?

LinusU:
published as 2.0.0-alpha.6

PR #1016, opened 2017-10-12
More accurate measuretext

chearon:
The values from `measureText()` are using Pango methods which round to the nearest pixel, since their use case is actual device pixels. This makes it like the browser and returns fractional pixels, which enables me to make text wrapping that matches the browser identically.

First commit is just minor cleanup

PS the measureText() tests for the additional TextMetrics properties are still failing on macOS

### DIFF
--- a/src/CanvasRenderingContext2d.cc
+++ b/src/CanvasRenderingContext2d.cc
@@ -2073,14 +2073,31 @@ NAN_METHOD(Context2d::MeasureText) {
   String::Utf8Value str(info[0]->ToString());
   Local<Object> obj = Nan::New<Object>();
 
-  PangoRectangle ink_rect, logical_rect;
+  PangoRectangle _ink_rect, _logical_rect;
+  float_rectangle ink_rect, logical_rect;
   PangoFontMetrics *metrics;
   PangoLayout *layout = context->layout();
 
   pango_layout_set_text(layout, *str, -1);
   pango_cairo_update_layout(ctx, layout);
 
-  pango_layout_get_pixel_extents(layout, &ink_rect, &logical_rect);
+  // Normally you could use pango_layout_get_pixel_extents and be done, or use
+  // pango_extents_to_pixels, but both of those round the pixels, so we have to
+  // divide by PANGO_SCALE manually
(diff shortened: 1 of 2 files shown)


src/CanvasRenderingContext2d.cc
```diff
@@ -2073,14 +2073,29 @@ NAN_METHOD(Context2d::MeasureText) {
...
+
+  logical_rect.x = 1.0 * _logical_rect.x / PANGO_SCALE;
+  logical_rect.y = 1.0 * _logical_rect.y / PANGO_SCALE;
+  logical_rect.width = 1.0 * _logical_rect.width / PANGO_SCALE;
+  logical_rect.height = 1.0 * _logical_rect.height / PANGO_SCALE;
+
+  ink_rect.x = 1.0 * _ink_rect.x / PANGO_SCALE;
+  ink_rect.y = 1.0 * _ink_rect.y / PANGO_SCALE;
+  ink_rect.width = 1.0 * _ink_rect.width / PANGO_SCALE;
+  ink_rect.height = 1.0 * _ink_rect.height / PANGO_SCALE;
```
zbjornson: Think it's preferable to explicitly cast one or both operands to a float/double. You could cast `PANGO_SCALE` once and be done.

chearon: Makes sense, updated PR.

chearon:
Before this PR (`measureText('Hello!')`):

```
{ width: 26,
  actualBoundingBoxLeft: 0,
  actualBoundingBoxRight: 26,
  actualBoundingBoxAscent: 7,
  actualBoundingBoxDescent: 1,
  emHeightAscent: 7,
  emHeightDescent: 3,
  alphabeticBaseline: -0 }
```

After:

```
{ width: 25.56640625,
  actualBoundingBoxLeft: 0,
  actualBoundingBoxRight: 25.56640625,
  actualBoundingBoxAscent: 6.47265625,
  actualBoundingBoxDescent: 0.890625,
  emHeightAscent: 7,
  emHeightDescent: 3,
  alphabeticBaseline: -0.7001953125 }
```

LinusU:
Cheers!

PR #1017, opened 2017-10-12
fix incorrect path to cairo

chearon:
I'm not sure what's going on here, but

* `HAVE_PANGO` is not a thing anymore
* `cairo/cario.h` is not the correct path to Cairo

Not sure how it was building successfully, but one time it failed on me and this fixed it


zbjornson:
✔️ builds on Windows.

LinusU:
Neat 👍

PR #1019, opened 2017-10-16
Fix incorrect dependency for Ubuntu in readme

lustoykov:
Replace libjpeg8-dev with libjpeg62-turbo-dev

This happened on one of our machines (docker node 8.4.0 base image)

> Reading state information...
>   Package libjpeg8-dev is not available, but is referred to by another package.
>   This may mean that the package is missing, has been obsoleted, or
>   is only available from another source
>   However the following packages replace it:
>     libjpeg62-turbo-dev
>   
>  [91mE: Package ‘libjpeg8-dev’ has no installation candidate
>   [0mThe command ‘/bin/sh -c apt-get update && apt-get install -y libcairo2-dev libjpeg8-dev libpango1.0-dev libgif-dev build-essential g++’ returned a non-zero code: 100

Replacing libjpeg8-dev with libjpeg62-turbo-dev fixed the problem.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -36,7 +36,7 @@ You can quickly install the dependencies by using the command for your OS:
 OS | Command
 ----- | -----
 OS X | `brew install pkg-config cairo pango libpng jpeg giflib`
-Ubuntu | `sudo apt-get install libcairo2-dev libjpeg8-dev libpango1.0-dev libgif-dev build-essential g++`
+Ubuntu | `sudo apt-get install libcairo2-dev libjpeg-dev libpango1.0-dev libgif-dev build-essential g++`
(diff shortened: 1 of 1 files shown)


zbjornson:
I think the correct fix is actually to use `libjpeg-dev`, which is a dummy package that points to libjpeg8 for pre-Jessie and libjpeg62-turbo for Jessie and later: https://packages.debian.org/search?keywords=libjpeg

lustoykov:
@zbjornson thanks, makes sense, I will verify that and create a new PR in next few days. Feel free to close this one.

zbjornson:
They've deleted their fork. I just opened two PRs (one for 2.x and one for 1.x) to replace this one in the interest of time.

lustoykov:
Thanks guys, sorry I was inactive!

PR #1022, opened 2017-10-27
Fix reading fillStyle after setting it from gradient to color

caldwell:
The getter for `fillStyle` in context2d.js prioritizes `this.lastFillStyle` over `this.fillColor`, so it needs to be cleared when we set `fillStyle` to a color.

I modified the linear gradient test so that it shows the behavior.

### DIFF
--- a/lib/context2d.js
+++ b/lib/context2d.js
@@ -187,6 +187,7 @@ Context2d.prototype.__defineSetter__('fillStyle', function(val){
     this.lastFillStyle = val;
     this._setFillPattern(val);
   } else if ('string' == typeof val) {
+    this.lastFillStyle = undefined;
     this._setFillColor(val);
   }
 });
--- a/test/public/tests.js
+++ b/test/public/tests.js
@@ -387,6 +387,10 @@ tests['createLinearGradient()'] = function(ctx){
 
   ctx.fillRect(10,10,130,130);
   ctx.strokeRect(50,50,50,50);
+
+  ctx.fillStyle = '#13b575';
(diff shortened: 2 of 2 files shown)


LinusU:
Thank you!

Just need to figure out why Travis is failing 😞

zbjornson:
Version 8 of "standard" dropped support of Node.js <4.0.

LinusU:
I'm not sure that that is the actual problem, since Standard is supposed to bail out and exit with a `0` (success) status code on older nodes 🤔 

> standard: Node v4 or greater is required. `standard` did not run.

LinusU:
Hmm, it seems like a newer version of `supports-color` is being pulled in, although I don't really know why since we specify `"mocha": "^3.1.2"` which should resolve to 3.5.3 which specifies `"supports-color": "3.1.2"` which doesn't use `const` on line 2 of index.js 🤔 

> /home/travis/build/Automattic/node-canvas/node_modules/mocha/node_modules/supports-color/index.js:2
> const os = require('os');
> ^^^^^
> SyntaxError: Use of const in strict mode.

zbjornson:
Derp, my bad.

Invoking old node with `--harmony` I think allows that to work, but it might not address the real issue as you just pointed out.

zbjornson:
CI passes after a rebase, since Linus fixed it :). https://travis-ci.org/zbjornson/node-canvas/builds/319985242

zbjornson:
Ported to master/2.x as b18ae15

PR #1028, opened 2017-11-01
Fix two typos in require statements

zbjornson:
Somehow I missed having a test for `context2d.currentTransform` :(.


LinusU:
Great 👍

PR #1034, opened 2017-11-13
Add unit test for valid jpeg EOI marker

binarykitchen:


LinusU:
Could you not add the `yarn.lock` file? Not everyone is using yarn so it will quickly get out of date... It's also not as important in a library as in an app...

binarykitchen:
@LinusU done, removed

LinusU:
Awesome, thanks you!

PR #1038, opened 2017-11-16
use the PangoLayout to get baseline

### LINKED ISSUE #1037: drawText location incorrect on scaled Canvas (CentOS 7 and Ubuntu 16)
## Issue

This is a special case of #903 . I'm opening a new issue since I can't reopen that one. Feel free to close if this is the wrong way to surface the problem! 

I'm having the same issue on CentOS 7, and @zbjornson commented saying that he has encountered it on Ubuntu 16. See #903 for the results of some preliminary investigation.

## Your Environment
* Version of node-canvas: 1.6.7
* Environment: node v6.11.5 on CentOS 7

chearon:
I believe this fixes #1037, I wasn't able to reproduce it on Linux but the baselines all seem fine at different rotations/scales for me.

I need to fix the measureText tests too, the implementation is summing ink and logical rects so something is wrong there.

### DIFF
--- a/src/CanvasRenderingContext2d.cc
+++ b/src/CanvasRenderingContext2d.cc
@@ -1945,19 +1945,23 @@ NAN_METHOD(Context2d::StrokeText) {
 }
 
 /*
- * Gets the baseline adjustment in device pixels, taking into account the
- * transformation matrix. TODO This does not handle skew (which cannot easily
- * be extracted from the matrix separately from rotation).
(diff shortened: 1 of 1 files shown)


chearon:
Tests passing now. macOS users will notice that the baseline now matches the browser. Needs a Windows check (paging @zbjornson)

LinusU:
Awesome!

Issue #289, opened 2013-05-30
Image given has not completed loading


tarunkurapati:
Error:
      Error: Image given has not completed loading
    at /home/f-user/passport-facebook/examples/login/app.js:315:17
    at fs.readFile (fs.js:176:14)
    at Object.oncomplete (fs.js:297:15) 

Code:

var canvas = new Canvas(650, 650)
                               , ctx = canvas.getContext('2d');
fs.readFile(__dirname + '/public/resimg/1.jpg', function(err, squid){
      if (err) throw err;
        img = new Canvas.Image;
          img.src = squid;
          img.onload = function(){
              console.log("ian here in onload");
            ctx.drawImage(img, 0, 0, img.width / 4, img.height / 4);
          }
});

I have seen similar issues here, they mentioned to use "img.onload" in case we get this error "Image given has not completed loading"

But img.onload is not getting fired in my case.
May i know how to fix for this issue.
Thanks

redspider:
For everybody on this ticket and for all those who will arrive in the future, please note that this error message MAY occur because you are attempting to load an image type that canvas does not support - probably due to a limited cairo build. You can verify this by trying to load a PNG instead which is typically supported.

dosapati:
Having the same issue to load jpg images. Tried installing libjpeg through brew still no luck. Works for PNG images and not for JPG images

noam3127:
I was having this same problem, and the solution given by **yhostc** discussed in [this issue](https://github.com/Automattic/node-canvas/issues/348) ended up solving it:
brew update
brew doctor
xcode-select --install
brew install cairo gobject-introspection pixman

tomerb15:
@redspider How can we add JPEG support for node-canvas?

redspider:
@tomerb15 node-canvas doesn't care about image types, if I recall correctly it is cairo that cares (cairo is a dependency). You have to make sure your cairo build has JPEG enabled. I can't help you with that though.

coolfeature:
I was having the issue every time when I tried to render the PNG image below onto the canvas.  

![withinkscapechunk](https://cloud.githubusercontent.com/assets/6413939/13752268/df832f32-ea05-11e5-99ac-628fd9df2917.png)

I noticed that the rendering worked fine for other PNG images. So I used a hex editor (http://entropymine.com/jason/tweakpng/) to see what is different about the file. 

The obvious difference was the header:
 
[...]

PR #1044, opened 2017-11-29
Add /usr/lib64 to the list of system paths

azangru:
**Problem:** On CentOS, node-canvas is not supporting jpeg images (we were getting the `Image given has not completed loading` error discussed in other issues (such as https://github.com/Automattic/node-canvas/issues/289)

**Cause:** The cause of the particular error that we were experiencing was that, on CentOS, the `libjpeg-turbo-devel` library is installed in the `/usr/lib64` folder, which is not listed among the paths for lib checking.

**Solution:** This PR adds `/usr/lib64` to the list of paths that are checked for the presence of system libs.

### DIFF
--- a/util/has_lib.js
+++ b/util/has_lib.js
@@ -5,6 +5,7 @@ var childProcess = require('child_process')
 var SYSTEM_PATHS = [
   '/lib',
   '/usr/lib',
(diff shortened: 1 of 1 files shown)


LinusU:
Neat!

PR #1049, opened 2017-12-11
Make Context2d's currentTransform configurable

targos:
This allows the library to be used with the jest test runner.

See https://github.com/targos/canvas-jest for a simple repro.


LinusU:
cool

LinusU:
Do you want this in 1.x as well?

zbjornson:
Does jest actually work with this module at all? I thought it breaks with native dependencies because it tampers with the require cache.

targos:
Thanks!

> Do you want this in 1.x as well?

I think 1.x is not affected by this.

> Does jest actually work with this module at all?

Yes, with this change it works.

LinusU:
Published as 2.0.0-alpha.8

PR #1053, opened 2017-12-15
Add install instructions for MacPorts on OS X

thenickdude:
Also adds a link for MacPorts/Homebrew sites for those not already familiar with these package managers.


LinusU:
Nice!

PR #1054, opened 2017-12-17
Sanitize some parameters before HasInstance calls

madprog:
Some parameters were passed to HasInstance without being checked first.
If they were null, this would crash the entire nodejs engine.
Throw a proper exception instead.

I actually had my tests fail on this, and I had a hard time finding where this was from, because the output was actually not very explicit:
```
% jest tests/unit/classification-pane.spec.js

 RUNS  tests/unit/classification-pane.spec.js
zsh: segmentation fault (core dumped)  node_modules/.bin/jest 
```

It seems that fetching an image resulted in an error because of the test environment, and that my code resulted in calling `DrawImage` with a `null` (or `undefined`?) image.
This patch fixes the issue, and prevents it for a few other functions.
Hope it helps someone not having my gdb-fu 😉

### DIFF
--- a/src/CanvasRenderingContext2d.cc
+++ b/src/CanvasRenderingContext2d.cc
@@ -552,6 +552,8 @@ NAN_METHOD(Context2d::New) {
     return Nan::ThrowTypeError("Class constructors cannot be invoked without 'new'");
   }
 
+  if (!info[0]->IsObject())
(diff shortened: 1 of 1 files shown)


zbjornson:
Looks good. There are a few other missing `HasInstance` checks we need to add sometime, but this is a definite improvement.

madprog:
Thank you! 😃

PR #1062, opened 2017-12-21
Fix Node.js 0.x compatibility

LinusU:


LinusU:
Seems like we broke Node.js 0.x compat without anyone noticing 😂

PR #1064, opened 2017-12-21
Add support for context globalAlpha for gradients and patterns

### LINKED ISSUE #1061: Gradients and Patterns do not work with globalAlpha
## Issue or Feature

Gradients and patterns do not work with the ctx.globalAlpha. they are always solid no matter what.

I have a working fix that i'm cleaning up and submitting for approval

## Steps to Reproduce

I extended the current test.js in this way to verify:

```js
tests['createPattern() with opacity'] = function(ctx, done) {
  var img = new Image;
  img.onload = function(){
    var pattern = ctx.createPattern(img, 'repeat');
    ctx.globalAlpha = 0.2;
    ctx.scale(0.1, 0.1)
    ctx.fillStyle = pattern;
    ctx.fillRect(0, 0, 2000, 2000);
    done();
  };
  img.src = 'face.jpeg';
};

tests['createLinearGradient() with opacity'] = function(ctx){
  ctx.globalAlpha = 0.2;
  var lingrad = ctx.createLinearGradient(0,0,0,150);
  lingrad.addColorStop(0, '#00ABEB');
  lingrad.addColorStop(0.5, '#fff');
  lingrad.addColorStop(0.5, '#26C000');
  lingrad.addColorStop(1, '#fff');

  var lingrad2 = ctx.createLinearGradient(0,50,0,95);
  lingrad2.addColorStop(0.5, '#000');
  lingrad2.addColorStop(1, 'rgba(0,0,0,0)');

  ctx.fillStyle = lingrad;
  ctx.strokeStyle = lingrad2;
  ctx.globalAlpha = 0.2;
  ctx.fillRect(10,10,130,130);
  ctx.strokeRect(50,50,50,50);
};
```

## Result
image

## Your Environment
* Version of node-canvas (e.g. 1.6.8):
* Environment (e.g. node 6.9.5 on Mac OS X Sierra):

asturur:
close #1061

So this is another discrepancy i found out inspecting differences between browser canvas and node-canvas.

Pattern and gradients are always solid.

Now i added those tests here:
Pattern fil and stroke, with and without opacity
image

Gradients with rgba, gradients and globalAlpha and gradients and trasnforms
image

The code is a bit ... meh ...
The point of the fix for patterns i think is ok, i pick up the chosen pattern, i create a new surface that is as big as the pattern surface, i paint it with alpha, i create a new pattern from that surface.
This is the only thing that was working for me.

For Gradients i do same, create a new gradient, multiply alpha and move on,

I just created 2 functions in the the code, i did not make them class method, i m not a C++ or node-gyp dev, so this is what i could do, i m open to suggestion to improve it.

I also think i have to destroy some resources after this process, any hint how/where?

### DIFF
--- a/src/CanvasRenderingContext2d.cc
+++ b/src/CanvasRenderingContext2d.cc
@@ -339,10 +339,16 @@ create_transparent_pattern(cairo_pattern_t *source, float alpha) {
     width,
     height);
   cairo_t *mask_context = cairo_create(mask_surface);
+  if (cairo_status(mask_context) != CAIRO_STATUS_SUCCESS) {
+    // context creation failed. Return the original source and destroy the surface
+    cairo_surface_destroy(mask_surface);
+    return source;
+  }
   cairo_set_source(mask_context, source);
   cairo_paint_with_alpha(mask_context, alpha);
+  // remove the context, but not the surface since is owned by the pattern
+  cairo_destroy(mask_context);
   cairo_pattern_t* newPattern = cairo_pattern_create_for_surface(mask_surface);
-  cairo_surface_destroy(mask_surface);
   return newPattern;
 }
 
@@ -396,7 +402,7 @@ Context2d::fill(bool preserve) {
       ? shadow(cairo_fill)
       : cairo_fill(_context);
   }
-  if (new_pattern != nullptr) {
+  if (new_pattern != state->fillPattern && new_pattern != nullptr) {
     cairo_pattern_destroy(new_pattern);
   }
 }
@@ -438,7 +444,7 @@ Context2d::stroke(bool preserve) {
       ? shadow(cairo_stroke)
       : cairo_stroke(_context);
   }
-  if (new_pattern != nullptr) {
+  if (new_pattern != state->strokePattern && new_pattern != nullptr) {
     cairo_pattern_destroy(new_pattern);
   }
 }
--- a/src/CanvasRenderingContext2d.cc
+++ b/src/CanvasRenderingContext2d.cc
@@ -341,12 +341,10 @@ create_transparent_pattern(cairo_pattern_t *source, float alpha) {
     height);
   cairo_t *mask_context = cairo_create(mask_surface);
   if (cairo_status(mask_context) != CAIRO_STATUS_SUCCESS) {
-    // context creation failed. Return the original source and destroy the surface
     Nan::ThrowError("Failed to initialize context");
   }
   cairo_set_source(mask_context, source);
   cairo_paint_with_alpha(mask_context, alpha);
-  // remove the context, but not the surface since is owned by the pattern
   cairo_destroy(mask_context);
   cairo_pattern_t* newPattern = cairo_pattern_create_for_surface(mask_surface);
   cairo_surface_destroy(mask_surface);
--- a/src/CanvasRenderingContext2d.cc
+++ b/src/CanvasRenderingContext2d.cc
@@ -370,7 +370,7 @@ Context2d::setFillRule(v8::Local<v8::Value> value) {
 
 void
 Context2d::fill(bool preserve) {
-  cairo_pattern_t *new_pattern = nullptr;
+  cairo_pattern_t *new_pattern = NULL;
   if (state->fillPattern) {
     if (state->globalAlpha < 1) {
       new_pattern = create_transparent_pattern(state->fillPattern, state->globalAlpha);
@@ -402,7 +402,7 @@ Context2d::fill(bool preserve) {
       ? shadow(cairo_fill)
       : cairo_fill(_context);
   }
-  if (new_pattern != state->fillPattern && new_pattern != nullptr) {
+  if (new_pattern != state->fillPattern && new_pattern != NULL) {
     cairo_pattern_destroy(new_pattern);
   }
 }
@@ -413,7 +413,7 @@ Context2d::fill(bool preserve) {
 
 void
 Context2d::stroke(bool preserve) {
-  cairo_pattern_t *new_pattern = nullptr;
+  cairo_pattern_t *new_pattern = NULL;
   if (state->strokePattern) {
     if (state->globalAlpha < 1) {
       new_pattern = create_transparent_pattern(state->strokePattern, state->globalAlpha);
@@ -444,7 +444,7 @@ Context2d::stroke(bool preserve) {
       ? shadow(cairo_stroke)
       : cairo_stroke(_context);
   }
-  if (new_pattern != state->strokePattern && new_pattern != nullptr) {
+  if (new_pattern != state->strokePattern && new_pattern != NULL) {
     cairo_pattern_destroy(new_pattern);
   }
 }
--- a/src/CanvasRenderingContext2d.cc
+++ b/src/CanvasRenderingContext2d.cc
@@ -321,6 +321,7 @@ create_transparent_gradient(cairo_pattern_t *source, float alpha) {
     newGradient = cairo_pattern_create_radial(x0, y0, r0, x1, y1, r1);
   } else {
     Nan::ThrowError("Unexpected gradient type");
+    return source;
   }
   for ( i = 0; i < count; i++ ) {
     cairo_pattern_get_color_stop_rgba(source, i, &offset, &r, &g, &b, &a);
@@ -342,6 +343,7 @@ create_transparent_pattern(cairo_pattern_t *source, float alpha) {
   cairo_t *mask_context = cairo_create(mask_surface);
   if (cairo_status(mask_context) != CAIRO_STATUS_SUCCESS) {
     Nan::ThrowError("Failed to initialize context");
+    return source;
   }
   cairo_set_source(mask_context, source);
   cairo_paint_with_alpha(mask_context, alpha);
@@ -374,7 +376,9 @@ Context2d::fill(bool preserve) {
     if (state->globalAlpha < 1) {
       new_pattern = create_transparent_pattern(state->fillPattern, state->globalAlpha);
       cairo_set_source(_context, new_pattern);
-      cairo_pattern_destroy(new_pattern);
+      if (new_pattern != state->fillPattern) {
+        cairo_pattern_destroy(new_pattern);
+      }
     } else {
       cairo_set_source(_context, state->fillPattern);
     }
@@ -386,7 +390,9 @@ Context2d::fill(bool preserve) {
       new_pattern = create_transparent_gradient(state->fillGradient, state->globalAlpha);
       cairo_pattern_set_filter(new_pattern, state->patternQuality);
       cairo_set_source(_context, new_pattern);
-      cairo_pattern_destroy(new_pattern);
+      if (new_pattern != state->fillGradient) {
+        cairo_pattern_destroy(new_pattern);
+      }
     } else {
       cairo_pattern_set_filter(state->fillGradient, state->patternQuality);
       cairo_set_source(_context, state->fillGradient);
@@ -416,7 +422,9 @@ Context2d::stroke(bool preserve) {
     if (state->globalAlpha < 1) {
       new_pattern = create_transparent_pattern(state->strokePattern, state->globalAlpha);
       cairo_set_source(_context, new_pattern);
-      cairo_pattern_destroy(new_pattern);
+      if (new_pattern != state->strokePattern) {
+        cairo_pattern_destroy(new_pattern);
+      }
     } else {
       cairo_set_source(_context, state->strokePattern);
     }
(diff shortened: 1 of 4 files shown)


asturur:
well i just realized my fix does not work with alpha and transforms.
image

gonna work over it to see if how i can fix it.

i think my surface fix is crap and is better i create a new gradient multiplying alpha values.

src/CanvasRenderingContext2d.cc
```diff
@@ -292,6 +292,58 @@ Context2d::restorePath() {
...
+  int i;
+  double offset;
+  double r;
+  double g;
+  double b;
+  double a;
+  cairo_pattern_t *newGradient;
+  cairo_pattern_type_t type = cairo_pattern_get_type(source);
+  cairo_pattern_get_color_stop_count(source, &count);
+  if (type == 2) {
```
zbjornson: can you use the `cairo_pattern_type_t` constants instead? https://developer.gnome.org/cairo/stable/cairo-cairo-pattern-t.html#cairo-pattern-type-t

src/CanvasRenderingContext2d.cc
```diff
@@ -292,6 +292,58 @@ Context2d::restorePath() {
...
+    cairo_pattern_get_linear_points (source, &x0, &y0, &x1, &y1);
```
zbjornson: nit: no space between fn name and parentheses
same a few lines down (322)

src/CanvasRenderingContext2d.cc
```diff
@@ -311,16 +363,27 @@ Context2d::setFillRule(v8::Local<v8::Value> value) {
...
+      cairo_pattern_set_filter(new_gradient, state->patternQuality);
+      cairo_set_source(_context, new_gradient);
+    } else {
+      cairo_pattern_set_filter(state->fillGradient, state->patternQuality);
+      cairo_set_source(_context, state->fillGradient);
+    }
   } else {
     setSourceRGBA(state->fill);
   }
-
```
zbjornson: I can't comment far down enough, but after the `shadow`/`fill` calls a few lines down, you need to free the `new_pattern` or `new_gradient` `cairo_pattern_t`s with https://developer.gnome.org/cairo/stable/cairo-cairo-pattern-t.html#cairo-pattern-destroy.

src/CanvasRenderingContext2d.cc
```diff
@@ -339,11 +402,22 @@ Context2d::fill(bool preserve) {
...
-    cairo_pattern_set_filter(state->strokeGradient, state->patternQuality);
-    cairo_set_source(_context, state->strokeGradient);
+    if (state->globalAlpha < 1) {
+      cairo_pattern_t *new_gradient = create_transparent_gradient(state->strokeGradient, state->globalAlpha);
+      cairo_pattern_set_filter(new_gradient, state->patternQuality);
+      cairo_set_source(_context, new_gradient);
+    } else {
+      cairo_pattern_set_filter(state->strokeGradient, state->patternQuality);
+      cairo_set_source(_context, state->strokeGradient);
+    }
```
zbjornson: Ditto about freeing the `cairo_pattern_t`s at the end of this function.

src/CanvasRenderingContext2d.cc
```diff
@@ -292,6 +292,58 @@ Context2d::restorePath() {
...
+  return newGradient;
+}
+
+cairo_pattern_t*
+create_transparent_pattern(cairo_pattern_t *source, float alpha) {
+  cairo_surface_t *surface;
+  cairo_pattern_get_surface(source, &surface);
+  int width = cairo_image_surface_get_width(surface);
+  int height = cairo_image_surface_get_height(surface);
+  cairo_surface_t *mask_surface = cairo_image_surface_create(
```
zbjornson: `*mask_surface` needs to be freed at some point. I *think* it's safe to call [`cairo_surface_destroy`](https://www.cairographics.org/manual/cairo-cairo-surface-t.html#cairo-surface-destroy) just after your `cairo_pattern_create_for_surface(mask_surface)` call, so you can do memory management within this function.

src/CanvasRenderingContext2d.cc
```diff
@@ -292,6 +292,58 @@ Context2d::restorePath() {
...
+    CAIRO_FORMAT_ARGB32,
+    width,
+    height);
+  cairo_t *mask_context = cairo_create(mask_surface);
```
zbjornson: `*mask_context` needs to be freed. Looks like you don't need it after `cairo_paint_with_alpha` so that's easy.

You should also check [`cairo_status(mask_context)`](https://www.cairographics.org/manual/cairo-cairo-t.html#cairo-status) since this could fail to allocate.

asturur:
I think i solved all the comments @zbjornson

src/CanvasRenderingContext2d.cc
```diff
@@ -292,6 +292,66 @@ Context2d::restorePath() {
...
+  cairo_pattern_type_t type = cairo_pattern_get_type(source);
+  cairo_pattern_get_color_stop_count(source, &count);
+  if (type == CAIRO_PATTERN_TYPE_LINEAR) {
+    cairo_pattern_get_linear_points (source, &x0, &y0, &x1, &y1);
+    newGradient = cairo_pattern_create_linear(x0, y0, x1, y1);
+  }
+  if (type == CAIRO_PATTERN_TYPE_RADIAL) {
+    cairo_pattern_get_radial_circles(source, &x0, &y0, &r0, &x1, &y1, &r1);
+    newGradient = cairo_pattern_create_radial(x0, y0, r0, x1, y1, r1);
+  }
```
zbjornson: To get rid of the maybe-uninitialized warning here, can you make this `if (linear) {} else if (raidal) {} else { do something reasonable like throwing an error }` please?

src/CanvasRenderingContext2d.cc
zbjornson: I still don't see this being freed... calling `cairo_surface_destroy` decreases the ref count, but doesn't destroy it unless the ref count is 0. `cairo_pattern_create_for_surface` [increases the ref count on the surface](https://github.molgen.mpg.de/git-mirror/cairo/blob/747cab741cf63f30aa6bc2b787fc13e7e7a2b859/src/cairo-pattern.c#L749). There's an example of this exact lifecycle pattern in the Cairo source code [here](https://github.molgen.mpg.de/git-mirror/cairo/blob/747cab741cf63f30aa6bc2b787fc13e7e7a2b859/src/cairo-default-context.c#L238-L255). Make sense?

src/CanvasRenderingContext2d.cc
```diff
@@ -330,6 +402,9 @@ Context2d::fill(bool preserve) {
       ? shadow(cairo_fill)
       : cairo_fill(_context);
   }
+  if (new_pattern != state->fillPattern && new_pattern != NULL) {
```
zbjornson: I think the `new_pattern != state->fillPattern` part of this makes the `cairo_pattern_destroy` unreachable in cases of global alpha < 1. Here again the `destroy` fn actually just decreases ref count; it should be safe to call it whenever `new_pattern != NULL`.

Actually, you can make the `destroy` call after line 388 I think, right after `cairo_set_source(_context, new_pattern)`. That'll let you shrink the scope of `new_pattern` as well.

src/CanvasRenderingContext2d.cc
```diff
@@ -357,6 +444,9 @@ Context2d::stroke(bool preserve) {
       ? shadow(cairo_stroke)
       : cairo_stroke(_context);
   }
+  if (new_pattern != state->strokePattern && new_pattern != NULL) {
```
zbjornson: Ditto

src/CanvasRenderingContext2d.cc
asturur: i did not get any warning, but definitely i can guard against it.

src/CanvasRenderingContext2d.cc
asturur: i thought that since the pattern is alive and to be used you have to keep the surface.
I get this feeling from the sentence 'the pattern owns the surface'  but maybe i m wrong. I'll try to delete everything, if it still draws is ok.

src/CanvasRenderingContext2d.cc
asturur: i added this check because the new_pattern can be NULL globalAlpha = 1 or can be fillPattern if the creation of the new pattern failed ( because we return the same pattern from the other function ).
Why you think is unreachable?

asturur:
I now essentially moved the destroy call right after pattern usage.
I m also now trowing errors as soon as something goes wrong, i will not return the original pattern if creating the new one fails, and i will not check for the equality later.
Seems to render everything fine anyway.

src/CanvasRenderingContext2d.cc
```diff
@@ -292,6 +292,65 @@ Context2d::restorePath() {
...
+  cairo_pattern_type_t type = cairo_pattern_get_type(source);
+  cairo_pattern_get_color_stop_count(source, &count);
+  if (type == CAIRO_PATTERN_TYPE_LINEAR) {
+    cairo_pattern_get_linear_points (source, &x0, &y0, &x1, &y1);
+    newGradient = cairo_pattern_create_linear(x0, y0, x1, y1);
+  } else if (type == CAIRO_PATTERN_TYPE_RADIAL) {
+    cairo_pattern_get_radial_circles(source, &x0, &y0, &r0, &x1, &y1, &r1);
+    newGradient = cairo_pattern_create_radial(x0, y0, r0, x1, y1, r1);
+  } else {
+    Nan::ThrowError("Unexpected gradient type");
```
LinusU: I don't remember exactly how this works but it feels like there should be a `return` statement here:

```c++
Nan::ThrowError("Unexpected gradient type");
return nullptr;
```

Hmm, actually, possibly that needs to be guarded at the call-site as well? I don't see how `Nan::ThrowError` can alter the normal flow of a c++ function, but maybe I'm missing something 🤔

asturur: if i only knew c++... i have C knowledge coming from embedded platforms with proprietary api. I stop there.

How can i test it? i can force a gradient type comparision and see what happens?

LinusU: > i can force a gradient type comparision and see what happens?

that sounds good 👍

asturur: oh by the way using nullptr kills nodejs < 4.

asturur:
image

this is the best error managment i can get.
The server do not crash, it continues staying alive, it spit an error in the console. Rendering do not happen:

image

i triggered the error in this way:
image

adding a number to the enums so they do not match.

asturur:
image

this is what i can get for patterns.

src/CanvasRenderingContext2d.cc
```diff
@@ -310,17 +371,35 @@ Context2d::setFillRule(v8::Local<v8::Value> value) {
...
 Context2d::fill(bool preserve) {
+  cairo_pattern_t *new_pattern;
   if (state->fillPattern) {
-    cairo_set_source(_context, state->fillPattern);
+    if (state->globalAlpha < 1) {
+      new_pattern = create_transparent_pattern(state->fillPattern, state->globalAlpha);
+      cairo_set_source(_context, new_pattern);
+      if (new_pattern != state->fillPattern) {
+        cairo_pattern_destroy(new_pattern);
+      }
```
zbjornson: Looks great! Last comment... do I have this right: `new_pattern == state->fillPattern` only if `create_transparent_pattern` fails? If so, then I'd consider changing this (and the equivalent other blocks of code) to:

```cpp
new_pattern = create_transparent_pattern(state->fillPattern, state->globalAlpha);
if (new_pattern == state->fillPattern) {
  // failed to allocate; Nan::ThrowError has already been called, so return from this fn.
  return;
}
cairo_set_source(_context, new_pattern);
cairo_pattern_destroy(new_pattern);
```

I think that clarifies the flow a bit, and explicitly stops execution if we've reached an error state.

asturur: and what about gradient? gradient has a similiar case, a slower chance to fail since i have no idea how you can create gradients type that are not linear or radial, but if this happen,still we are in a sort of error state, or at least rendering something different from what is expected.
Do we want to return also in that case?

zbjornson: I think you can do the same thing...

```cpp
new_pattern = create_transparent_gradient(state->fillGradient, state->globalAlpha);
if (new_pattern == state->fillGradient) return;
cairo_pattern_set_filter(new_pattern, state->patternQuality);
cairo_set_source(_context, new_pattern);
```

Also, since you're now bailing out in case of error, you might as well return `NULL` from `create_transparent_gradient`/`create_transparent_pattern` in case of error, and then just check if `new_pattern == NULL`.

asturur: but can i return null from a function that expect a pointer to cairo_pattern_t?

zbjornson: Yep!

asturur:
Ok, i used NULL the compiler does not complain.

zbjornson:
Ported to master/v2.x in d17e9ce. Thanks!

PR #1066, opened 2017-12-22
Add support for pattern repeat and no-repeat

asturur:
This PR aims at adding support for the data structure that contains repeat and no-repeat.
Cases like repeat-x and repeat-y needs extra rendering processing and i will try to fix them later on.
(I'm not sure i can do that with my cairo skills.)

Added a small test for it
image

### DIFF
--- a/lib/context2d.js
+++ b/lib/context2d.js
@@ -66,16 +66,16 @@ Context2d.prototype.__defineGetter__('imageSmoothingEnabled', function(val){
  * @api public
  */
 
-Context2d.prototype.createPattern = function(image, repetition){
-  repeat = 1;
-  if (repetition === 'no-repeat') {
-    repeat = 0;
-  } else if (repetition === 'repeat-x') {
-    repeat = 2;
-  } else if (repetition === 'repeat-y') {
-    repeat = 3;
+Context2d.prototype.createPattern = function(image, repeat){
+  var _repeat = 1;
+  if (repeat === 'no-repeat') {
+    _repeat = 0;
+  } else if (repeat === 'repeat-x') {
+    _repeat = 2;
+  } else if (repeat === 'repeat-y') {
+    _repeat = 3;
   }
-  return new CanvasPattern(image, repeat);
+  return new CanvasPattern(image, _repeat);
 };
 
 /**
--- a/lib/context2d.js
+++ b/lib/context2d.js
@@ -136,6 +136,7 @@ Context2d.prototype.__defineSetter__('fillStyle', function(val){
     this.lastFillStyle = val;
     this._setFillPattern(val);
   } else if ('string' == typeof val) {
+    this.lastFillStyle = undefined;
     this._setFillColor(val);
   }
 });
--- a/lib/context2d.js
+++ b/lib/context2d.js
@@ -66,16 +66,8 @@ Context2d.prototype.__defineGetter__('imageSmoothingEnabled', function(val){
  * @api public
  */
 
-Context2d.prototype.createPattern = function(image, repeat){
-  var _repeat = 1;
-  if (repeat === 'no-repeat') {
-    _repeat = 0;
-  } else if (repeat === 'repeat-x') {
-    _repeat = 2;
-  } else if (repeat === 'repeat-y') {
-    _repeat = 3;
-  }
-  return new CanvasPattern(image, _repeat);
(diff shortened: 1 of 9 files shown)


src/CanvasRenderingContext2d.cc
```diff
@@ -312,8 +312,13 @@ void
 Context2d::fill(bool preserve) {
   if (state->fillPattern) {
     cairo_set_source(_context, state->fillPattern);
-    cairo_pattern_set_extend(cairo_get_source(_context), CAIRO_EXTEND_REPEAT);
-    // TODO repeat/repeat-x/repeat-y
+    repeat_type_t *repeat = static_cast<repeat_type_t *>(cairo_pattern_get_user_data(state->fillPattern, pattern_repeat_key));
```
asturur: any lint rule here for this long line? any better way to write this thing here?

zbjornson: You could do the cast on a separate line perhaps. Also I'd store the pointer's value ASAP.

```cpp
void *ud = cairo_pattern_get_user_data(state->fillPattern, pattern_repeat_key);
repeat_type_t repeat = *reinterpret_cast<repeat_type_t*>(ud);
```

Since you do this twice, you might want to put this in a static method on `Pattern` (like `get_repeat_type_for_cairo_pattern`) and make `pattern_repeat_key` `static const` (better encapsulation).

src/CanvasPattern.cc
```diff
@@ -73,8 +75,10 @@ NAN_METHOD(Pattern::New) {
  * Initialize linear gradient.
  */
 
-Pattern::Pattern(cairo_surface_t *surface) {
+Pattern::Pattern(cairo_surface_t *surface, repeat_type_t *repeat) {
   _pattern = cairo_pattern_create_for_surface(surface);
+  _repeat = repeat;
```
zbjornson: (continued) the 2nd arg would be `repeat_type_t repeat`, and then this constructor would set the pointer value.

src/CanvasPattern.cc
```diff
@@ -83,4 +87,5 @@ Pattern::Pattern(cairo_surface_t *surface) {
 
 Pattern::~Pattern() {
   cairo_pattern_destroy(_pattern);
+  delete _repeat;
```
zbjornson: (continued) and then you can remove this line.

LinusU:
Very good comments from @zbjornson, also would be nice with a squash when/before merging, to avoid cluttering master with all the small commits (e.g. "whatever", "a test")

lib/context2d.js
```diff
@@ -66,9 +66,16 @@ Context2d.prototype.__defineGetter__('imageSmoothingEnabled', function(val){
  * @api public
  */
 
-Context2d.prototype.createPattern = function(image, repetition){
-  // TODO Use repetition (currently always 'repeat')
-  return new CanvasPattern(image);
+Context2d.prototype.createPattern = function(image, repeat){
```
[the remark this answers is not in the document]
asturur: sure. i did not notice i changed it.

asturur:
Squashing before eventually merging is not good enough?

LinusU:
That would work 👍 

Just wanted to comment as a reminder to myself or whomever it is that will merge :)

asturur:
dividing the requested changes in different commits, because i m not sure i m doing the right thing

asturur:
i think i addressed both the major and the minor comment.
Please check carefully, it renders good, unsure i m managing the pointer good sending the address with `&`. my understanding is that i have a static var now, and i m sending up its address pretending is a pointer. Is this fine?

asturur:
i also think adding a small UT for testing the new constructor is worth?

asturur:
ok as not said, you do not have tests for the C part of the library, no idea how to start them

zbjornson:
Cool, looks good with a quick glance, but will look more closely tomorrow.

I wouldn't worry about unit testing the C code.

(If I'm the one who merges I'll squash.)

zbjornson:
Ported to master/v2.x as 5c28c09.

Thanks for the contribution!

eNepper:
Will this also work with another canvas as the pattern instead of an image?

asturur:
i did not try it but it should. if it doesn't please let me know. i want to work on repeatX and repeatY when i m back from vacation and may look into that.

PR #1071, opened 2017-12-28
Update font example to mention Fontname

bmathews:
Inside of FontForge, the `Fontname` property is what should be referenced and then used as the `family` value for `registerfont`, not `Family Name`.

### DIFF
--- a/examples/font.js
+++ b/examples/font.js
@@ -9,7 +9,7 @@ function fontFile (name) {
 // Pass each font, including all of its individual variants if there are any, to
 // `registerFont`. When you set `ctx.font`, refer to the styles and the family
 // name as it is embedded in the TTF. If you aren't sure, open the font in
-// FontForge and visit Element -> Font Information and copy the Family Name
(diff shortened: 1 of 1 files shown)


chearon:
Do you have an example font for that? According to [FontForge](https://fontforge.github.io/fontinfo.html#TTF-Names) it sounds like that's not right: 

>the fontname (which is the family name with a bunch of modifiers like Bold Italic Condensed tacked on to the end.

But the modifiers get parsed in node-canvas at the beginning (e.g. `Bold  10px Arial`). It's possible putting it in the family works too (`10px Arial Bold`) if the font selection system (depends on your OS) specially parses the family name, I don't remember the details of that. I would be surprised if only the latter worked though.

bmathews:
Ah, it appears that the issue was actually because I wasn't wrapping the the family name in quotes when setting ctx.font. Switching to `fontname` allowed it to work because it had no spaces.

My apologies!

Issue #1075, opened 2018-01-15
node-canvas fails to install, build output shows type conversion error


logicbomb421:
When installing, the node-gyp build phase eventually fails with a `make failed with exit code 2` message. Looking through the build output (full output below), I noticed the following:

> ../src/CanvasRenderingContext2d.cc:525:65: error: invalid conversion from ‘int’ to ‘cairo_format_t {aka _cairo_format}’ [-fpermissive]
else if (!strcmp(*utf8PixelFormat, "RGB16_565")) format = CAIRO_FORMAT_RGB16_565;

My guess is this is causing make to fail. I have followed the installation instructions in the readme (specifically: `sudo yum install cairo cairo-devel cairomm-devel libjpeg-turbo-devel pango pango-devel pangomm pangomm-devel giflib-devel`), however this error still appears.

### Full Output
```

> bcrypt@1.0.3 install /var/project/web/node_modules/@project/core/node_modules/bcrypt
> node-pre-gyp install --fallback-to-build

[bcrypt] Success: "/var/project/web/node_modules/@project/core/node_modules/bcrypt/lib/binding/bcrypt_lib.node" is installed via remote

> canvas@2.0.0-alpha.9 install /var/project/web/node_modules/canvas
> node-gyp rebuild

make: Entering directory `/var/project/web/node_modules/canvas/build'
  SOLINK_MODULE(target) Release/obj.target/canvas-postbuild.node
  COPY Release/canvas-postbuild.node
  CXX(target) Release/obj.target/canvas/src/backend/Backend.o
  CXX(target) Release/obj.target/canvas/src/backend/ImageBackend.o
../src/backend/ImageBackend.cc: In member function ‘int32_t ImageBackend::approxBytesPerPixel()’:
../src/backend/ImageBackend.cc:28:3: warning: case value ‘4’ not in enumerated type ‘cairo_format_t {aka _cairo_format}’ [-Wswitch]
   case CAIRO_FORMAT_RGB16_565:
   ^
  CXX(target) Release/obj.target/canvas/src/backend/PdfBackend.o
In file included from ../src/backend/PdfBackend.cc:4:0:
/usr/include/libpng12/png.h:2657:31: warning: invalid suffix on literal; C++11 requires a space between literal and identifier [-Wliteral-suffix]
        fprintf(PNG_DEBUG_FILE,"%s"m PNG_STRING_NEWLINE,(num_tabs==1 ? "\t" : \
                               ^
[...]

PR #1087, opened 2018-02-02
Fix for centos 6.x

wsw0108:


wsw0108:
`CAIRO_FORMAT_RGB16_565` was added since cairo-v1.2.0, but mark as DEPRECATED from next version cairo-v1.2.2, until cairo-v1.10.0

wsw0108:
related to #1075

zbjornson:
Cairo has built-in version macros actually: https://cairographics.org/manual/cairo-Version-Information.html

```cpp
##if CAIRO_VERSION >= CAIRO_VERSION_ENCODE(1, 0, 0)
printf ("Compiling with suitable cairo version: %s\n", %CAIRO_VERSION_STRING);
##endif

if (cairo_version() >= CAIRO_VERSION_ENCODE(1, 0, 0))
    printf ("Running with suitable cairo version: %s\n", cairo_version_string ());
```

Want to use that instead?

wsw0108:
I tried use the version macros from `cairo.h`, but it cause error on ci.

src/CanvasRenderingContext2d.cc
```diff
@@ -635,7 +635,9 @@ NAN_METHOD(Context2d::New) {
         if (!strcmp(*utf8PixelFormat, "RGBA32")) format = CAIRO_FORMAT_ARGB32;
         else if (!strcmp(*utf8PixelFormat, "RGB24")) format = CAIRO_FORMAT_RGB24;
         else if (!strcmp(*utf8PixelFormat, "A8")) format = CAIRO_FORMAT_A8;
+#if CAIRO_VERSION >= CAIRO_VERSION_ENCODED(1, 10, 0)
```
zbjornson: The macro is `ENCODE`, not `ENCODED` -- hopefully that fixes it.

wsw0108:
I make this PR because some cloud host like aliyun still provide CentOS-6 as guest os.

LinusU:
A downside with just `ifdef`ing these out is that passing `RGB16_565` will create a canvas using `CAIRO_FORMAT_RGB24` without giving any indication to the user that something went wrong.

If we are going down this route, I think that we need to add `throw`s when the library is called with something that we cannot support 🤔

wsw0108:
I agree with you. 
I will add some check for this.

wsw0108:
For now I think there is no reason to support old CentOS6.x, so I will close this.

PR #1092, opened 2018-02-12
Support disabling chroma subsampling in JPEG output

thenickdude:
This gives significant quality improvements for images with fine colourful details (mostly seen on synthetic images, most photos won't benefit).

Test program:

```js
const fs = require('fs')

const canvas = createCanvas(230, 200)

ctx.fillStyle = 'white';
ctx.fillRect(0, 0, canvas.width, canvas.height)

loadImage('test.png')
	.then((image) => {
		ctx.drawImage(image, 0, 0)
	})
	.then(() => new Promise((resolve, reject) => {
		var
			out = fs.createWriteStream(__dirname + '/test-default.jpg'),
			stream = canvas.jpegStream({
				quality: 94 // Chosen to make both images about the same final filesize
			})
		
		out.on('finish', function () {
			resolve()
		});
		
		stream.pipe(out);
	}))
	.then(() => new Promise((resolve, reject) => {
		var
			out = fs.createWriteStream(__dirname + '/test-no-chroma-subsample.jpg'),
			stream = canvas.jpegStream({
				quality: 90,
				disableChromaSubsampling: true
			})
		
		out.on('finish', function () {
			resolve()
		});
		
		stream.pipe(out);
	}))
```

The test.png input file:

test

The output with the default node-canvas options (2x2 chroma subsampling, the current default on the master branch) (25kB):
test-default

The output with chroma subsampling disabled (25kB):
test-no-chroma-subsample

Here are those two images blown up and shown side-by-side, notice how the colourful details are much better preserved even with the same filesize (at the red tail feathers, for example):

comparison

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -177,6 +177,7 @@ var stream = canvas.jpegStream({
     bufsize: 4096 // output buffer size in bytes, default: 4096
   , quality: 75 // JPEG quality (0-100) default: 75
   , progressive: false // true for progressive compression, default: false
+  , disableChromaSubsampling: false // true to disable 2x2 subsampling of the chroma components, default: false
 });
 ```
 
--- a/lib/canvas.js
+++ b/lib/canvas.js
@@ -94,6 +94,8 @@ Canvas.prototype.createJPEGStream = function(options){
       bufsize: clampedBufSize
     , quality: options.quality || 75
     , progressive: options.progressive || false
+    , chromaHSampFactor: options.disableChromaSubsampling ? 1 : 2
+    , chromaVSampFactor: options.disableChromaSubsampling ? 1 : 2
   });
 };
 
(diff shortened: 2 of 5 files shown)


zbjornson:
Code looks good. Would it make more sense to expose the numerical option (default 2) instead of true/false? I've never played with this option before, don't know if >2 is ever useful.

thenickdude:
I think >2 might be useful to somebody (maybe for extreme thumbnail
compression), but libturbo quits with an error message if I try it and I
couldn't figure out how to get it working, lol.

zbjornson:
Hm, since this hasn't been released yet ... it might be nicer to have the option be `chromaSubsampling: true` instead of `disableChromaSubsampling: false`. What do you think? I can open a follow-up PR later today.

thenickdude:
Maybe call it chromaSubsampling with supported values false, true, 1 and 2? Support for numeric values higher than 2 could be added later if someone figures out why libturbo complains about them.

PR #1094, opened 2018-02-13
Added framebuffer device support

timstableford:
* Framebuffer device support.
* Optional double buffering.
* Examples of using framebuffer backend.


zbjornson:
Cool, but this will break support for Windows and I think MacOS... If this goes in, it'll need conditional compilation to maintain support.

timstableford:
Alright, any examples of conditional compilation I should base on? And any suggestions on detecting Linux are welcome although I'm sure I can find something.

zbjornson:
I'm not sure how to detect FB, but it's not as simple as detecting linux. A lot of BSDs, all or almost all headless kernels and many cloud images (including the stock ones from AWS, Azure and GCP) don't have it built in.

A better approach would probably be to link against node-canvas and do this as an add-on so we can keep node-canvas relatively "pure." Linking against node-canvas is trivial, but I haven't looked at how easy it would be to stick in a new backend. (I'd support reasonable changes to node-canvas that make pluggable external backends possible, since this issue keeps coming up once a year or so.)

timstableford:
For those scenarios as far as I know it won't break compilation, it just won't be able to open the frame-buffer device at runtime. 
I think it would be a useful built-in feature which is why I've made this PR but I'm happy to make it an add-on. From what I can tell as long as I can link against node-canvas and access the headers node-canvas won't require any changes.

piranna:
Here commiter of backends support.

At https://github.com/Automattic/node-canvas/pull/571 you have the original pull-request where I extracted the backends support, and as you can see there my main focus was support FbDEV too. Check FbDEV is easy since it's a Linux-only interface, so you only need to check for the OS. There's no extra libs since Cairo support it natively, and added library size is minimal. Later, to know if the system has support for FbDEV is just only done by checking if there's any `/dev/fbX` device, you don't need to worry about anything else.

Framebuffer support is available on BSD I think, but it's using other incompatible APIs. The same happens for Windows and OSX, that they would need their own backends too, that's why the backends support was done. An initial work for X11 support is on my pull-request, but it has some flaws regarding surface initialization and refresh.

A plugable API for backends is a nice idea, but being so few backends available (don't think there would be more that the ones already discussed) I think it's overkill and would be better to have them directly in node-canvas source code, enabled by platform detection and compile flags.

timstableford:
I tend to agree with @piranna 's sentiment on the limited number of backends. It's now unlikely that I'll develop it as an addon. I am however happy to support this PR and any changes requested.

timstableford:
Closing this now that there's basic framebuffer support in master

timstableford:
This is what I saw https://github.com/Automattic/node-canvas/wiki/Using-with-framebuffer-devices

I also half-heartedly updated this PR before closing it and I'm not sure how functional it is. I decided that connecting node-canvas to Xorg would work better for me so I can use the event handling. Check out my fork if that would be useful to you.

piranna:
Yes, that wiki page is what I said is not working for them. That's no framebuffer support at all.

timstableford:
Okay. I've checked this PR on a device now and it still works.

src/backend/FramebufferBackend.cc
```diff
@@ -0,0 +1,200 @@
...
+            (unsigned char *)this->fbp,
+            CAIRO_FORMAT_RGB16_565,
+            vinfo.xres, vinfo.yres,
+            cairo_format_stride_for_width(CAIRO_FORMAT_RGB16_565, vinfo.xres));
+
+    // Store the width and height to be used further up the stack.
+    this->width = vinfo.xres;
+    this->height = vinfo.yres;
+
+    // If double buffering is enabled then allocate an in memory buffer.
```
piranna: Instead of creating a buffer in memory and copy it, it's more performant to create a single buffer in VRAM with doubled height and just set the offset were it should be readed for each frame, or if that's not possible, at least copy copy the data between the two VRAM regions instead of from the main RAM.

src/backend/FramebufferBackend.cc
```diff
@@ -0,0 +1,200 @@
...
+}
+
+void FramebufferBackend::setHeight(int height)
+{
+    Nan::ThrowError("Framebuffer setHeight is not supported");
+}
+
+cairo_format_t FramebufferBackend::getFormat()
+{
+    return CAIRO_FORMAT_RGB16_565;
```
piranna: Why is `CAIRO_FORMAT_RGB16_565` format hardcoded? Receng graphic cards use RGBA32 by default... Currently RGB16 is only being used by default QEmu Cirrus VGA graphic card, and it's heavily recomended to use the SVGA one that between other things don't have this limitation.

src/backend/FramebufferBackend.cc
timstableford: That is an interesting idea but I'm not sure how to tell a framebuffer to change where it's looking

src/backend/FramebufferBackend.cc
timstableford: Tasty Pi's :) You're right though, this should really be a parameter

src/backend/FramebufferBackend.cc
piranna: In the framebuffer docs there's a reference to how to set an offset where it should start reading.

src/backend/FramebufferBackend.cc
piranna: If you take a look on my pull-request, I have support for several color schemes there :-)

piranna:
Why got this clossed?

timstableford:
I think it's now possible to write to a canvas in the core library I think? You're welcome to fork/update and re-open if you like

piranna:
Can you point to a reference of the support in the core library?

timstableford:
https://github.com/Automattic/node-canvas/wiki/Using-with-framebuffer-devices

piranna:
> Automattic/node-canvas/wiki/Using-with-framebuffer-devices

That's not a solution at all. What's being done there is writing the canvas in memory and later copy its content by hand to the framebuffer, totally a performance and resources waste. To be able to have high FPS is needed to draw directly on the framebuffer device or using DRM framework.

Issue #1099, opened 2018-02-20
Assertion failed


6-8-axnw1bom81v5xa3nh48c:
![1](https://user-images.githubusercontent.com/35625521/36417186-870d3092-162b-11e8-961e-6562e6cd45f3.png)

## Steps to reproduce

```js
var {Canvas} = require('canvas');
var canvas = new Canvas(a => a);
```

## Environment
* **Version of node-canvas**: master
* **Environment**: node master, os windows 8.1

PR #1100, opened 2018-02-20
Perform some checks to avoid assertion fault

Hakerh400:
Fixes https://github.com/Automattic/node-canvas/issues/1099

### DIFF
--- a/src/Canvas.cc
+++ b/src/Canvas.cc
@@ -104,11 +104,13 @@ NAN_METHOD(Canvas::New) {
       backend = new ImageBackend(width, height);
   }
   else if (info[0]->IsObject()) {
-    v8::Local<v8::Object> arg = info[0]->ToObject();
-    if(arg.IsEmpty() || arg->InternalFieldCount() == 0){
+    if (Nan::New(ImageBackend::constructor)->HasInstance(info[0]) ||
+        Nan::New(PdfBackend::constructor)->HasInstance(info[0]) ||
+        Nan::New(SvgBackend::constructor)->HasInstance(info[0])) {
+      backend = Nan::ObjectWrap::Unwrap<Backend>(info[0]->ToObject());
+    }else{
       return Nan::ThrowTypeError("Invalid arguments");
     }
-    backend = Nan::ObjectWrap::Unwrap<Backend>(arg);
   }
   else {
     backend = new ImageBackend(0, 0);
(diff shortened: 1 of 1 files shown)


src/Canvas.cc
```diff
@@ -104,8 +104,11 @@ NAN_METHOD(Canvas::New) {
       backend = new ImageBackend(width, height);
   }
   else if (info[0]->IsObject()) {
-    // TODO need to check if this is actually an instance of a Backend to avoid a fault
-    backend = Nan::ObjectWrap::Unwrap<Backend>(info[0]->ToObject());
+    v8::Local<v8::Object> arg = info[0]->ToObject();
+    if(arg.IsEmpty() || arg->InternalFieldCount() == 0){
+      return Nan::ThrowTypeError("Invalid arguments");
+    }
```
zbjornson: What you actually want to check here is that the object is an instance of a backend. Something like:

```cpp
(Nan::New(Backend::constructor)->HasInstance(obj))
```

(However I forget how inheritance in backends affects this.)

Hakerh400: @zbjornson I admit, this was not the perfect solution, but at least it suppresses assertion fault error.

Anyway, I'll add some commits to actually check for the instance, but I'm not sure what behavior should be implemented if a backed instance is passed as the first argument, since `Backend` itself is an abstract class.

My idea now is to implement static method `Nan::Persistent<FunctionTemplate> constructor` in `Backend` class (since it currently doesn't have one) and then in `Backend::New` construct `ImageBackend`, `PdfBackend` or `SvgBackend` based on the argument type. Or you have something other in mind?

> However I forget how inheritance in backends affects this.

I'm not sure either, but current implementation should probably be changed. `Canvas` constructor doesn't check if the passed object is a valid backend, it just wraps it using `nan` module which may result in an error or segfault.

zbjornson: Ah right, probably easier to just check each of the implementing classes explicitly. There aren't that many possible Cairo backends, so I'm not worried about this growing out of control.

```cpp
if (Nan::New(ImageBackend::constructor)->HasInstance(info[0]) ||
    Nan::New(PdfBackend::constructor)->HasInstance(info[0]) ||
    Nan::New(SvgBackend::constructor)->HasInstance(info[0])) {
    backend = Nan::ObjectWrap::Unwrap<Backend>(info[0]->ToObject());
}
```

I don't think it's worth the effort of doing it more nicely than that (adding a `Backend::constructor`)...

> Canvas constructor doesn't check if the passed object is a valid backend

Yeah, this slipped through when backend support was added. Thanks for addressing the TODO.

Hakerh400:
Updated, as you suggested. Please recheck.

Issue #1093, opened 2018-02-13
registerFont Memory usage...


smaznet:
Hi i tested registerFont and ... this is result of

![image](https://user-images.githubusercontent.com/10239885/36135811-a1215794-10a2-11e8-801e-d690d47449fe.png)

``` js
const Canvas = require('canvas');


for(let i = 0 ; i < 10000 ; i++){
	Canvas.registerFont("../data/fonts/font.ttf",{family: "font"})
}
```

chearon:
Well shoot that's not good. There is a lot of memory (de)allocations that happen there so it's not hard to believe. I'll take a look when I get a chance.

PR #1103, opened 2018-03-02
Deallocate library structure at get_pango_font_description

Hakerh400:
Should fix https://github.com/Automattic/node-canvas/issues/1093


src/register_font.cc
```diff
@@ -212,6 +212,7 @@ get_pango_font_description(unsigned char* filepath) {
       pango_font_description_set_style(desc, get_pango_style(face->style_flags));
 
       FT_Done_Face(face);
+      FT_Done_FreeType(library);
```
chearon: Thanks! Looks like I also missed a `pango_font_description_free(desc);` here though too

Issue #1091, opened 2018-02-11
Crash with font file (attached)


Darker:
No errors available - I'd probably need to debug the native part of the library. It just stops immediatelly.

How I reister the font:

```
Canvas.registerFont('assets/font_lightd.ttf', { family: 'MyFont' });
```

How I use it (crashes on this line):

```
ctx.font = '12px "MyFont"';
```

Font is attached.
[error_font.zip](https://github.com/Automattic/node-canvas/files/1714285/error_font.zip)

PR #1105, opened 2018-03-03
Prevent segfault caused by invalid fonts

Hakerh400:
Fix https://github.com/Automattic/node-canvas/issues/1091

Pango is unable to parse the font attached in the issue, so probably the best we can do is to allow pango to properly display the message and use fallback font.

### DIFF
--- a/src/Canvas.cc
+++ b/src/Canvas.cc
@@ -758,6 +758,11 @@ Canvas::ResolveFontDescription(const PangoFontDescription *desc) {
     for (; it != _font_face_list.end(); ++it) {
       if (g_ascii_strcasecmp(families[i], pango_font_description_get_family(it->user_desc)) == 0) {
         char *name = g_strdup(pango_font_description_get_family(it->sys_desc));
+
+        if (!name) {
+          name = g_strdup("");
+        }
+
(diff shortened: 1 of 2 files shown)


src/Canvas.cc
```diff
@@ -758,6 +758,11 @@ Canvas::ResolveFontDescription(const PangoFontDescription *desc) {
     for (; it != _font_face_list.end(); ++it) {
       if (g_ascii_strcasecmp(families[i], pango_font_description_get_family(it->user_desc)) == 0) {
         char *name = g_strdup(pango_font_description_get_family(it->sys_desc));
+
+        if (!name) {
```
chearon: Thanks again for another bug fix.

I think it can be caught more early on during registration, when the `sys_desc` is built. In [`get_pango_font_description`](https://github.com/Automattic/node-canvas/blob/master/src/register_font.cc#L199) I think it should be bailing early [here](https://github.com/Automattic/node-canvas/blob/master/src/register_font.cc#L209) (plus memory cleanup), otherwise there's a null font description present which would be nice to avoid altogether.

That'd also throw the `"Could not parse font file"` error which people could handle.

Hakerh400: I've attached another commit, is it better now?

PR #1114, opened 2018-03-14
Update changelog

zbjornson:
Not sure I got absolutely everything, but got most of it, and I think I got all breaking changes.

Second commit renames the file from "history.md" to "CHANGELOG.md" -- hopefully that makes it easier to find.

---

2.x is **big** compared to the previous releases :).


LinusU:
Awesome work 🙌

PR #1115, opened 2018-03-15
Propagate async context

kkoopa:
Thanks for contributing!

- [N] Have you updated CHANGELOG.md?

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
   },
   "dependencies": {
     "node-pre-gyp": "^0.9.0",
-    "nan": "^2.4.0"
+    "nan": "^2.9.2"
   },
   "devDependencies": {
     "assert-rejects": "^0.1.1",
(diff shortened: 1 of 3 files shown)


zbjornson:
@kkoopa Your original commit got rid of (or came before?) the deprecation warnings for `Nan::MakeCallback`, but since updating this branch with master, the warnings are back. I also tried this branch with nan 2.4.0, but [the warnings remain](https://travis-ci.org/zbjornson/node-canvas/jobs/356604738#L637). Is this PR supposed to fix them?

kkoopa:
That is a (slightly) different set of deprecation warnings. However, they would fall within the scope of this PR, since they also are related to async context propagation. I will add a commit when I have time.

kkoopa:
That should do it. Not the most elegant solution, but I am not rewriting that closure callback mess.

zbjornson:
Sweet, thanks @kkoopa!

PR #1119, opened 2018-03-17
Update node-pre-gyp to 0.9.0

dragonfire535:
Simple update of the `node-pre-gyp` dependency, so that canvas can be without the `request` library and begin the dependency count cut-down.


LinusU:
Thanks

Issue #1120, opened 2018-03-18
Rendering context becomes unusable if rotation angle is set to NaN


jsdeveloper97:
Example:

```js
ctx.rotate(NaN);
ctx.resetTransform();
ctx.fillStyle = 'red';
ctx.fillRect(0, 0, 100, 100);
```

* **In browsers:** properly resets the transform and draws the square
* **In node-canvas:** `ctx` becomes unusable, it doesn't draw anything no matter what I try

## Your Environment
* Version of node-canvas (e.g. 1.4.0): 2.0.0
* Environment: node 9.7 windows 10

PR #1121, opened 2018-03-19
Return if the rotation angle is NaN or infinite

Hakerh400:
Fix: https://github.com/Automattic/node-canvas/issues/1120
Spec ref: [dom-context-2d-rotate](https://html.spec.whatwg.org/multipage/canvas.html#dom-context-2d-rotate)

### DIFF
--- a/src/CanvasRenderingContext2d.cc
+++ b/src/CanvasRenderingContext2d.cc
@@ -1890,8 +1890,7 @@ NAN_METHOD(Context2d::Rotate) {
     return;
 
   Context2d *context = Nan::ObjectWrap::Unwrap<Context2d>(info.This());
-  cairo_rotate(context->context()
-    , info[0]->IsNumber() ? info[0]->NumberValue() : 0);
+  cairo_rotate(context->context(), angle);
 }
 
 /*
--- a/src/CanvasRenderingContext2d.cc
+++ b/src/CanvasRenderingContext2d.cc
@@ -1884,6 +1884,11 @@ NAN_METHOD(Context2d::ClosePath) {
  */
 
 NAN_METHOD(Context2d::Rotate) {
+  double angle = info[0]->NumberValue();
+
+  if (isnan(angle) || isinf(angle))
+    return;
+
   Context2d *context = Nan::ObjectWrap::Unwrap<Context2d>(info.This());
   cairo_rotate(context->context()
     , info[0]->IsNumber() ? info[0]->NumberValue() : 0);
--- a/test/canvas.test.js
+++ b/test/canvas.test.js
@@ -1498,7 +1498,7 @@ describe('Canvas', function () {
 
     // Number
     ctx.resetTransform();
-    testAngle(1.23);
+    testAngle(1.23, 1.23);
 
     // String
     ctx.resetTransform();
@@ -1513,7 +1513,11 @@ describe('Canvas', function () {
     testAngle([7.8], 7.8);
 
     // Object
-    var obj = {[Symbol.toPrimitive](){ return 0.89; }};
+    var obj = Object.create(null);
+    if (+process.version.match(/\d+/) >= 6)
+      obj[Symbol.toPrimitive] = function () { return 0.89; };
+    else
+      obj.valueOf = function () { return 0.89; };
     ctx.resetTransform();
     testAngle(obj, 0.89);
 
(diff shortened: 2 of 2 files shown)


src/CanvasRenderingContext2d.cc
```diff
@@ -1884,6 +1884,11 @@ NAN_METHOD(Context2d::ClosePath) {
...
 
 NAN_METHOD(Context2d::Rotate) {
+  double angle = info[0]->NumberValue();
+
+  if (isnan(angle) || isinf(angle))
+    return;
+
   Context2d *context = Nan::ObjectWrap::Unwrap<Context2d>(info.This());
   cairo_rotate(context->context()
     , info[0]->IsNumber() ? info[0]->NumberValue() : 0);
```
LinusU: This line should probably be changed as well. 

btw, in the browser, if you set it to NaN, does it set it to 0 or just don't change it?

Hakerh400: @LinusU 

> This line should probably be changed as well.

Sorry, which line?

> In the browser, if you set it to NaN, does it set it to 0 or just don't change it?

Well, the easiest way to find it out is to check manually. All modern browsers (which correctly follow specification for HTML5 canvas) simply ignore NaNs as rotation angles. They don't set the angle to 0, just return from the function without taking any actions. I posted the link to the part from the specification which defines behavior in case of NaN as the argument.

LinusU: Sorry, missed the specification link, that's perfect!

> Sorry, which line?

```c++
cairo_rotate(context->context()
     , info[0]->IsNumber() ? info[0]->NumberValue() : 0);
```

could be updated to just:

```c++
cairo_rotate(context->context(), angle);
```

Also, it would be nice to double check what `info[0]->NumberValue()` returns if `info[0]-IsNumber() == false`. I would guess that it returns `NaN`, but if you could add a test case that tries setting the rotation to a string, boolean, array, etc. and see that it's being properly ignored, that would be good 👍

Hakerh400: @LinusU 
Right, that check was completely redundant, I've just removed it.

> I would guess that it returns `NaN`

What actually happens is that javascript engine calls  `Symbol.toPrimitive` of it's prototype and converts the returned value to a number. For strings it would try to parse it as a double, for arrays it will stringify the array and again convert it to a double, etc.

> If you could add a test case that tries setting the rotation to a string, boolean, array, etc. and see that it's being properly ignored, that would be good

Ok, I'll add a test case, but not all non-number values should be ignored, as I explained, only if the return value of `Symbol.toPrimitive` is NaN or infinite.

LinusU: Ah, cool, yeah it seems like the browser accepts e.g. `'20'`. Good to know :)

Perfect then! Thank you for your work on this!

test/canvas.test.js
```diff
@@ -1492,4 +1492,53 @@ describe('Canvas', function () {
...
+    ctx.resetTransform();
+    ctx.rotate(0.91);
+    testAngle(NaN, 0.91);
+
+    // Infinite value
+    ctx.resetTransform();
+    ctx.rotate(0.94);
+    testAngle(-Infinity, 0.94);
+
+    function testAngle(angle, expected = angle){
```
LinusU: Default arguments are not supported on Node.js 4 which we currently are supporting

Hakerh400: @LinusU . Already fixed.

LinusU:
Great! Thanks for the PR, I'll give some time for @zbjornson or @chearon to look it over as well, but for me we are good to go 👍

Hakerh400:
@zbjornson 

> Probably something to do systematically for all methods, sometime...

I'm preparing a solution for it too, but I think I'll open another PR for it soon.

LinusU:
Thanks!

LinusU:
I think that this maybe introduced the following error for me:

```text
  1) Canvas Context2d#rotate(angle):

      AssertionError [ERR_ASSERTION]: 0.7895037396899504 === 0.7895037396899505
      + expected - actual

      -0.7895037396899504
      +0.7895037396899505
      
      at testAngle (test/canvas.test.js:1542:14)
      at Context.<anonymous> (test/canvas.test.js:1527:5)
```

@Hakerh400 any ideas?

Hakerh400:
@LinusU 
That is probably a floating point rounding error. I think `assert.ok(Math.abs(mat.m12 - sin) < 1e-15);` would solve the problem.

LinusU:
Yeah, comparing against `< Number.EPSILON` would probably be even better.

PRs welcome ☺️

Hakerh400:
@LinusU 
I've fixed it in #1129.

PR #1129, opened 2018-03-22
Systematically check for NaNs in ctx methods

Hakerh400:
Fixed NaN checking in all relevant `CanvasRenderingContext2D` methods.  
I don't think it is worth adding a test for each method, but if it is, I'll add them.

### DIFF
--- a/src/CanvasRenderingContext2d.cc
+++ b/src/CanvasRenderingContext2d.cc
@@ -2647,8 +2647,11 @@ bool checkArgs(const Nan::FunctionCallbackInfo<Value> &info, double *args, int a
   for (int i = offset; i < argsEnd; i++) {
     double val = info[i]->NumberValue();
 
-    if (areArgsValid){
-      if (isnan(val) || isinf(val)){
+    if (areArgsValid) {
+      if (isnan(val) || isinf(val)) {
+        // We should continue the loop instead of returning immediately
+        // See https://html.spec.whatwg.org/multipage/canvas.html
+
         areArgsValid = false;
         continue;
       }
--- a/src/CanvasRenderingContext2d.cc
+++ b/src/CanvasRenderingContext2d.cc
@@ -2660,5 +2660,5 @@ bool checkArgs(const Nan::FunctionCallbackInfo<Value> &info, double *args, int a
     }
   }
 
-  return true;
+  return areArgsValid;
 }
--- a/src/CanvasRenderingContext2d.cc
+++ b/src/CanvasRenderingContext2d.cc
@@ -40,14 +40,13 @@ Nan::Persistent<FunctionTemplate> Context2d::constructor;
  */
 
 #define RECT_ARGS \
-  if (!info[0]->IsNumber() \
-    ||!info[1]->IsNumber() \
-    ||!info[2]->IsNumber() \
-    ||!info[3]->IsNumber()) return; \
-  double x = info[0]->NumberValue(); \
-  double y = info[1]->NumberValue(); \
-  double width = info[2]->NumberValue(); \
-  double height = info[3]->NumberValue();
+  double args[4]; \
+  if(!checkArgs(info, args, 4)) \
+    return; \
+  double x = args[0]; \
+  double y = args[1]; \
+  double width = args[2]; \
+  double height = args[3];
 
 /*
  * Text baselines.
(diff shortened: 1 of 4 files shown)


src/CanvasRenderingContext2d.cc
```diff
@@ -2643,3 +2639,26 @@ NAN_METHOD(Context2d::Ellipse) {
...
+
+        areArgsValid = false;
+        continue;
+      }
+
+      args[i - offset] = val;
+    }
+  }
+
+  return true;
```
LinusU: `return areArgsValid`?

src/CanvasRenderingContext2d.cc
```diff
@@ -2643,3 +2639,26 @@ NAN_METHOD(Context2d::Ellipse) {
...
+bool checkArgs(const Nan::FunctionCallbackInfo<Value> &info, double *args, int argsNum, int offset){
+  int argsEnd = offset + argsNum;
+  bool areArgsValid = true;
+
+  for (int i = offset; i < argsEnd; i++) {
+    double val = info[i]->NumberValue();
+
+    if (areArgsValid) {
+      if (isnan(val) || isinf(val)) {
+        // We should continue the loop instead of returning immediately
```
LinusU: I don't see how it could affect the outcome to continue loop?

Hakerh400:
@LinusU 

 > I don't see how it could affect the outcome to continue loop?

Here is an example:

```js
var count = 0;
var obj = Object.create(null);

obj[Symbol.toPrimitive] = () => {
  count++;
  return NaN;
};

ctx.moveTo(obj, obj);
console.log(count);
```

If we continue the loop, it will log `2`, if we return it will log `1`, but according to the spec it should log `2`.

> return areArgsValid?

Yap, that was a typo, let me fix it.

LinusU:
Haha, that's amazing, good catch!

src/CanvasRenderingContext2d.cc
```diff
@@ -2643,3 +2639,26 @@ NAN_METHOD(Context2d::Ellipse) {
...
+
+bool checkArgs(const Nan::FunctionCallbackInfo<Value> &info, double *args, int argsNum, int offset){
+  int argsEnd = offset + argsNum;
+  bool areArgsValid = true;
+
+  for (int i = offset; i < argsEnd; i++) {
+    double val = info[i]->NumberValue();
+
+    if (areArgsValid) {
+      if (isnan(val) || isinf(val)) {
```
zbjornson: (continued from previous comment) ...

GCC and clang are marginally better with `isinf`, but MSVC is significiantly better with `val == std::numeric_limits<double>::infinity() || val == -std::numeric_limits<double>::infinity()`. MSVC doesn't inline `isinf` in this case, and has a different implementation:

```
(inline)
00007FFBE4F48DF5  ucomisd     xmm6,xmm7  
00007FFBE4F48DF9  jp          checkArgs+8Dh (07FFBE4F48DFDh)  
00007FFBE4F48DFB  je          checkArgs+93h (07FFBE4F48E03h)  

vs

(call)
00007FFBE4F5C854  mov         rcx,7FFFFFFFFFFFFFFFh  
00007FFBE4F5C85E  movsd       mmword ptr [rsp+8],xmm0  
00007FFBE4F5C864  mov         rax,qword ptr [_bitsy]  
00007FFBE4F5C869  and         rax,rcx  
00007FFBE4F5C86C  mov         rcx,7FF0000000000000h  
00007FFBE4F5C876  sub         rax,rcx  
00007FFBE4F5C879  shr         rax,3Fh  
00007FFBE4F5C87D  ret  
```

GCC and Clang: https://godbolt.org/g/GxxiqG


```
MSVC benchmarks:
  (ops/sec)      master     orig PR  ==infinity
lineTo()     22,447,439  20,242,779  21,345,058
arc()         1,883,895   1,747,626   1,872,457
strokeRect() 13,206,246  12,629,641  13,189,635
```

I guess leave it as-is since more servers run Linux and Windows.

/ramble

zbjornson:
Also, can you add a line to the changelog please? :)

Hakerh400:
I've attached some fixes and added relevant info to the changelog.  
Seems like travis is failing at `Context2d#measureText().width` due to timeout limit. Not sure why, I tested it locally on node8 and seems to work...

Hakerh400:
@zbjornson 
Is there anything else to be fixed?

PR #1149, opened 2018-04-20
Add ARM support to SYSTEM_PATHS

techborn:
On a Raspi - Kept getting issues with loading a JPEG image from disk. After some digging, I found out the has_lib util kept returning false for jpeg lib (probably for other libs too). Added these lines and re-configured and rebuilt the module. It works fine after that. 

System: Raspi 3 Model B (ARMv7)

### DIFF
--- a/util/has_lib.js
+++ b/util/has_lib.js
@@ -9,7 +9,9 @@ var SYSTEM_PATHS = [
   '/usr/local/lib',
   '/opt/local/lib',
   '/usr/lib/x86_64-linux-gnu',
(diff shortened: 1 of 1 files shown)


chearon:
@LinusU I'm interested in that too, do you think it would help with library path detection though?

techborn:
It does have library detection ala ldconfig, but the issue is that check if the system HAS ldconfig in the first place, it requires the use of the hash command/library, which apparently wasn't (isn't?) on my Raspi (debian stretch).

PR #1151, opened 2018-04-21
Prevent segfaults caused by creating a too large canvas

Hakerh400:
Prevent segfault caused by this:

```js
new Canvas(1e9, 1e9).getContext('2d').getImageData(0, 0, 1, 1);
```

- [X] Added test
- [X] Updated the changelog


src/backend/ImageBackend.cc
```diff
@@ -37,9 +37,17 @@ int32_t ImageBackend::approxBytesPerPixel() {
...
-	assert(!this->surface);
-	this->surface = cairo_image_surface_create(this->format, width, height);
-	assert(this->surface);
+	assert(!surface);
+	surface = cairo_image_surface_create(format, width, height);
+
+  if (!surface)
+    return NULL;
+
+  if (width && height && !cairo_image_surface_get_data(surface)) {
```
zbjornson: Is this `cairo_image_surface_get_data` check necessary? The docs say it returns:

> a pointer to the image data of this surface or NULL if surface is not an image surface, or if cairo_surface_finish() has been called.

which are not conditions that I think we need to worry about here.

It sounds like checking `surface` as you've done above is sufficient:

> `cairo_image_surface_create` always returns a valid pointer, but it will return a pointer to a "nil" surface if an error such as out of memory occurs

zbjornson:
Wonder if it's also worth checking for the max canvas size in the constructor. That would let us return a nicer error...

```c
 /* Limit on the width / height of an image surface in pixels.  This is
 * mainly determined by coordinates of things sent to pixman at the
 * moment being in 16.16 format. */
 #define MAX_IMAGE_SIZE 32767
```

https://github.molgen.mpg.de/git-mirror/cairo/blob/747cab741cf63f30aa6bc2b787fc13e7e7a2b859/src/cairo-image-surface.c#L58-L61

Hakerh400:
@zbjornson 

> Is this `cairo_image_surface_get_data` check necessary? It sounds like checking `surface` as you've done above is sufficient

The segfault is caused by image data being null. `cairo_image_surface_create` returns a valid pointer even if the size of the canvas is `1e9 x 1e9`. Removing that check causes segfault.

> Wonder if it's also worth checking for the max canvas size in the constructor.

It seems that `MAX_IMAGE_SIZE` is not reliable. I tested on different machines and seems that `23169` is actually the maximum valid image size. Everything above it breaks image data. But double checking in the constructor sounds like a good idea, I'll add a commit for it.

zbjornson:
Ah, I misread the docs -- I read "'nil' surface" as "nil (null) pointer".

I think what you actually want is to check `cairo_surface_status(surface)`. For me that's returning `CAIRO_STATUS_INVALID_SIZE` for >32767 px. Checking that and then returning a JS Exception with `cairo_status_to_string` would probably eliminate the need to check w/h > 32,767 in the constructor.

---

Re: `MAX_IMAGE_SIZE` hm, I've been able to get it that large (up to the limits of my available RAM):

```js
> c = new Canvas.Canvas(32767, 5), ctx = c.getContext("2d"), ctx.getImageData(0, 0, 32767, 5)
ImageData {
  data:
   Uint8ClampedArray [
     0, [snip]
```

(32768 ends the process.)

Hakerh400:
I'll fix it as soon as I get home (probably in a few days).

Hakerh400:
@zbjornson 

Sorry for waiting. I've attached some fixes. The problem was not only caused by creating a too large canvas, but also by obtaining image data of zero-width or zero-height canvas. I didn't find a good way to handle that case, but since there is a very little posibility that someone will require image data of zero-width canvas, I made it to throw an error in such cases.

Is it acceptable, or should we strictly follow the standard?

---

Hm.. I've just noticed that for some reason tabs are converted to spaces in several c++ files. Not sure why it happened. If it is a problem, I'll try to rollback it.

src/backend/Backend.cc
```diff
@@ -2,88 +2,105 @@
...
 
-	o << "operation " << this->operation_name;
-	o << " not supported by backend " + backend->getName();
+  o << "operation " << this->operation_name;
+  o << " not supported by backend " + backend->getName();
 
-	return o.str().c_str();
-};
+  return o.str().c_str();
+};
```
zbjornson: Can you add the newline back please?

Hakerh400:
@zbjornson 

Added the new line back.

Issue #1146, opened 2018-04-18
Missing sync canvas.toDataURL("image/jpeg") API


challakoushik:
I am using jsdom and node-canvas to run frontend html canvas code within node js, one of my script files has the line
`canvas.toDataURL('image/jpeg',{quality:1},function(){})`

the third parameter is because node-canvas demands 3 parameters over here at line 301 in canvas.js

```
else if ('image/jpeg' === type) {
    if (undefined === fn) {
      throw new Error('Missing required callback function for format "image/jpeg"');
    }

    var stream = this.jpegStream(opts);
    // note that jpegStream is synchronous
    var buffers = [];
    stream.on('data', function (chunk) {
      buffers.push(chunk);
    });
    stream.on('error', function (err) {
      fn(err);
    });
    stream.on('end', function() {
      var result = 'data:image/jpeg;base64,' + Buffer.concat(buffers).toString('base64');
      fn(null, result);
    });
  }
```
but jsdom passes only two arguments to this, and it might not be their fault since the official frontend canvas.toDataURL is supposed to take only 2 arguments. 
I need to generate pdfs using jpg images generated by canvas in node. Would appreciate any help

zbjornson:
You're right, `"image/jpeg"` currently only has async support, and the async API is non-standard.

Until that's fixed, you could use `canvas.jpegStream()` to create your JPEG buffers (in the same way that's done in the code that you pasted).

challakoushik:
yes, but again the canvas on which I'm using toDataURL() is an htmlcanvas element, which goes through jsdom and then goes to node canvas, and jsdom does not have jpegStream()

so is there a way i can create a node-canvas object using a html canvas element. If so, that would solve the issue

zbjornson:
Sorry -- I meant you could do that by accessing the node-canvas Canvas instance that jsdom wraps, which I think is just `canvas._getCanvas()`.

They actually provide their own implementation of `toBlob` that supports `image/jpeg` that you might be able to use: https://github.com/jsdom/jsdom/blob/2e90c57c282eb06023fab09304e29622268b91cb/lib/jsdom/living/nodes/HTMLCanvasElement-impl.js#L62-L86

challakoushik:
[...]

Issue #716, opened 2016-02-15
Output high resolution PNG


domoritz:
Hi, I'd like to output PNG images with a higher resolution.

I tried to scale the context but unfortunately that does not change the resolution of the generated PNG image. What do I need to change?

```js
var ctx = canvas.getContext('2d');
ctx.scale(2, 2);

var stream = canvas.createPNGStream();
stream.on('data', function(chunk) { out.write(chunk); });
```

LinusU:
You need to run `.scale(2, 2)` before you draw anything, it will only affect later calls. You should also double both the width and the height of your canvas...

Lord-Protector:
I did that use say and there is no any result, I just got bigger resolution with my old image in the left-top corner.

Issue #743, opened 2016-03-25
Canvas 2.0 🚀


LinusU:
I would like to start bringing up the discussion on the next major bump. This is a quick list on the top of my head on what I personally would like to get in.
- Remove sync streams
- Drop support for Node 0.8 (#740)
- Require pango (#715)

Would love to get some comments from both maintainers and users of the module :raised_hands:

jakeg:
Sounds great! Especially pango.

Does anyone with necessary know-how know if getting svg import (not export!) #494 . Would be possible? Happy to sponsor someone's time to work on this. In fact generally is it possible to donate some money for your and other developers' time?

zbjornson:
Awesome! I'd add:

* https://github.com/Automattic/node-canvas/pull/458 if it doesn't get merged sooner than 2.0

Re: remove sync streams, any use in adding a sync API? (e.g. `let pngBuffer = canvas.toPng();`)

LinusU:
That's right, I forgot about that one :+1:

I think we already have sync api? `canvas.toBuffer('png')`

jakeg:
Would love to see CMYK JPG import support too #425, or at least an exception thrown if trying to add one to a canvas.

piranna:
I would like to have static build so there's no dependencies on libraries installed on the system, this will fix a lot of install issues. Also I would like to have configurable backend support, so node-canvas could be used to draw on the Linux framebuffer and other on-screen outputs. I have support for both of them on my [pull-request](https://github.com/Automattic/node-canvas/pull/571).

chearon:
> I would like to have static build so there's no dependencies on libraries installed on the system, this will fix a lot of install issues 

[node-pre-gyp](https://github.com/mapbox/node-pre-gyp) seems to be the new way do do that. Canvas could move to that and there wouldn't have to be compilation on the client. I think people would still have to have Cairo and stuff though and of course Windows is still a nightmare.

Downsides are that it adds another step for publishing, and it relies on S3 which has to be paid for

piranna:
node-pre-gyp is a partial fix, my pull-request fix it for all platforms
(including OSX and Windows) by allowing to use statically linked libraries,
no external dependencies on Cairo, and also makes prebuilds more portable.
On the other hand, prebuild-install is a simpler and more flexible
alternative to node-pre-gyp, and by default uses GitHub releases for
distribution, so there's no publishing problem here, too.

piranna:
[...]

Issue #306, opened 2013-07-26
toBuffer optimization


kangax:
I ran into performance issues with `toBuffer`. Here's an example of drawing (rather) large image onto canvas, then calling `toBuffer`. `toBuffer` is what's taking the longest here (`drawImage`, for example, is insignificant). I'm getting ~7-7.5 sec on `toBuffer` call. This is on node v0.10.13, node-canvas v1.0.3, and Mac OS X (10.7.5).

Is there a way to speed things up? Anything at all? I also tried async `toBuffer` but it doesn't really help.

``` js
var fs = require('fs'),
    Canvas = require('canvas'),
    Image = Canvas.Image;

// 3.9MB, 4686x4668
fs.readFile(__dirname + '/big_img.jpg', function(err, bigImg) {
  if (err) throw err;

  img = new Image;
  img.src = bigImg;

  var canvas = new Canvas(img.width, img.height);
  var ctx = canvas.getContext('2d');

  ctx.drawImage(img, 0, 0);

  console.time('toBuffer');
  canvas.toBuffer();
  console.timeEnd('toBuffer'); // 7196ms, 7435ms, 7237ms
});
```

Attaching image used in a test.

![big_img](https://f.cloud.github.com/assets/383/862630/b4f4d900-f5f7-11e2-888e-2c2289c6edd0.jpg)

benzmuircroft:
https://github.com/Automattic/node-canvas/issues/562 webp is 13% in size of a png and 87% space saved! If webp support is not provided by the uses choice in browser then they will have png fall back.

Webp is supported on Chrome so users of chrome should be advantaged with lower size images and a faster browsing experience.

adamhooper:
Isn't the obvious solution to make this library _not_ encode image data?

In my dream world, `toBuffer()` returns a Buffer of native-endian ARGB32 pixel data: no encoding whatsoever. That would give users their choice of encoding libraries, at the cost of a single `memcpy()`.

Side-benefits: node-canvas would be easier to maintain and wouldn't depend on libpng or libjpeg.

Obviously, that's a breaking change. During transition, `toBuffer('argb32')` would be backward-compatible....

zbjornson:
@adamhooper `getImageData` returns a `Uint8ClampedArray` of RGBA values, which sounds like what your asking for, except for the color depth?

adamhooper:
@zbjornson Ooh, didn't see that method.

[...]

PR #1152, opened 2018-04-21
Support toBuffer("image/jpeg") and unify export configs

### LINKED ISSUE #982: Support toBuffer(cb, mimeType, quality)?
Some discussion in https://github.com/Automattic/node-canvas/issues/743, but in essence there should be a non-streaming API for all formats with all config options (e.g. quality, PNG filter).

`toBuffer(cb: function, mimeType?: string, quality?: number)` would parallel [`toBlob`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/toBlob).

```typescript
toBuffer(cb, "image/jpeg", quality: number);
toBuffer(cb, "image/png", compressionLevel: number);
// accommodate non-standard options:
toBuffer(cb, "image/png", {compressionLevel?: number, filter?: number, palette?: Uint8ClampedArray, backgroundIndex?: number});
toBuffer(cb, "image/svg"); // throws if not an SVG canvas
toBuffer(cb, "raw"); // Maybe? No need for this to use a callback as it's just a memcpy.
toBuffer(cb, "image/gif"); // maybe in the future
```

Alternatively, @adamhooper proposed separate `toPNG`/`toJPEG` functions.

Should there be sync versions that return a value instead of accepting a callback?

See also #934 about cleaning up the PNG filter API.

I can work on this sometime soon. Feedback welcome.

zbjornson:
Docs showing the new API: https://github.com/zbjornson/node-canvas/tree/tobufferjpeg#canvastobuffer

* **Added**: support for `canvas.toBuffer("image/jpeg")`. This unblocks #1146 and helps make `canvas.toBuffer()` **the** universal method for getting encoded data out of a canvas. Unblocks adding `toBlob` if we want it -- a later topic.

* **Changed (breaking)**: PNG filter and ZLIB compression options are now named instead of positional for `canvas.toBuffer` and (undocumented) `canvas.pngStream`. Less awkward API, easier support for more parameters in the future (including #716/#766).

* **Improved**: `canvas.toBuffer`, `canvas.jpegStream` and `canvas.pngStream` now all accept the same config objects. Previously some options were only available when streaming.

* **Removed (breaking)**: The weird argument handling that allowed the compression level `"0"` (string). The logic in that section was all sort of weird, not sure if I'm misreading it.

* **Changed (breaking)**: The quality argument for `jpegStream` now goes from 0 to 1 to match the quality argument in `canvas.toBlob`.

* **Question**: Should there be a `canvas.toBuffer(callback, "raw")`? **NO** per below. <strike>There's no async work to be done here (just a memcpy). Adding this would make the API consist, but would encourage a bad usage pattern (callback-all-the-things!). There's also no `canvas.toBuffer(callback)` support for SVG or PDF canvases.</strike>

* `image/jpeg` uses the threadpool. <strike>**Question**: `image/jpeg` runs in the main thread currently, but that *could* reasonably be offloaded to the threadpool later. It's easy enough to invoke the callback in the next tick now so that if it moves to the threadpool in the future it'll be async now and in the future -- should it? One less semver major change in the future.</strike>

Fixes #982

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,10 +10,10 @@ project adheres to [Semantic Versioning](http://semver.org/).
 
 **Upgrading from 1.x**
 ```js
-// (1) The quality argument for canvas.jpegStream now goes from 0 to 1 instead
-//     of from 0 to 100:
-canvas.jpegStream({quality: 50}) // old
-canvas.jpegStream({quality: 0.5}) // new
+// (1) The quality argument for canvas.createJPEGStream/canvas.jpegStream now
+//     goes from 0 to 1 instead of from 0 to 100:
+canvas.createJPEGStream({quality: 50}) // old
+canvas.createJPEGStream({quality: 0.5}) // new
 
 // (2) The ZLIB compression level and PNG filter options for canvas.toBuffer are
 //     now named instead of positional arguments:
@@ -29,10 +29,12 @@ canvas.pngStream({compressionLevel: 3, filters: canvas.PNG_FILTER_NONE}) // new
 
 // (4) canvas.syncPNGStream() and canvas.syncJPEGStream() have been removed:
 canvas.syncPNGStream() // old
-canvas.pngStream() // new
+canvas.createSyncPNGStream() // old
+canvas.createPNGStream() // new
 
 canvas.syncJPEGStream() // old
-canvas.jpegStream() // new
+canvas.createSyncJPEGStream() // old
+canvas.createJPEGStream() // new
 ```
 
 ### Breaking
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,12 +8,43 @@ project adheres to [Semantic Versioning](http://semver.org/).
 2.0.0 (unreleased -- encompasses all alpha versions)
 ==================
 
+**Upgrading from 1.x**
+```js
+// (1) The quality argument for canvas.jpegStream now goes from 0 to 1 instead
+//     of from 0 to 100:
+canvas.jpegStream({quality: 50}) // old
+canvas.jpegStream({quality: 0.5}) // new
+
+// (2) The ZLIB compression level and PNG filter options for canvas.toBuffer are
+//     now named instead of positional arguments:
+canvas.toBuffer(undefined, 3, canvas.PNG_FILTER_NONE) // old
+canvas.toBuffer(undefined, {compressionLevel: 3, filters: canvas.PNG_FILTER_NONE}) // new
+// or specify the mime type explicitly:
+canvas.toBuffer("image/png", {compressionLevel: 3, filters: canvas.PNG_FILTER_NONE}) // new
+
+// (3) #2 also applies for canvas.pngStream, although these arguments were not
+//     documented:
+canvas.pngStream(3, canvas.PNG_FILTER_NONE) // old
+canvas.pngStream({compressionLevel: 3, filters: canvas.PNG_FILTER_NONE}) // new
+
+// (4) canvas.syncPNGStream() and canvas.syncJPEGStream() have been removed:
+canvas.syncPNGStream() // old
+canvas.pngStream() // new
+
+canvas.syncJPEGStream() // old
+canvas.jpegStream() // new
+```
+
 ### Breaking
  * Drop support for Node.js <4.x
- * Remove sync streams (bc53059). Note that all or most streams are still
-   synchronous to some degree; this change just removed `syncPNGStream` and
-   friends.
+ * Remove sync stream functions (bc53059). Note that most streams are still
+   synchronous (run in the main thread); this change just removed `syncPNGStream`
+   and `syncJPEGStream`.
  * Pango is now *required* on all platforms (7716ae4).
+ * Make the `quality` argument for JPEG output go from 0 to 1 to match HTML spec.
+ * Make the `compressionLevel` and `filters` arguments for `canvas.toBuffer()`
+   named instead of positional. Same for `canvas.pngStream()`, although these
+   arguments were not documented.
 
 ### Fixed
  * Prevent segfaults caused by loading invalid fonts (#1105)
@@ -35,8 +66,8 @@ project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
  * Prebuilds (#992) with different libc versions to the prebuilt binary (#1140)
- * Support canvas.getContext("2d", {alpha: boolean}) and
-   canvas.getContext("2d", {pixelFormat: "..."})
+ * Support `canvas.getContext("2d", {alpha: boolean})` and
+   `canvas.getContext("2d", {pixelFormat: "..."})`
  * Support indexed PNG encoding.
  * Support `currentTransform` (d6714ee)
  * Export `CanvasGradient` (6a4c0ab)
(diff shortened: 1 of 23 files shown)


LinusU:
🙌 

Haven't looked too closely yet, but looks super good!

zbjornson:
Alright, finally ready for review! Original post edited (two remaining questions).

LinusU:
> Should there be a `canvas.toBuffer(callback, "raw")`? Adding this would make the API consist, but would encourage a bad usage pattern (callback-all-the-things!).

My vote goes to not adding it in.

> `image/jpe`g runs in the main thread currently, but that could reasonably be offloaded to the threadpool later. It's easy enough to invoke the callback in the next tick now so that if it moves to the threadpool in the future it'll be async now and in the future -- should it? 

Sounds good 👍

chearon:
Phew, this is gonna make things a lot less confusing. But why would someone use `toBuffer(callback, ...)` instead of `.pngStream()` or `.jpegStream()`? Is this to essentially match `HTMLCanvasElement.toBlob()`?

Since `.toBlob()` only supports JPEG/PNG, is there any reason we should have `.toBuffer("raw")`? It seems what `ImageData` is for, but maybe if people like it/are already using it we should leave it, but then there's the lack of `toBuffer(callback, "raw"` which might be confusing.

I don't know this part of the code that well, but am I right that this is also changing the behavior of `toBuffer()` without any arguments? It's a good change since it matches `.toBlob()` but I think this means people can't do `image.src = canvas.toBuffer()` like we say in the readme.

zbjornson:
> My vote goes to not adding [toBuffer(callback, "raw")] in.

Mine as well. We can add it later if someone has a compelling reason.

> why would someone use toBuffer(callback, ...) instead of .pngStream() or .jpegStream()

You get the whole buffer at once instead of dealing with a stream. (Merits of each discussed starting here: https://github.com/Automattic/node-canvas/issues/743#issuecomment-324102309.)

> any reason we should have .toBuffer("raw")

You mean vs. `canvas.getImageData()`? See @adamhooper's nice justification here: https://github.com/Automattic/node-canvas/pull/819#issuecomment-353054345 and https://github.com/Automattic/node-canvas/issues/306#issuecomment-252067945

> am I right that this is also changing the behavior of toBuffer() without any arguments

No change there! There's still the synchronous version that returns a buffer if you don't provide a callback for the first arg.

---

I'll work on making JPEG streams pseudo-async shortly -- I have that in some branch from a year ago.

zbjornson:
Okay, ready to go! I'm sorry this got so big, but I decided to clean up `closure_t` now instead of making it a bigger Frankenstein's monster.

`toBuffer(callback, "image/jpeg")` is truly async, using the libuv threadpool, not just `nextTick`.

Removed a redundant memcpy that was called in a loop.

Can someone with MacOS check that this builds please? (I enabled CPP exceptions in the gyp file and can only test on Win and Linux.)

chearon:
Builds on macOS for me! Will try to provide bigger feedback soon.

jfizz:
Hi @zbjornson, I tried your branch on my macbook. The sync version `canvas.toBuffer('image/jpeg', {quality: 0.5})` worked but the async version `canvas.toBuffer((err, buf) => {}, 'image/jpeg', {quality: 0.95})` failed with the error `Error: <unknown error status>`. Any thoughts? Below are my versions of the required dependencies.
```
node 8.9.4
pkg-config 0.29.2
cairo 1.14.12
pango 1.42.1
libpng 1.6.34
jpeg 9b
giflib 5.1.4
```
Also, after reading your comments above about `toBuffer(callback, "image/jpeg")` being truly async, the following line in the docs should probably be updated:
```
canvas.toBuffer((err, buf) => {
  if (err) throw err; // encoding failed
  // buf is JPEG-encoded image at 95% quality
  // Note that this callback is currently called synchronously.
}, 'image/jpeg', {quality: 0.95})
```

zbjornson:
@jfizz thanks for trying it out and for catching the doc error.

That message sounds like it comes from libjpeg... Could you post a repro case for it please, or does it happen with any canvas?

jfizz:
Here is the repo case.
```
const {createCanvas} = require('canvas');
canvas.toBuffer((err, buf) => {
  console.log(err);
  console.log(buf);
}, 'image/jpeg', {quality: 0.95});
```
However, the plot thickens. I ran the above script 10 times in a row and it worked 7 / 10 times. For the 3 times it failed the error was `Error: <unknown error status>`.

chearon:
Can confirm same behavior on macOS 

```
$ node test.js
null
<Buffer ff d8 ff e0 00 10 4a 46 49 46 00 01 01 00 00 01 00 01 00 00 ff db 00 43 00 08 06 06 07 06 05 08 07 07 07 09 09 08 0a 0c 14 0d 0c 0b 0b 0c 19 12 13 0f ... >
$ node test.js
null
<Buffer ff d8 ff e0 00 10 4a 46 49 46 00 01 01 00 00 01 00 01 00 00 ff db 00 43 00 08 06 06 07 06 05 08 07 07 07 09 09 08 0a 0c 14 0d 0c 0b 0b 0c 19 12 13 0f ... >
$ node test.js
Error: <unknown error status>
undefined
```

zbjornson:
Fixed docs and crash, thanks for finding. Want to give it another try please?

jfizz:
Works for me!

chearon:
Not sure if it belongs in this PR or not but `createPNGStream` and `createJPEGStream` aren't documented, but are used in the examples instead of `pngStream` and `jpegStream`. Choosing one or the other might further reduce confusion.

zbjornson:
I guess `create` is idiomatic (following `fs.createReadStreaam` for example)...

LinusU:
Looks awesome!

KFoxder:
@chearon @LinusU @zbjornson So I was trying to use the `resolution` option like so:
```js
canvas.toBuffer('image/png', { compressionLevel: 0, resolution: 200 });
```

And I noticed no difference in the actual PPI of the image. I went digging through this code and I found the `parsePNGArgs` method. Nowhere in this code does it ever parse the `resolution` option. Can you confirm that this is a bug and not intended or that I am missing something?

```cc
static void parsePNGArgs(Local<Value> arg, PngClosure& pngargs) {
  if (arg->IsObject()) {
    Local<Object> obj = Nan::To<Object>(arg).ToLocalChecked();

    Local<Value> cLevel = obj->Get(Nan::New("compressionLevel").ToLocalChecked());
    if (cLevel->IsUint32()) {
      uint32_t val = Nan::To<uint32_t>(cLevel).FromMaybe(0);
      // See quote below from spec section 4.12.5.5.
      if (val <= 9) pngargs.compressionLevel = val;
    }

    Local<Value> filters = obj->Get(Nan::New("filters").ToLocalChecked());
    if (filters->IsUint32()) pngargs.filters = Nan::To<uint32_t>(filters).FromMaybe(0);

    Local<Value> palette = obj->Get(Nan::New("palette").ToLocalChecked());
    if (palette->IsUint8ClampedArray()) {
      Local<Uint8ClampedArray> palette_ta = palette.As<Uint8ClampedArray>();
      pngargs.nPaletteColors = palette_ta->Length();
      if (pngargs.nPaletteColors % 4 != 0) {
        throw "Palette length must be a multiple of 4.";
      }
      pngargs.nPaletteColors /= 4;
      Nan::TypedArrayContents<uint8_t> _paletteColors(palette_ta);
      pngargs.palette = *_paletteColors;
      // Optional background color index:
      Local<Value> backgroundIndexVal = obj->Get(Nan::New("backgroundIndex").ToLocalChecked());
      if (backgroundIndexVal->IsUint32()) {
        pngargs.backgroundIndex = static_cast<uint8_t>(Nan::To<uint32_t>(backgroundIndexVal).FromMaybe(0));
      }
    }
  }
}
```

PR #1153, opened 2018-04-21
disableChromaSubsampling: false -> chromaSubsampling: true

zbjornson:
@thenickdude what do you think?


zbjornson:
Came with #1152.

PR #1171, opened 2018-05-28
Allow assigning non-string values to fillStyle and strokeStyle

Hakerh400:
When a non-string value is assigned to `fillStyle` or `strokeStyle` browsers convert it to string, but node-canvas does nothing in such case (except if the value is an instance of gradient or pattern).

This fix allows arbitrary object to be used as a color.

- [x] Updated the changelog
- [x] Added a test

### DIFF
--- a/lib/context2d.js
+++ b/lib/context2d.js
@@ -131,7 +131,7 @@ Context2d.prototype.__defineSetter__('fillStyle', function(val){
     this._setFillPattern(val);
   } else {
     this.lastFillStyle = undefined;
-    this._setFillColor(val + '');
+    this._setFillColor(String(val));
   }
 });
 
@@ -158,7 +158,7 @@ Context2d.prototype.__defineSetter__('strokeStyle', function(val){
     this.lastStrokeStyle = val;
     this._setStrokePattern(val);
   } else {
-    this._setStrokeColor(val + '');
+    this._setStrokeColor(String(val));
   }
 });
 
--- a/lib/context2d.js
+++ b/lib/context2d.js
@@ -131,7 +131,7 @@ Context2d.prototype.__defineSetter__('fillStyle', function(val){
     this._setFillPattern(val);
   } else {
     this.lastFillStyle = undefined;
-    this._setFillColor(val);
+    this._setFillColor(val + '');
   }
 });
 
@@ -158,7 +158,7 @@ Context2d.prototype.__defineSetter__('strokeStyle', function(val){
     this.lastStrokeStyle = val;
     this._setStrokePattern(val);
   } else {
-    this._setStrokeColor(val);
+    this._setStrokeColor(val + '');
   }
 });
 
--- a/lib/context2d.js
+++ b/lib/context2d.js
@@ -130,7 +130,7 @@ Context2d.prototype.__defineSetter__('fillStyle', function(val){
     || 'CanvasPattern' == val.constructor.name) {
     this.lastFillStyle = val;
     this._setFillPattern(val);
-  } else if ('string' == typeof val) {
+  } else {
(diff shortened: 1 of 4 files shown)


src/CanvasRenderingContext2d.cc
```diff
@@ -1802,8 +1802,17 @@ NAN_GETTER(Context2d::GetShadowColor) {
...
+  Isolate *isolate = Isolate::GetCurrent();
+  v8::TryCatch try_catch(isolate);
+  MaybeLocal<String> color = info[0]->ToString(Context::New(isolate));
+
+  if (try_catch.HasCaught()) {
+    try_catch.ReThrow();
+    return;
+  }
+
+  String::Utf8Value str(color.ToLocalChecked());
```
zbjornson: I think it'd be easier to just do `this._setFillColor(String(val))` in the JS file and let it throw there if it fails. But if you want to keep this in C++, can you switch to `Nan::TryCatch` please?

lib/context2d.js
```diff
@@ -125,12 +125,11 @@ Object.defineProperty(Context2d.prototype, 'currentTransform', {
  */
 
 Context2d.prototype.__defineSetter__('fillStyle', function(val){
-  if (!val) return;
-  if ('CanvasGradient' == val.constructor.name
-    || 'CanvasPattern' == val.constructor.name) {
+  if (val instanceof CanvasGradient
+    || val instanceof CanvasPattern) {
```
zbjornson: 👍

Hakerh400:
@zbjornson 

IMHO, C++ checks are more reliable and faster than JavaScript ones.
I've replaced v8::TryCatch with Nan::TryCatch.

LinusU:
> C++ checks are more reliable and faster than JavaScript ones.

I don't think this is true, since (as far as I know) the most costly in V8 is to jump between JavaScript and C++.  That being said, it probably won't make much of a difference...

Hakerh400:
@LinusU 

I did a performance test

```js
var color = ['#ff0000'];
var t = Date.now();
for(var i = 0; i < 1e5; i++)
  ctx.fillStyle = color;
console.log((Date.now() - t) / 1e3);
```

|C++ checks|JS checks|
|:-:|:-:|
|38.237|0.045|

So definitely you're right. I moved all checks and stringifyings to JS.

lib/context2d.js
```diff
@@ -125,14 +125,13 @@ Object.defineProperty(Context2d.prototype, 'currentTransform', {
...
     this.lastFillStyle = val;
     this._setFillPattern(val);
-  } else if ('string' == typeof val) {
+  } else {
     this.lastFillStyle = undefined;
-    this._setFillColor(val);
+    this._setFillColor(val + '');
```
zbjornson: `val + ''` uses `val.valueOf()` instead of `toString` if there's a `valueOf` method. Whatever  Chrome does internally uses `toString` (e.g. `this._setFillColor(String(val))`.

```js
> let x = {toString: () => "#0000ff", valueOf: () => "#00ff00"}
> ctx.fillStyle = x
> ctx.fillStyle
"#0000ff"
```

LinusU: also, I think `String(val)` makes the intent more clear ☺️

lib/context2d.js
zbjornson: That would only work by coincidence I think. We've never tried to freeze those class structures.

LinusU: Cool 👍

zbjornson:
Rough numbers for future reference:
Calling into C++ from JS takes about 8-15 ns.
Calling into JS/v8 from C++ takes about 60-70 ns.

PR #1177, opened 2018-06-08
Suppress deprecation warning on NodeJS v10

nishidayuya:
Thanks for contributing!


zbjornson:
Thanks for this!

Node 4.x is EOL as of 4-Apr-2018. Should we just require Node.js 6.x, which has `Buffer.from()`? @chearon @LinusU

nishidayuya:
[`Buffer.from`](https://nodejs.org/api/buffer.html#buffer_class_method_buffer_from_string_encoding) and [`Buffer.allocUnsafe`](https://nodejs.org/api/buffer.html#buffer_class_method_buffer_allocunsafe_size) are added in Node.js 5.10.0. So, if node-canvas drop Node.js 4.x support, [`BufferCreator`](https://github.com/Automattic/node-canvas/pull/1177/commits/ab7d4e738da160844da5c5916279e3cdad6a9b42#diff-6012519ec52c046301c8c69fe01f0793R9) is not needed.

Should I remove `BufferCreator`?

LinusU:
> Should we just require Node.js 6.x

Lets! 🙌

zbjornson:
@nishidayuya that's a yes! You can drop the `BufferCreator` and just use `Buffer.from`. While you're at it, can you remove node version 4 from .travis.yml please?

zbjornson:
... and this line of Readme.md please:

> Currently the minimum version of node required is 4.0.0

nishidayuya:
@zbjornson Thank you for your reply! I dropped Node.js 4.x support with updating `package.json` and removed `BufferCreator`.

zbjornson:
Awesome, thanks!

nishidayuya:
:tada:

PR #1179, opened 2018-06-11
Add resolution option for PNG format

### LINKED ISSUE #716: Output high resolution PNG

### LINKED ISSUE #766: PNG files written with wrong resolution in header
Currently, a PNG encoded by node-canvas doesn't contain the correct physical size information in the header. In `canvas_write_png()` in `PNG.h` you never call `png_set_pHYs()` to set the physical units in the PNG header, so it defaults to 72 PPI. Most screens are 96 PPI.

Generally it would be awesome to have a way to set the physical size of the canvas from javascript. I personally would like to use this for print, where the DPI in the header changes the way the document is printed. Right now I am running an imagemagick command after the PNG is written to a file, but this is inefficient.

I could offer a patch, but of course it would be another non-standard function. I'd like to hear the thoughts of the project leaders on whether this would be something of general interest or not before I write anything.

This is potentially related to Issue #716, in that scaling the canvas may indicate that the user is looking to produce a PNG for higher PPI screens, and the PNG/SVG/PDF should probably reflect the true physical dimensions of the display device.

zbjornson:
> It seems browsers also always set the PPI to 72 
> -- https://github.com/Automattic/node-canvas/issues/766#issuecomment-218946252

Edge, Firefox and Chrome don't specify a `pHYs` chunk, so I left default behavior unchanged to stay consistent with them. (The internets say that the resolution doesn't mean much for on-screen display anyway.)

Fixes #766
Fixes #716

Thanks for contributing!

### DIFF
--- a/test/canvas.test.js
+++ b/test/canvas.test.js
@@ -497,6 +497,26 @@ describe('Canvas', function () {
       var buf = createCanvas(200,200).toBuffer('image/png');
       assert.equal('PNG', buf.slice(1,4).toString());
     });
+
+    it('Canvas#toBuffer("image/png", {resolution: 96})', function () {
+      const buf = createCanvas(200, 200).toBuffer('image/png', {resolution: 96});
+      // 3780 ppm ~= 96 ppi
+      for (let i = 0; i < buf.length - 12; i++) {
+        if (buf[i] === 0x70 &&
+          buf[i + 1] === 0x48 &&
+          buf[i + 2] === 0x59 &&
+          buf[i + 3] === 0x73) { // pHYs
+          assert.equal(buf[i + 4], 0);
+          assert.equal(buf[i + 5], 0);
+          assert.equal(buf[i + 6], 0x0e);
+          assert.equal(buf[i + 7], 0xc4); // x
+          assert.equal(buf[i + 8], 0);
+          assert.equal(buf[i + 9], 0);
+          assert.equal(buf[i + 10], 0x0e);
+          assert.equal(buf[i + 11], 0xc4); // y
+        }
+      }
+    })
   
     it('Canvas#toBuffer("image/png", {compressionLevel: 5})', function () {
       var buf = createCanvas(200,200).toBuffer('image/png', {compressionLevel: 5});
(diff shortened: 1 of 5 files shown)


test/canvas.test.js
```diff
@@ -497,6 +497,26 @@ describe('Canvas', function () {
...
     });
+
+    it('Canvas#toBuffer("image/png", {resolution: 96})', function () {
+      const buf = createCanvas(200, 200).toBuffer('image/png', {resolution: 96});
+      // 3780 ppm ~= 96 ppi
+      for (let i = 0; i < buf.length - 12; i++) {
+        if (buf[i] === 0x70 &&
+          buf[i + 1] === 0x48 &&
+          buf[i + 2] === 0x59 &&
+          buf[i + 3] === 0x73) { // pHYs
```
LinusU: 😄

KFoxder:
@zbjornson I do not see this working for me? Don't you need to update the `parsePNGArgs` method in this PR?

zbjornson:
Hrm it looks like I lost the changes to canvas.cc that were supposed to be in this PR. The test passes because pHYs is omitted if resolution isn't specified. Will fix tonight. Thanks for pointing that out.

KFoxder:
@zbjornson Awesome thanks! Much appreciated.

zbjornson:
@mikehalmamoj yes, #1376.

image

As an ancillary chunk, readers are allowed to ignore pHYs. That might be why you don't see it doing anything. Above shows that the chunk is encoded in the file correctly though.

mikehalmamoj:
Thanks for the quick reply @zbjornson. That's ever so strange because I don't see the resolution in the image info when generating from this lib but if I add a resolution with the [pngcrush CLI](http://manpages.ubuntu.com/manpages/bionic/man1/pngcrush.1.html) then I see the that resolution. 🤷‍♂️  I'll have to do some more digging.

Thanks again!

mikehalmamoj:
You're absolutely correct and it works fine - it seems the issue is the haphazard support for the `pHYs` chunk. Thanks again @zbjornson and keep up the good work! :smile:

PR #1187, opened 2018-06-23
Documentation fix: async to sync example

adtm:
`writeFile` should ask for a callback if it's async, in the example, the arguments are for sync, although it is used as async


LinusU:
👍

Issue #1159, opened 2018-05-07
C4324: 'canvas_jpeg_error_mgr': structure was padded due to alignment specifier & 'v8::String::Utf8Value::Utf8Value': was declared deprecated


advanceddeveloper:
Hello, its me again

I am trying to build node canvas from source, but I am getting few warnings:

```console
warning C4996: 'v8::String::Utf8Value::Utf8Value': was declared deprecated
warning C4324: 'canvas_jpeg_error_mgr': structure was padded due to alignment specifier
```

The first one is repeated multiple times and in multiple files, but the second one appears in image.cc, so I am guessing utf8value:utf8value should be replaced with something other, since it says it is deprecated. I dont get such warning in node v8.5.0, but only in node 10.

Is the second warning indicating that something is wrong, or are you aware of it? I mean, it says that something is padded, should it be padded, or should it not be padded?

r_

zbjornson:
The alignment warning can be ignored. A few more bytes will be used but no harm otherwise. We can either silence it, or embed the base `jpeg_err_mgr` in another struct instead of extending it.

advanceddeveloper:
> We can silence it

That would be nice. Every time I compile I think something is wrong when I see yellow text all over my screen.

But what about Utf8Value deprecation? I think it will be permanently removed in a few semvermajor node versions (when it gets removed from v8), which will break build completely. I think v8 team explained in their blog which alternative functions should be used instead.

zbjornson:
Yep, definitely a to-do.

PR #1191, opened 2018-06-29
Remove deprecation warning for v8::String::Utf8Value

Hakerh400:
* Fix deprecation warning related to String::Utf8Value
* Suppress alignment warning

Fixes https://github.com/Automattic/node-canvas/issues/1159

---

Also, there was a test which was failing at master (before this fix)

https://github.com/Automattic/node-canvas/blob/013223f92afdbb3265a7073368b9800fbacc40cf/test/canvas.test.js#L550-L558

Am I missing something here? Shouldn't it be like

```diff
- var buf = createCanvas(200,200).toBuffer(function (err, buff) {
+ createCanvas(200,200).toBuffer(function (err, buf) {
```

since the function is asynchronous and `var buf` is `undefined`?

### DIFF
--- a/src/Canvas.cc
+++ b/src/Canvas.cc
@@ -93,9 +93,9 @@ NAN_METHOD(Canvas::New) {
     if (info[1]->IsNumber()) height = info[1]->Uint32Value();
 
     if (info[2]->IsString()) {
-      if (0 == strcmp("pdf", *String::Utf8Value(Isolate::GetCurrent(), info[2])))
+      if (0 == strcmp("pdf", *Nan::Utf8String(info[2])))
         backend = new PdfBackend(width, height);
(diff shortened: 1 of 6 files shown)


zbjornson:
Thanks!

The `(isolate, obj)` form of `Utf8Value` isn't available in Node 6.x. If you switch everything to https://github.com/nodejs/nan/blob/master/doc/v8_misc.md#api_nan_utf8_string then it should work across versions.

zbjornson:
> Shouldn't it be like `createCanvas(200,200).toBuffer(function (err, buf) {`

Yeah... 🤔 . The test should also probably be `it('label...', function (done) {` and call `done` from within the callback, right?

Hakerh400:
> The (isolate, obj) form of Utf8Value isn't available in Node 6.x

Thanks. Fixed.

PR #1193, opened 2018-07-01
Clean up img.src=

### LINKED ISSUE #564: Image#src vs Image#source: Image#src doesn't exist, but is in README.md
The examples given in the readme.md on master are wrong, as the only way to load an external buffer is by using Image#source. Setting Image#src doesn't do anything.

### LINKED ISSUE #807: Support for SVG as data URL missing
I'm using the library dom-to-image (in node.js) to convert some DIV content to an image. The library adds the DIV to an SVG foreign object, converts the SVG to a data URL and updates the src property of a canvas image. So far so good. However, the last step fails with the message `Error: error while reading from input stream`.

From other issues, I found that the log message is thrown in Image.cc and that it comes from the variable CAIRO_STATUS_READ_ERROR. Looking at the behaviour of Image.cc, it looks like the above functionality is not supported. In the load methods (load/loadFromBuffer), there is only support for PNG, GIF and JPEG. 

I understood that there was already support for SVG. Does this include the above functionality?
- If so, would it be possible to indicate how I can resolve the above issue? 
- If not, is there another way to convert an SVG data URL to a PNG image? 
### Test code

```
var svgString = `
  <svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" version="1.1" width="400" height="400" viewBox="0 0 100 100">
    <rect x="0" y="0" width="100" height="100" fill="green"></rect>
    <rect x="0" y="0" width="50" height="50" fill="blue"></rect>
  </svg>
`.trim();

var svgData = 'data:image/svg+xml;charset=utf-8,' + svgString;

var svg = new Canvas.Image;
svg.onload = function() {
    console.log("TODO create png from svg image");
}
svg.onerror = function(error) {
    console.log("ERROR: " + error);
};
svg.src = svgData;
```

### LINKED ISSUE #945: Support img.src = <url>

### LINKED ISSUE #1079: SVG load error
SVG load error 

Image given has not completed loading

zbjornson:
Fixes #945 - support `img.src = <url>`
Fixes #807 - support for non-base64 `data:` URIs
Fixes #1079 - ditto
Closes #564 - it works, probably didn't pass a string or buffer
Makes all examples and the README use onload, onerror.

### DIFF
--- a/lib/image.js
+++ b/lib/image.js
@@ -34,7 +34,13 @@ Object.defineProperty(Image.prototype, 'src', {
         val = val.slice(commaI + 1)
         this.source = Buffer.from(val, isBase64 ? 'base64' : 'utf8')
       } else if (/^\s*http/.test(val)) { // remote URL
-        const onerror = err => typeof this.onerror === 'function' && this.onerror(err)
+        const onerror = err => {
+          if (typeof this.onerror === 'function') {
+            this.onerror(err)
+          } else {
+            throw err
+          }
+        }
         http.get(val, res => {
           if (res.statusCode !== 200) {
             return onerror(new Error(`Server responded with ${res.statusCode}`))
--- a/lib/image.js
+++ b/lib/image.js
@@ -12,35 +12,51 @@
 
 const bindings = require('./bindings')
 const Image = module.exports = bindings.Image
+const http = require("http")
 
-/**
- * Src setter.
- *
- *  - convert data uri to `Buffer`
- *
- * @param {String|Buffer} val filename, buffer, data uri
- * @api public
- */
+Object.defineProperty(Image.prototype, 'src', {
+  /**
+   * src setter. Valid values:
+   *  * `data:` URI
+   *  * Local file path
+   *  * HTTP or HTTPS URL
+   *  * Buffer containing image data (i.e. not a `data:` URI stored in a Buffer)
+   *
+   * @param {String|Buffer} val filename, buffer, data URI, URL
+   * @api public
+   */
+  set(val) {
+    if (typeof val === 'string') {
+      if (/^\s*data:/.test(val)) { // data: URI
+        const commaI = val.indexOf(',')
+        // 'base64' must come before the comma
+        const isBase64 = val.lastIndexOf('base64', commaI)
+        val = val.slice(commaI + 1)
+        this.source = Buffer.from(val, isBase64 ? 'base64' : 'utf8')
+      } else if (/^\s*http/.test(val)) { // remote URL
+        const onerror = err => typeof this.onerror === 'function' && this.onerror(err)
+        http.get(val, res => {
+          if (res.statusCode !== 200) {
+            return onerror(new Error(`Server responded with ${res.statusCode}`))
+          }
+          const buffers = []
+          res.on('data', buffer => buffers.push(buffer))
+          res.on('end', () => {
+            this.source = Buffer.concat(buffers)
+          })
+        }).on('error', onerror)
+      } else { // local file path assumed
+        this.source = val
+      }
+    } else if (Buffer.isBuffer(val)) {
+      this.source = val
+    }
+  },
 
-Image.prototype.__defineSetter__('src', function(val){
-  if ('string' == typeof val && 0 == val.indexOf('data:')) {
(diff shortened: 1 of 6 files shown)


lib/image.js
```diff
@@ -12,37 +12,53 @@
...
+  set(val) {
+    if (typeof val === 'string') {
+      if (/^\s*data:/.test(val)) { // data: URI
+        const commaI = val.indexOf(',')
+        // 'base64' must come before the comma
+        const isBase64 = val.lastIndexOf('base64', commaI)
+        val = val.slice(commaI + 1)
+        this.source = Buffer.from(val, isBase64 ? 'base64' : 'utf8')
+      } else if (/^\s*http/.test(val)) { // remote URL
+        const onerror = typeof this.onerror === 'function' ? this.onerror : () => {}
```
LinusU: I _think_ that it's not too uncommon to do something like:

```js
img.src = '...'
img.onerror = ...
```

How about something like:

```js
const onerror = (err) => typeof this.onerror === 'function' && this.onerror(err)
```

LinusU: Good point, we could at the very least just log an error? Sort of like when a Promise rejects and you haven't caught it:

```
> const a = Promise.reject(new Error('Something went wrong'))
undefined
> (node:20870) UnhandledPromiseRejectionWarning: Error: Something went wrong
    at repl:1:26
    at Script.runInThisContext (vm.js:91:20)
    at REPLServer.defaultEval (repl.js:317:29)
    at bound (domain.js:396:14)
    at REPLServer.runBound [as eval] (domain.js:409:12)
    at REPLServer.onLine (repl.js:615:10)
    at REPLServer.emit (events.js:187:15)
    at REPLServer.EventEmitter.emit (domain.js:442:20)
    at REPLServer.Interface._onLine (readline.js:290:10)
    at REPLServer.Interface._line (readline.js:638:8)
(node:20860) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
```

zbjornson: Now that I think about it more, it might be a bad idea...

The `UnhandledPromiseRejectionWarning` can be silenced (`node --no-warnings`). There'd be no equivalent way for a user to silence an `onerror` console message. I also can't think of a stream or other event emitter that causes a warning or message if you don't listen to its `"error"` event.

Hopefully having the docs always show using the `onerror` handler is enough.

LinusU: Every event emitter, streams included, actually crashes the entire process if you emit an `error` and there is no listeners for it.

```text
➜  /tmp cat test.js 
const events = require('events')
const ee = new events.EventEmitter()

ee.emit('error', new Error('Hello, World!'))

setTimeout(() => console.log('Nothing happened...'), 2000)
➜  /tmp node test.js
events.js:160
      throw er; // Unhandled 'error' event
      ^

Error: Hello, World!
    at Object.<anonymous> (/private/tmp/test.js:4:18)
    at Module._compile (module.js:570:32)
    at Object.Module._extensions..js (module.js:579:10)
    at Module.load (module.js:487:32)
    at tryModuleLoad (module.js:446:12)
    at Function.Module._load (module.js:438:3)
    at Module.runMain (module.js:604:10)
    at run (bootstrap_node.js:389:7)
    at startup (bootstrap_node.js:149:9)
    at bootstrap_node.js:502:3
➜  /tmp 
```

lib/image.js
```diff
@@ -12,37 +12,59 @@
...
+          const buffers = []
+          res.on('data', buffer => buffers.push(buffer))
+          res.on('end', () => {
+            this.source = Buffer.concat(buffers)
+          })
+        }).on('error', onerror)
+      } else { // local file path assumed
+        this.source = val
+      }
+    } else if (Buffer.isBuffer(val)) {
```
chearon: Slight change in behavior: it used to always set `source` but now it only sets it if it's a string in the right format or buffer. New behavior is probably better since that's how `ctx.font` etc work.

chearon:
(Need to fix conflict with the changelog)

zbjornson:
and with that merged ... image is less than 200! 🎉

PR #1194, opened 2018-07-02
Support canvas.toDataURL("image/jpeg") (sync)

### LINKED ISSUE #1146: Missing sync canvas.toDataURL("image/jpeg") API

zbjornson:
This adds the missing, standard sync API; we only had async.

Also fixes a bug I introduced in #1152 (JPEG quality needs to go from 0 to 1, not 0 to 100).

Fixes #1146

Also allows some code to be freed up in JSDOM (https://github.com/jsdom/jsdom/issues/2199).

### DIFF
--- a/test/canvas.test.js
+++ b/test/canvas.test.js
@@ -663,38 +663,31 @@ describe('Canvas', function () {
     ctx.fillRect(100,0,100,100);
 
     it('toDataURL() works and defaults to PNG', function () {
-      assert.ok(0 == canvas.toDataURL().indexOf('data:image/png;base64,'));
+      assert.ok(canvas.toDataURL().startsWith('data:image/png;base64,'));
     });
 
     it('toDataURL(0.5) works and defaults to PNG', function () {
-      assert.ok(0 == canvas.toDataURL(0.5).indexOf('data:image/png;base64,'));
+      assert.ok(canvas.toDataURL(0.5).startsWith('data:image/png;base64,'));
     });
 
     it('toDataURL(undefined) works and defaults to PNG', function () {
-      assert.ok(0 == canvas.toDataURL(undefined).indexOf('data:image/png;base64,'));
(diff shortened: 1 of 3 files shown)


test/canvas.test.js
```diff
@@ -663,38 +663,31 @@ describe('Canvas', function () {
     ctx.fillRect(100,0,100,100);
 
     it('toDataURL() works and defaults to PNG', function () {
-      assert.ok(0 == canvas.toDataURL().indexOf('data:image/png;base64,'));
+      assert.ok(canvas.toDataURL().startsWith('data:image/png;base64,'));
```
LinusU: 🙌

asturur:
@zbjornson  This is very cool!
Do you think it can be backported on 1.X or is heavily based on 2.x code?

zbjornson:
It's heavily 2.x, but once the memory leaks are addressed I think 2.x can be released.

PR #1199, opened 2018-07-08
Explicitly export some C++ symbols

zbjornson:
This only affects the C++ API on Windows, where symbols are not always exported by default.

Linking against node-canvas used to happen to work, until backends were added and the combination of inlining lead to some missing symbols.

This is the correct thing to do when offering a C++ API. For example, node does it in node.h: https://github.com/nodejs/node/blob/602da6492f278c1345f7f2d82014d9248254476b/src/node.h#L25-L33 and v8: https://github.com/nodejs/node/blob/ca480719199d2ff38223aff8e301aced25d7e6f1/deps/v8/src/base/base-export.h. The directives for GCC are unnecessary since we don't use DLL_LOCAL yet and it exports by default, but using DLL_LOCAL in the future can offer some significant performance benefits (see https://gcc.gnu.org/wiki/Visibility).

Thanks for contributing!

---

For now, this API is sufficient to get the cairo surface, its data pointer, its width/height and its type. That's enough to write `toBuffer` encoder formats (e.g. webp, which I'm playing with now as a pluggable extension).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -69,6 +69,7 @@ canvas.createJPEGStream() // new
  * Fix DEP0005 deprecation warning
  * Don't assume `data:` URIs assigned to `img.src` are always base64-encoded
  * Fix formatting of color strings (e.g. `ctx.fillStyle`) on 32-bit platforms
+ * Explicitly export symbols for the C++ API
 
 ### Added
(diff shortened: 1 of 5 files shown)


LinusU:
@chearon using Cmake.js? I would love to merge that 😍

PR #1201, opened 2018-07-10
CSS colors: case-insensitive, faster lookup

### LINKED ISSUE #235: Setting fillStyle or strokeStyle to colors with capital letters causes it to fail
Setting a fillStyle or strokeStyle to a CSS color string with a capital letter such as `LightGreen` or `Blue` will be ignored. The following example shows that clearly when one rect is set to fill as `red` and the other `Blue`. Both end up red, as the Blue fill is never set.

``` javascript
// Module dependencies
var http = require("http");

// Navigating to lochalhost:8888 will show an image of a Diagram made server-side with Node.js
http.createServer(function(request, response) {
    , canvas = new Canvas(200,200)
    , ctx = canvas.getContext('2d');

  ctx.fillStyle = 'red';
  ctx.fillRect(0,0,50,50);
  ctx.fillStyle = 'Blue';
  ctx.fillRect(50,50,50,50);


  response.writeHeader(200, {"Content-Type": "text/html"});
  response.write('<img src="' + canvas.toDataURL() + '" />');
  response.end();
}).listen(8888);
```

zbjornson:
Fixes #235

Also improves performance:

| color | AoS loop | `std::map` |
| --- | ---: | ---: |
| "whitesmoke" (near end of array) | 668,905 | 2,764,503 |
| "transparent" (beginning of array) | 4,519,027 | 4,578,934 |

### DIFF
--- a/src/color.cc
+++ b/src/color.cc
@@ -8,7 +8,8 @@
 #include <cstdlib>
 #include <cmath>
 #include <limits>
-
+#include <string>
+#include <algorithm>
 #include "color.h"
 
 // Compatibility with Visual Studio versions prior to VS2015
@@ -398,7 +399,6 @@ static struct named_color {
   , { "whitesmoke", 0xF5F5F5FF }
   , { "yellow", 0xFFFF00FF }
   , { "yellowgreen", 0x9ACD32FF }
-  , { NULL, 0 }
 };
 
 /*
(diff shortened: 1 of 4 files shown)


src/color.cc
```diff
@@ -727,11 +724,11 @@ rgba_from_hex_string(const char *str, short *ok) {
...
-  int i = 0;
-  struct named_color color;
-  while ((color = named_colors[i++]).name) {
-    if (*str == *color.name && 0 == strcmp(str, color.name))
-      return *ok = 1, color.val;
+  std::string lowered(str);
+  std::transform(lowered.begin(), lowered.end(), lowered.begin(), tolower);
+  auto color = named_colors.find(lowered);
+  if (color != named_colors.end()) {
+    return *ok = 1, color->second;
```
LinusU: I'm not sure which C++ standard we are targeting, but this function could return `std::optional<int32_t>` and drop the `ok` parameter if we target C++17

zbjornson: That'd be nice. [Node targets C++14](https://github.com/nodejs/node/blob/573744caaedf2f941e152b8e9533cd5629d55c1c/common.gypi#L348) currently though, so I'm afraid to target past that.

PR #1204, opened 2018-07-17
Better error handling

### LINKED ISSUE #283: Corrupt JPG images don't always trigger onerror()
When creating an Image by setting img.src to raw binary image data I need to be able to catch errors if the image binary could not be interpreted as a valid image. In my tests blatantly bogus or corrupted image data which has no JPG header will nicely trigger the .onerror() method, giving me something I can catch and report back to the client.

However, in certain cases with a specifically corrupted JPG it is possible to end up in a limbo state where canvas hangs, outputting a couple lines of error messages to the console, but never triggering either the .onerror() or .oncomplete() methods.

I have experienced this most often with image binary data in which it has all the headers of a JPG file but the contents are damaged.

Example console output from Canvas:

```
Corrupt JPEG data: 147 extraneous bytes before marker 0xc4
Invalid JPEG file structure: SOS before SOF
```

Neither .onerror() or .oncomplete() were ever called, resulting in no callback to my application code, and therefore a nasty hang.

Link to a corrupted JPG file which triggers this issue with Canvas:

https://s3.amazonaws.com/storydesk/tests/assets/damaged.jpg

Is there another way that I can catch these type of errors nicely through Canvas instead of having to write my own custom code wrapper to detect if Canvas has hung without calling either the error or complete callback?

### LINKED ISSUE #486: 'Image given has not completed loading' is misleading.
If you try to load an invalid image file, node-canvas reports the misleading error 'Image given has not completed loading', rather than reporting that the image file is invalid.

To reproduce, first create an invalid image:

``` bash
echo 'File not found.' > test.png
```

Then try this script:

``` js
var fs = require("fs"),
    Canvas = require("canvas"),
    Image = Canvas.Image;

var image = new Image;
image.src = fs.readFileSync("test.png");

var canvas = new Canvas(256, 256),
    context = canvas.getContext("2d");
context.drawImage(image, 0, 0); // Error: Image given has not completed loading
```

It’d be even better if the error were reported when `image.src` is assigned, rather than deferring it until context.drawImage.

### LINKED ISSUE #1048: LoadImage throwing error (Error while reading from input stream)
## Issue or Feature

When I try to use loadImage I get this error

```

(node:1016) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 2): Error: error while reading from input stream

```

This is the code that actually centers around the error

```js

const loadImageIf = (condition, src, func) => {

  if(condition) loadImage(src).then((image) => {func(image)}); else func()

}

```

And then later in my code

```

loadImageIf(targetStats.profileSettings.usedBackground != undefined, `./../backgrounds/${client.config.commands.profile.backgrounds[targetStats.profileSettings.usedBackground]}`, (image) => {

  if(image) ctx.drawImage(0, 0, config.width, config.height)



  ...More stuff to add to the canvas here, not more images...



  message.channel.send({ files: [canvas.toBuffer()] })

})

```

So let me explain, in the sibling directory of this module are my backgrounds, in this instance I am trying to use `"fox_in_snow.png"`, (which is the 0th index of `client.commands.profile.backgrounds`) if, the user has set `usedBackground` (Which they do in this case), if they do not it will run the code without the image, which works fine. I am also sending this all to a buffer which I am 

sending via Discord.js, which works fine.

## My Environment

* node-canvas version 2.0.0-alpha.6
* Environment node 6.11.2 on Windows 10, version 1709

zbjornson:
* Provide coded errors (e.g. ENOENT, EACCESS with associated syscall name and path where available) instead of generic "error while reading from input stream".
* Capture libjpeg errors instead of printing to stderr/stdout

Closes #1048 (most likely cause)
Fixes #283
Fixes #486


test/image.test.js
```diff
@@ -83,6 +83,24 @@ describe('Image', function () {
     return assertRejects(loadImage(`${png_clock}fail`), Error)
   })
 
+  it('returns a nice, coded error for fopen failures', function (done) {
+    const img = new Image()
+    img.onerror = err => {
+      assert.equal(err.code, 'ENOENT')
```
LinusU: maybe assert `.syscall` and `.path` here as well?

zbjornson: 👍 added

PR #1209, opened 2018-07-31
support reading data URL on Canvas.Image#src.

nishidayuya:
This pull-request makes we can read data URL on `Canvas.Image#src`.

This pull-request doesn't have idealistic solution. I think idealistic solution is: `Canvas.Image#src` returns data URL by internal Cairo objects. But that solution has several flags and complex implementation.

So, this pull-request uses simply and pragmatic solution (save original data).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -107,6 +107,7 @@ canvas.createJPEGStream() // new
    `canvas.createPNGStream()`
  * Support for `canvas.toDataURI("image/jpeg")` (sync)
  * Support for `img.src = <url>` to match browsers
+ * Support reading data URL on `img.src`
 
 1.6.x (unreleased)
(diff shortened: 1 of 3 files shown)


zbjornson:
Makes sense. Kinda sad to have to retain the memory for the image source, but I guess there's no other way to handle this aside from re-generating the data URL in the `src` getter.

Might want to do this from C++ so that there's not another visible property on `Image` instances ... or at least prefix the new property name with an underscore to indicate that it's private.

nishidayuya:
Thank you for your comment!

I see. I didn't check private property rule. I'm sorry... 

I added `_` prefix to new private property which is like `DOMMatrix#_is2D`. Please check it.

nishidayuya:
:tada:

PR #1223, opened 2018-08-18
Support https urls

lostfictions:
Hey there, seems like http support was added for `Image.src`, but it throws an error when providing an https URL. This PR should resolve that.

I would've updated/changed tests, but it doesn't look like there are any for URL `src` values.

### DIFF
--- a/lib/image.js
+++ b/lib/image.js
@@ -36,10 +36,18 @@ Object.defineProperty(Image.prototype, 'src', {
         this.source = Buffer.from(content, isBase64 ? 'base64' : 'utf8')
         this._originalSource = val
       } else if (/^\s*https?:\/\//.test(val)) { // remote URL
+        const onerror = err => {
+          if (typeof this.onerror === 'function') {
+            this.onerror(err)
+          } else {
+            throw err
+          }
+        }
+
         const type = /^\s*https:\/\//.test(val) ? https : http
         type.get(val, res => {
           if (res.statusCode !== 200) {
-            return this.onHttpError(new Error(`Server responded with ${res.statusCode}`))
+            return onerror(new Error(`Server responded with ${res.statusCode}`))
           }
           const buffers = []
           res.on('data', buffer => buffers.push(buffer))
@@ -47,7 +55,7 @@ Object.defineProperty(Image.prototype, 'src', {
             this.source = Buffer.concat(buffers)
             this._originalSource = undefined
           })
-        }).on('error', this.onHttpError.bind(this))
+        }).on('error', onerror)
       } else { // local file path assumed
         this.source = val
         this._originalSource = undefined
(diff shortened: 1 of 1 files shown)


lib/image.js
```diff
@@ -62,7 +68,13 @@ Object.defineProperty(Image.prototype, 'src', {
       this._originalSource = undefined
     }
   },
-
+  onHttpError: err => {
```
Hakerh400: `onHttpError` should not be defined here. This object is a temporary object that is passed to `Object.defineProperty`. You can use something like

```js
Image.prototype.onHttpError = function(err){
  if (typeof this.onerror === 'function') {
    this.onerror(err)
  } else {
    throw err
  }
};
```

Also, it should not be an arrow function if you want to access `this`

lib/image.js
```diff
@@ -34,25 +35,30 @@ Object.defineProperty(Image.prototype, 'src', {
...
+            return this.onHttpError(new Error(`Server responded with ${res.statusCode}`))
           }
-        }
+          const buffers = []
+          res.on('data', buffer => buffers.push(buffer))
+          res.on('end', () => {
+            this.source = Buffer.concat(buffers)
+            this._originalSource = undefined
+          })
+        }).on('error', this.onHttpError)
```
Hakerh400: Should be `this.onHttpError.bind(this)`

lib/image.js
```diff
@@ -34,25 +35,30 @@ Object.defineProperty(Image.prototype, 'src', {
         const content = val.slice(commaI + 1)
         this.source = Buffer.from(content, isBase64 ? 'base64' : 'utf8')
         this._originalSource = val
-      } else if (/^\s*http/.test(val)) { // remote URL
-        const onerror = err => {
-          if (typeof this.onerror === 'function') {
-            this.onerror(err)
-          } else {
-            throw err
+      } else if (/^\s*https:\/\//.test(val)) { // https URL
```
Hakerh400: Is duplicating code necessary? Can it be refactored to something like

```js
} else if (/^\s*https?:\/\//.test(val)) { // URL
  const type = /^\s*https:\/\//.test(val) ? https : http;
  type.get(val, res => {
    if (res.statusCode !== 200) {
      return this.onHttpError(new Error(`Server responded with ${res.statusCode}`))
    }
    const buffers = []
    res.on('data', buffer => buffers.push(buffer))
    res.on('end', () => {
      this.source = Buffer.concat(buffers)
      this._originalSource = undefined
    })
  }).on('error', this.onHttpError.bind(this))
}
```

lostfictions:
Done!

lib/image.js
```diff
@@ -85,3 +79,11 @@ Image.prototype.inspect = function(){
     + (this.complete ? ' complete' : '')
     + ']';
 };
+
+Image.prototype.onHttpError = function(err) {
```
zbjornson: I don't think there should be a new prototype method on Image. We want to stick as closely to the HTML Image definition as possible. Could you restore the `onerror` bit to how it was so that it remains hidden please?

lostfictions: Done

LinusU:
Neat! Thanks everyone ❤️

Issue #1167, opened 2018-05-16
Shadow applied to cropped image is wrong


jfizz:
I am having some issues when applying a shadow to a cropped image. The shadow is applied correctly when running similar code inside a browser.
```
const fs = require('fs');
const {createCanvas, loadImage} = require('canvas');
const canvas = createCanvas(1200, 500);
const ctx = canvas.getContext('2d');

loadImage('lime-cat.jpg').then((image) => {
  ctx.shadowColor = '#000';
  ctx.shadowBlur = 20;
  ctx.shadowOffsetX = 4;
  ctx.shadowOffsetY = 4;

  // no crop
  // elementToDraw, cropX, cropY, width, height, x, y, width, height
  ctx.drawImage(image, 0, 0, 480, 408, 10, 46, 480, 408);

  // cropped
  ctx.drawImage(image, 100, 100, 380, 308, 760, 126, 380, 308);

  fs.writeFileSync('out.png', canvas.toBuffer());
});
```
![out](https://user-images.githubusercontent.com/6319548/40124865-52a274a2-58de-11e8-9c95-cb174f8eb49f.png)

## Your Environment
* Version of node-canvas (e.g. 1.4.0): canvas@2.0.0-alpha.12
* Environment (e.g. node 4.2.0 on Mac OS X 10.8): node 8.9.4 on Mac OS X 10.13.2

PR #1228, opened 2018-08-25
Add visual comparision to tests

asturur:
This modify the html of the test script to add a visual compare of the canvases.
Makes easier to find out differences.

example:
image

Floating point coordinates produce some difference in aliasing:
image

I hope you find it usefull, i do.

Also added 3 tests, of which 2 failing, to show them fixed in the drawImagePr.

image

The failing tests are relative to the issue #1167


asturur:
What about ( in a future pr ) make those tests run under travis and pass or fail depending on number of pixel that are different?

asturur:
@LinusU @zbjornson this PR is ready.

asturur:
Added 3 tests for the smoothingImageEnabled

image

asturur:
@LinusU  if this thing here is interesting, i can update the test description here directly.

asturur:
@LinusU @zbjornson  updated test title

asturur:
Any follow up on this?

asturur:
I think this is a cool feature an can help looking faster at tests.
Do you have any though on this?

test/public/pixelmatch.js
```diff
@@ -0,0 +1,157 @@
+'use strict'
+
+function pixelmatch (img1, img2, output, width, height, options) {
```
zbjornson: Is this [mapbox's](https://github.com/mapbox/pixelmatch/blob/master/index.js)? If so, maybe add it as a devDependency instead (for attribution/licensing and less code for us to maintain)?

asturur: is called pixelmatch, the reason i copy pasted it is unclear now. I think i did not  manage to use on the webserver as a npm module? i ll try again

test/server.js
```diff
@@ -32,6 +32,10 @@ app.get('/', function (req, res) {
   res.sendFile(path.join(__dirname, 'public', 'app.html'))
 })
 
+app.get('/pixelmatch.js', function (req, res) {
+  res.sendFile(path.join(__dirname, '../node_modules/pixelmatch/', 'index.js'))
+})
+
```
asturur: this is the simplest way i found to serve it from the node_modules folder. i think is ok.

asturur:
@zbjornson @LinusU  added pixelmatch as a package. This should be ready

asturur:
i ping again here. Unsure if something is missing or just we do not like it.

zbjornson:
Sorry for not responding. I don't have a problem adding it, but it would be good to adjust the test runner so that all of the tests don't have to call `done()` when they've completed.

asturur:
there was a specific reason why i did that, rather than dividing from sync and async, but now i have to remember it. i ll let you know.

test/public/tests.js
```diff
@@ -12,19 +12,21 @@ if (typeof module !== 'undefined' && module.exports) {
   imageSrc = function (filename) { return filename }
 }
 
-tests['clearRect()'] = function (ctx) {
+tests['clearRect()'] = function (ctx, done) {
```
LinusU: Instead of having every test accepting a `done` parameter you could simply test the arity with `const hasCallback = tests['clearRect()'].length === 2`, that way you don't have to modify all the existing tests, and that is in line with how mocha does it.

asturur:
i m fine to change to that.
I m also fine to have all tests look the same with a callback.

I ll do as soon as i have time.

PR #1233, opened 2018-08-28
Make imagesmoothing enabled behave more like browser

asturur:
Image smoothing enabled currently is bound to patternQuality best/fast.

ImageSmoothing set to false should give the effect of nearest neighbour.
Also pattern quality setting should be independent of the imageSmoothing one, they should not be coupled, imageSmoothing being a standard, while patternQuality an addition.

### DIFF
--- a/lib/context2d.js
+++ b/lib/context2d.js
@@ -25,27 +25,6 @@ const DOMMatrix = require('./DOMMatrix').DOMMatrix
 var baselines = ['alphabetic', 'top', 'bottom', 'middle', 'ideographic', 'hanging'];
 
 /**
- * Enable or disable image smoothing.
- *
- * @api public
- */
-
-Context2d.prototype.__defineSetter__('imageSmoothingEnabled', function(val){
-  this._imageSmoothing = !! val;
-  this.patternQuality = val ? 'best' : 'fast';
-});
-
-/**
- * Get image smoothing value.
- *
- * @api public
- */
-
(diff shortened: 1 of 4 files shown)


zbjornson:
@asturur thanks for all the PRs! I'm buried at work right now so will be a little bit before I can review them.

At quick glance this one looks great!

LinusU:
Do you think we should remove `patternQuality` in favour of this? As far as I can tell there is no `patternQuality` in the Canvas spec...

asturur:
there is no patternQuality, but as long as the extra feature does not cross with the standard one, i guess is ok.

Please be aware i prepares tests but did.not try this change enough yet.

asturur:
@zbjornson and @linusU as a side note i really need the drawImage pr asap ( i perfectly understand this is not a concern no worries ) so if you have some time, start from that. I also need to backport it to 1.x

asturur:
So to be really accurate, browser canvas do a bad job in scaling down, always, also with imageSmoothingEnabled set to true. I do think canvas always play between nearest ( false ) and fast ( true ), and when scaling down is always nearest.

I do not think improving the quality of the output is a problem since devs can still tweak it with the non standard patternQuality parameter.
The important is that if you set it to false, and you want to have a small sprite rendered in high rest, with blocky effect, you are able to do so.

asturur:
i think i'm done

PR #1237, opened 2018-09-03
This is just some code reuse - for comments

asturur:
Code from fillText and strokeText is very similar,
I put it in a function with a boolean switch to either stroke or fill.

I also removed the scale and counter scale with a save, scale, restore.

I do think that while C double accuracy does not suffer from same JS float accuracy, scaling and counter scaling can eventually accumulate small math error?

The same can be done for fillRect/strokeRect/clearRect that are basically the same.
I understand this is a matter of code style / opinions and not real advantages/codequality

Let me know what you think

### DIFF
--- a/src/CanvasRenderingContext2d.cc
+++ b/src/CanvasRenderingContext2d.cc
@@ -2070,11 +2070,8 @@ get_text_scale(Context2d *context, char *str, double maxWidth) {
   }
 }
 
-/*
- * Fill text at (x, y).
- */
-
-NAN_METHOD(Context2d::FillText) {
+void
+paintText(const Nan::FunctionCallbackInfo<Value> &info, bool stroke) {
   int argsNum = info.Length() >= 4 ? 3 : 2;
   double args[3];
   if(!checkArgs(info, args, argsNum, 1))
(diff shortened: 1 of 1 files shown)


asturur:
do you think we should reuse the fillRect, strokeRect, clearRect code?

zbjornson:
> I do think that while C double accuracy does not suffer from same JS float accuracy, scaling and counter scaling can eventually accumulate small math error?

JS uses IEEE-754 doubles and has the same accuracy as C for these math basic math operations (divide, multiply), but yes, floating-point error happens in both. Good idea to save/restore.

zbjornson:
And yes, I'm all for more code de-duplication where you see opportunity! As long as it's easy to read. :)

asturur:
So about easy to read, after a bit i came out with this ( because i m slow at C )

image

Instead of adding a bool and and IF at runtime, i can use a pointer to the stroke and fill method.

Do you think this approach is worse/better?

zbjornson:
Also a good option! No preference from me.

If you use function pointers, you could use C++11 syntax:

```cpp
#include <functional>
void paintText(..., std::function<void(Context2d&, bool)> operation) {
    ...
    operation(*context, false);
```

(I think that's the cleanest usage of `std::function` ... sort of debatable if it's cleaner than C-style function pointers.)

asturur:
i like more your example, way more readable and reminds me of the bind syntax in JS when used. I guess later i ll refactor the fillRect strokeRect and move the text to this example

PR #1239, opened 2018-09-03
update the layout before calculating the extent fix for #1088

### LINKED ISSUE #1088: text maxWidth does not work for the first text render in the image & is inconsistent on second render
## Issue
When using the maxWidth argument of either fillText or strokeText, if the text being width limited is the first text on the image, maxWidth will not work. If any text is rendered before, the subsequent text render will be resized but will be inconsistent and shorter than it should unless it has the same text. All text renders after the first and second will work properly.

This issue is better explained with code and images:

## Steps to Reproduce
```js

ctx.font = '18px Helvetica'
ctx.fillText('I should be resized', 0, 20, 90)
ctx.fillText('I should be resized', 0, 40, 90)
ctx.fillText('I should be resized', 0, 70, 90)
ctx.fillText('I should be resized', 0, 100, 90)
ctx.fillText('I should be resized', 0, 130, 90)
ctx.fillText('I should be resized', 0, 170, 90)
```

The code above generates this image:
image

Now we change the text for the first text render:

```js

ctx.font = '18px Helvetica'
ctx.fillText('I should be resized a', 0, 20, 90)
ctx.fillText('I should be resized', 0, 40, 90)
ctx.fillText('I should be resized', 0, 70, 90)
ctx.fillText('I should be resized', 0, 100, 90)
ctx.fillText('I should be resized', 0, 130, 90)
ctx.fillText('I should be resized', 0, 170, 90)
```

The result is:
image

Notice that the second text is smaller than the others.

The behaviour above is consistent even if we remove the maxWidth property from the first fillText call.

image

The image above is the same code, just without maxWidth on the first fillText call and a longer text on it. It seems that the longer the first text, the smaller the second one will be.

strokeText is also affected in the same way.

## Environment
* Version of node-canvas: 2.0.0-alpha.9
* Environment: OS X 10.13.3

asturur:
In get_text_scale we are using pango_layout_get_pixel_extent without first updating the text for the layout.
So the bug in #1088, since measurement happen on last text used in measureText or setTextPath.

Added tests for regressiong ( visual )
image

Test before fix
image

close #1088

### DIFF
--- a/src/CanvasRenderingContext2d.cc
+++ b/src/CanvasRenderingContext2d.cc
@@ -2059,7 +2059,7 @@ NAN_METHOD(Context2d::Stroke) {
 
 double
 get_text_scale(Context2d *context, char *str, double maxWidth) {
-  PangoLayout *layout = context->layout()
+  PangoLayout *layout = context->layout();
   PangoRectangle logical_rect;
   pango_layout_set_text(layout, str, -1);
   pango_cairo_update_layout(context->context(), layout);
--- a/src/CanvasRenderingContext2d.cc
+++ b/src/CanvasRenderingContext2d.cc
@@ -2058,11 +2058,9 @@ NAN_METHOD(Context2d::Stroke) {
  */
 
 double
-get_text_scale(Context2d *context, char *str, double maxWidth) {
-  PangoLayout *layout = context->layout();
+get_text_scale(PangoLayout *layout, char *str, double maxWidth) {
+
   PangoRectangle logical_rect;
-  pango_layout_set_text(layout, str, -1);
-  pango_cairo_update_layout(context->context(), layout);
   pango_layout_get_pixel_extents(layout, NULL, &logical_rect);
 
   if (logical_rect.width > maxWidth) {
@@ -2085,8 +2083,13 @@ paintText(const Nan::FunctionCallbackInfo<Value> &info, bool stroke) {
   double scaled_by = 1;
 
   Context2d *context = Nan::ObjectWrap::Unwrap<Context2d>(info.This());
+  PangoLayout *layout = context->layout();
+
+  pango_layout_set_text(layout, str, -1);
+  pango_cairo_update_layout(context->context(), layout);
+
   if (argsNum == 3) {
-    scaled_by = get_text_scale(context, *str, args[2]);
+    scaled_by = get_text_scale(layout, *str, args[2]);
     cairo_save(context->context());
     cairo_scale(context->context(), scaled_by, 1);
   }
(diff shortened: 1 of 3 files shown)


asturur:
Question:
I have no idea of the cost of the 2 functions i added.
With the current code setup, the couple:

```c++
  pango_layout_set_text(layout, str, -1);
  pango_cairo_update_layout(context->context(), layout);
```

Could be added in `paintText`

and removed from both `get_text_scale` and `Context2d::setTextPath`. 
I have no idea if the method setTextPath, now used in just one place, is meant to be reused in more places and so needs to be independent.

chearon:
Yeah, ideally those functions should only be called once... We could just remove them from `Context2d::SetTextPath`. `SetTextPath` is supposed to add the paths to the canvas context, but doesn't necessarily need to do the Pango layout. Don't have any better ideas than that at the moment.

asturur:
Wanted just a confirm, i ll do that.

asturur:
setTextpath does need to know the text width in order to do some alignment.

chearon:
Yeah, but you were saying put it at the top of `paintText`? I think that works

asturur:
Yes so that i can pass the width to the eventual scale factor and the setTextPath.
it does feel weird that setTextPath has also to align, but i do not want to touch too much.

asturur:
I forgot i had this PR open :(
I ll follow up tomorrow, any thoughts apart what @chearon commented?

zbjornson:
Looks good to me once @chearon's comments are addressed. 👍

Lange:
Some functionality I'm working on for a charity fundraiser broadcast relies on this fix -- would love to see it finished and merged soon! ❤️

asturur:
Yeah i should do that

asturur:
argh i removed it somewhere else and forgot here.

asturur:
Updated here from the interface. Did not test.

asturur:
Should be fixed now, i was committing yesterday then i was called to board and had to left where i was.

LinusU:
@Lange published [2.0.0-alpha.15](https://www.npmjs.com/package/canvas/v/2.0.0-alpha.15) just for you 😉

Lange:
@LinusU Big ❤️! Thanks so much, I'll implement this in my code later today and report back with any issues.

Lange:
While the text is now indeed scaled to the correct size, it seems to be drawn at the wrong coordinates when `textAlign` is `center`. I'm not sure if I have time to make a minimal repro but I can certainly try if it would be helpful.

zbjornson:
@Lange thanks for finding, #1253. Not a regression thankfully, and looks like a straightforward logic error.

Lange:
Got it, I'll follow #1253

PR #1246, opened 2018-09-09
Support reading CMYK, YCCK JPEGs

### LINKED ISSUE #425: CMYK JPEG images are not properly painted
So here is example of CMYK Jpeg that we try just to draw on canvas in nodejs.
logo-neu-fl-01-unsharp

And here is an result of it.
screenshot 2014-05-21 13 04 50

Code that I am using:

```
        this.context.drawImage(this.imageObj, imageX, imageY, imageWidth, imageHeight);
```

I would suggest to return exception that JPEG CMYK is not supported.

zbjornson:
libjpeg automatically converts YCCK to CMYK, so we get that for free by adding CMYK support. Also adds a test for grayscale and does some cleanup.

Fixes #425


src/Image.cc
```diff
@@ -200,8 +200,10 @@ Image::clearData() {
     _surface = NULL;
   }
 
-  free(_data);
-  _data = NULL;
+  if (_data) {
+    delete[] _data;
+    _data = nullptr;
+  }
```
LinusU: Nice 🙌 

btw. I think it's safe to run `delete` on a `nullptr`, so I don't think you need the `if`? 🤔

test/public/tests.js
```diff
@@ -2253,19 +2240,15 @@ tests['image sampling (#1084)'] = function (ctx, done) {
...
   img2.onload = () => {
     loaded2 = true
     ctx.drawImage(img2, 182 - 100, -203, 352, 352)
     if (loaded1) done()
   }
 
-  img2.onerror = function () {
-    done(new Error('Failed to load image'))
-  }
+  img2.onerror = done
```
LinusU: Neat!

src/Image.cc
zbjornson: Yeah, it's defined behavior. Removed the guard.

marcj:
Is there a way to fix color space (or whatever gives too much colors)? I tried this PR and got following:

Browser (newest Chrome)

screen shot 2018-10-07 at 13 27 38

Node-Canvas:

screen shot 2018-10-07 at 13 27 49

Actual image (25MB): https://drive.google.com/open?id=1LttH_DNl_Y5Dap__hQ88MhHqqb2YgJfe

As you can see, the colors of the right image are a bit too colorfull. Any idea how I can debug?

zbjornson:
@marcj the EXIF data for that JPEG says it doesn't have an embedded color space, so I think that can be ruled out. Hard to say what layer that difference comes from... depending on how you got the images you embedded in your post (were they screenshots?) it could even be Chrome's display rendering vs. the MacOS Preview app or something.

marcj:
Its both a screenshot. The (wrong) image from node-canvas looks across multiple viewer the same, so I assume it does not read the CMYK file correctly.

I use now sharp (https://github.com/lovell/sharp) to convert the CMYK image to RGB first before using it in node-canvas, which looks very good. However, I'd not like to do that as it's rather slow.

PR #1247, opened 2018-09-11
Hide Image.prototype.source

Hakerh400:
- [x] Updated changelog
- [x] Added test

### DIFF
--- a/src/Image.cc
+++ b/src/Image.cc
@@ -56,12 +56,14 @@ Image::Initialize(Nan::ADDON_REGISTER_FUNCTION_ARGS_TYPE target) {
 
   // Prototype
   Local<ObjectTemplate> proto = ctor->PrototypeTemplate();
-  SetProtoAccessor(proto, Nan::New("source").ToLocalChecked(), GetSource, SetSource, ctor);
   SetProtoAccessor(proto, Nan::New("complete").ToLocalChecked(), GetComplete, NULL, ctor);
   SetProtoAccessor(proto, Nan::New("width").ToLocalChecked(), GetWidth, SetWidth, ctor);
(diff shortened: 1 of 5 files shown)


zbjornson:
Sorry for the delay, will try to review tomorrow!

src/Image.cc
```diff
@@ -182,7 +184,7 @@ NAN_SETTER(Image::SetHeight) {
  * Get src path.
  */
 
-NAN_GETTER(Image::GetSource) {
+void Image::GetSource(const Nan::FunctionCallbackInfo<v8::Value> &info) {
```
zbjornson: I think this and line 227 should/could be `NAN_METHOD(Image::GetSource)` ?

Hakerh400: Rebased properly?

Hakerh400:
@KenanSulayman

What do you mean? Which tests are broken by this PR? What you posted breaks all tests, since there is no more accessor property `source` on the image prototype. Are you sure you merged/fetched the updated files properly? Also, make sure to rebuild with node-gyp, so that C++ patches take effect.

Hakerh400:
Unable to reproduce.

```console
D:\node_modules\canvas\examples> node resize.js
Resized and saved in 26ms
```

Can you please open a new issue if you believe that something is broken.

KenanSulayman:
This is not meant to offend, but just because it works for you - and on your Windows installation - that does not mean it works for others.

Here's a recording of my bash session: https://asciinema.org/a/2VXF7vqdHS0Wk8FP30zfAvc7I.

KenanSulayman:
Here's another assumption:  does the git build fetch the remote build no matter what?

KenanSulayman:
Yes, that's the case. I can see the cause of this issue then.

zbjornson:
> I can see the cause of this issue then.

@KenanSulayman are you all set here, or do you think something is broken?

Issue #1249, opened 2018-09-19
CAIRO_EXTEND_REFLECT when extending image


ddcovery:
In version 1.6.12, method drawImage extends the imge using CAIRO_EXTEND REFLECT:

> if (needsExtraSurface) {
>    ...
>    cairo_pattern_set_extend(cairo_get_source(ctxTemp), CAIRO_EXTEND_REFLECT);
>    cairo_paint_with_alpha(ctxTemp, 1);
>    surface = surfTemp;
>  }

Finally you perform the drawing using the default cairo_pattern_set_extend

> cairo_pattern_set_extend(cairo_get_source(ctx), CAIRO_EXTEND_NONE);

This "NONE" is doing nothing, because extend ALLWAYS uses the surfTemp version that contains a CAIRO_EXTEND_REFLECT and final result is a REFLECT extend.

This changes de default drawImage behaviour that is CAIRO_EXTEND_NONE (similar to Firefox/Chrome/...).

Actually, I solve this problem cleaning the reflected areas with extra code... problem is when the canvas where I paint the image contains useful contents  this extra content is overpainted with the reflect.

ddcovery:
Ok, sorry.

I have 2 images:

* images/small.jpg, 801x1200
  ![small](https://user-images.githubusercontent.com/6564217/45802984-9daa7b00-bcb8-11e8-98ec-c644e0aff2c9.jpg)

* image/big.jpg 2400x1602
  ![big](https://user-images.githubusercontent.com/6564217/45802991-a13e0200-bcb8-11e8-92ed-dc6d527a8c78.jpg)

This code loads the two images and draws the small one over the big one forcing extend:

```javascript
const Canvas = require("canvas");
const Image = Canvas.Image;
const fs = require("fs");
const Q = require("bluebird");

// Draw cat with lime helmet
loadImage(__dirname + '/images/small.jpg').then(smallImg => loadImage('images/big.jpg').then(bigImg => {

  console.log("smallImg:", smallImg.width, smallImg.height); // 801 1200
  console.log("bigImg:", bigImg.width, bigImg.height); // 2400 1602
  
  var canvas = new Canvas(bigImg.width, bigImg.height);
  const ctx = canvas.getContext('2d');
  ctx.drawImage(bigImg, 0, 0);
  //return saveCanvas(canvas, __dirname + "/1.png");
[...]

PR #1255, opened 2018-09-25
Initial work to fix detail loss with additional canvas needed on drawImage

asturur:
## The problem

In the case when we specify a small destination area, but with a big scale of the context, we need to take in account that if generating an addtional canvas.

This PR introduces a function to extrapolate the scale value of the current transform ( together with the other properties ) And we will increase the area of the temporary surface in order to account the scale so that details are preserved.

After various iterations this code looks. good. I understand it adds complexity.

In case we need an additional canvas we want to know how bit it will be before creating it, so that we can create one that is as small as required to get the right amount of details.

In case the context has has a scaling of 3, and the drawImage is something like:

```
drawImage(source, 0, 0, 100, 100, 0, 0, 30, 30)
```

The draw image will need to create a 90x90 surface where to draw the 100x100 pixel from the source.
(since 30x3 = 90)

Also when drawImage gets used like:

```
drawImage(source, -10, -10, 100, 100, 0, 0, 30, 30)
```

the 10pixel from sx and sy arguments are blank, we do not need to have space in the extra canvas to allocate them.

There is an added level of complexity that is still missing:
It would be nice to understand when we are about to draw an image at more than 100% size and avoid allocating space in the temporary surface.


asturur:
This raw code fixed the bug and does not seem introducing more, but is not the final form of it and not want to commit. I opened the PR now so that you can start make questions or rise concerns or whatever you want to do , in advance

asturur:
@LinusU @zbjornson  i cannot comment and i m not able anymore to show here the code i pushed in a the branch. If i create a new branch i cannot open a new PR. Is the repo locked or something is happening with my account?

asturur:
@LinusU @zbjornson  i understand this code is a bit unclear what is doing.
I would like to work together on this so that:
- the code is readabale
- drawImage can continue to work on the 9 args call with global composite operation
- drawImage creates small temporary surfaces ( no more than the pixels needed to draw )

I understand not everything can be prevented, an drawCall outside the canvas width/height will not draw anything but if needed will still create a tempSurface.
I do not think we should write code to prevent that too, but if an image is 4000x4000 pixels, i would like to be able to create a temp surface that is not bigger than the image itself.

asturur:
this fixes also: #1335

@zbjornson  @bennlich

image

asturur:
@zbjornson @LinusU  i understand this PR is not of immediate readability and needs time to understand what is it doing.

Please find some time to read and make me all the questions, so i can move this forward and think about a couple of other things that deserve their own pr.
(further limit the extra surface, re-render SVG on ctx scaling so that it  scales correctly )

I added a new test, and also i see improvements in other tests:

this test takes small slices of images and draw them super big ( like 90k x 90k ), this to test the extra surface reduction works ( this tests crashes current master )

image

this test scale up an svg and show some improvements compared to master:
CURRENT:
image

MASTER:
image

the super blurred effect is actually a regression introduced with the code from me, those are extreme edge cases usage of canvas, but those use cases happens with high level libraries sometimes ( fabricJS for example triggers high level of scaling )

src/CanvasRenderingContext2d.cc
```diff
@@ -1269,41 +1283,74 @@ NAN_METHOD(Context2d::DrawImage) {
...
-  float fx = (float) dw / sw;
-  float fy = (float) dh / sh;
+  cairo_matrix_t matrix;
+  double transforms[6];
+  cairo_get_matrix(context->context(), &matrix);
+  decompose_matrix(matrix, transforms);
+  // extract the scale value from the current transform so that we know how many pixels we
+  // need for our extra canvas in the drawImage operation.
+  double current_scale_y = transforms[1];
+  double current_scale_x = transforms[2];
```
zbjornson: Are these flipped? Your code comment above says `1: scaleX, 2: scaleY`.

Likewise I think one of the comments on line 1296 or 1297 is supposed to say `Y`.

---

I like where this is headed (a systematic approach as opposed to special handling of individual cases) 👍 . Will give a closer read-through once the above are clarified.

asturur: yeah i flipped those, trying just with standard scaling i did not notice.

asturur:
@zbjornson i updated the accidentally swapped values. Thanks for spotting that!

asturur:
any update?

src/CanvasRenderingContext2d.cc
```diff
@@ -1266,41 +1280,74 @@ NAN_METHOD(Context2d::DrawImage) {
...
-  float fx = (float) dw / sw;
-  float fy = (float) dh / sh;
+  cairo_matrix_t matrix;
+  double transforms[6];
+  cairo_get_matrix(context->context(), &matrix);
+  decompose_matrix(matrix, transforms);
+  // extract the scale value from the current transform so that we know how many pixels we
+  // need for our extra canvas in the drawImage operation.
+  double current_scale_x = transforms[1];
+  double current_scale_y = transforms[2];
```
zbjornson: I think these values should just be `matrix.xx` and `matrix.yy` instead of decomposing the matrix. Try this test case -- it's broken with `transforms[1|2]` but works with `matrix.xx|yy`

```js
tests['transformed drawimage'] = function (ctx) {
  ctx.fillStyle = 'white'
  ctx.fillRect(0, 0, 200, 200)
  ctx.fillStyle = 'black'
  ctx.fillRect(5, 5, 50, 50)
  ctx.transform(1.2, 1, 1.8, 1.3, 0, 0);
  ctx.drawImage(ctx.canvas, 0, 0)
}
```

(The sign of scale_y from the decomposed matrix is negative when it should be positive.)

src/CanvasRenderingContext2d.cc
```diff
@@ -1266,41 +1280,74 @@ NAN_METHOD(Context2d::DrawImage) {
...
+  float extra_dx = 0;
+  float extra_dy = 0;
+  float fx = (float) dw / sw * current_scale_x; // transforms[1] is scale on X
+  float fy = (float) dh / sh * current_scale_y; // transforms[2] is scale on X
```
zbjornson: You don't need this cast; it happens implicitly.

src/CanvasRenderingContext2d.cc
```diff
@@ -1266,41 +1280,74 @@ NAN_METHOD(Context2d::DrawImage) {
...
-  float fy = (float) dh / sh;
+  cairo_matrix_t matrix;
+  double transforms[6];
+  cairo_get_matrix(context->context(), &matrix);
+  decompose_matrix(matrix, transforms);
+  // extract the scale value from the current transform so that we know how many pixels we
+  // need for our extra canvas in the drawImage operation.
+  double current_scale_x = transforms[1];
+  double current_scale_y = transforms[2];
+  float extra_dx = 0;
```
zbjornson: This function (before your PR even) is a weird mix of doubles and floats. They should all be doubles to minimize implicit conversions: the arguments (`info` array) are doubles, and the cairo APIs that take floating-point numbers take doubles. Can you please change the new variables you're adding in this PR to `double`, along with the variables declared on line 1208 while you're here?

src/CanvasRenderingContext2d.cc
```diff
@@ -1334,8 +1381,18 @@ NAN_METHOD(Context2d::DrawImage) {
...
+  float scaled_dx = dx;
+  float scaled_dy = dy;
+
+  if (needsExtraSurface && (current_scale_x != 1 || current_scale_y != 1)) {
+    // in this case our surface contains already current_scale_x, we need to scale back
+    cairo_scale(ctx, 1 / current_scale_x, 1 / current_scale_y);
+    scaled_dx *= current_scale_x;
+    scaled_dy *= current_scale_y;
+  }
+  // cairo_translate(ctx, extra_dx, extra_dy);
```
zbjornson: Does this need to be uncommented?

src/CanvasRenderingContext2d.cc
```diff
@@ -1266,41 +1280,74 @@ NAN_METHOD(Context2d::DrawImage) {
...
+  double current_scale_y = transforms[2];
+  float extra_dx = 0;
+  float extra_dy = 0;
+  float fx = (float) dw / sw * current_scale_x; // transforms[1] is scale on X
+  float fy = (float) dh / sh * current_scale_y; // transforms[2] is scale on X
   bool needScale = dw != sw || dh != sh;
   bool needCut = sw != source_w || sh != source_h || sx < 0 || sy < 0;
-  bool needCairoClip = sx < 0 || sy < 0 || sw > source_w || sh > source_h;
-
+  bool needAdditionalCut = sx < 0 || sy < 0 || sw > source_w || sh > source_h;
```
zbjornson: I don't quite understand `needCut` vs. `needAdditionalCut`. Both of them test `sx < 0 || sy < 0`. `sw != source_w` is also true for all cases where `sw > source_w` (ignoring `NaN`, which shouldn't be encountered here).

src/CanvasRenderingContext2d.cc
asturur: if matrix.xx is always the matrix[0] position, the number that we generally associate with scaleX, that is not enough. When you have a rotation in that position there are mixed values of scaling and angle of rotation, as well as in matrix[3] or matrix.yy.

Now the negative sign, that is a point and should be fixed, let me see if i can have some code that works with any sign ( the scale is the abs value, while the image needs a flip )

src/CanvasRenderingContext2d.cc
asturur: sure, i can see to clean everything in double.

src/CanvasRenderingContext2d.cc
asturur: no, deleted.

src/CanvasRenderingContext2d.cc
asturur: no with the code where it is now, those booleans can become 1. Probably before there was extra code paths to follow in a case or the other.

src/CanvasRenderingContext2d.cc
zbjornson: Do you have an example test case that fails using just `xx` and `yy`? My understanding of this code (which may be wrong) is that both the canvas and the image are rotated together, so xx and yy are all you need. I tried quite a few random matrices.

Also I was unclear on the "negative when it should be positive" bit -- `yy` was like ~2.0 and `transforms[2]` was like -0.6 -- not just an `abs()` would have fixed it.

asturur: well i may have done something wrong. i ll triple check. i use this decomposing with success elsewhere in javascript, not in c.

src/CanvasRenderingContext2d.cc
bennlich: @zbjornson @asturur  it looks to me like `needAdditionalCut` is a renaming of `needCairoClip`, which was introduced to fix #1249 in this related PR: #1251.

I think both this current PR and #1251 are fixes for (among other things) issues that were introduced with the use of `CAIRO_EXTEND_REFLECT`.

Please excuse comment from left-field here :) I hope it's helpful history.

asturur: Yes i introduced both, they can be unified. Before the code had different srategies depending on what was the reason of the cut.

asturur:
add busy period, i ll try to fix what is left

asturur:
@zbjornson i fixed your comment using abs, and i'm trying to demonstrate the reason behind it with this fiddle.

https://jsfiddle.net/gkwdo1ez/3/

So in this fiddle you see an image painted on a canvas using transform with the image you suggested.
Then the same image is painted on another canvas decomposing the matrix and then executing the oprations in this order:

ctx.translate, ctx.rotate, ctx.scale, ctx.skewX.

Then pixels of the resulting canvas are compared for differences, and no differences are found.

While is true that matri.xx generally represent the scale, this is no more true when skewing comes in play.

This decomposition aims at finding in the whole transformation, which part is a transform that is pure scaling. In this way we know how much our temporary surface has to be big to avoid detail loss or to avoid creating a giant unnecessary surface.

this can also be verified multiplying back the matrices together still in the fiddle.

image

now the scaling factor to me are X 1.5 and Y -0.15 ( and indeed the image is flipped somehow ), i see you get different numbers, i wonder if i implemented it bad.

Let me know what you think of this explanation, meanwhile i investigate why number here are different.

asturur:
So i see this code is breaking something. I'm trying to figure out what and why.
But yes is not mergeable right now, but please go over my description and see if it makes sense for you.

asturur:
I think i understand the error.
I cannot just scale back if there is scale.
But i need to swap the ctx transform with one built with the same logic, but without the scaling part.

asturur:
Ok after spending hours on this code i m more confident is not a nest of snakes and spiders.

It works with weird skewing with or without flippings
image

Let me know, i would like to work on other node-canvas things, but i would like to park settle this change here.

Of course if it does not convince you, let me know and i ll do my best to convince you.

asturur:
@zbjornson @LinusU  ready for re-review.

zbjornson:
Thanks @asturur. I'm probably not going to have time to review closely until Friday or this weekend, but will get to it ASAP!

asturur:
any news? i would like to start working on my next pr ( svg resizing with scale ) but i do not like having 2 open pr at time.

LinusU:
Pardon the delay, merging 🚀

asturur:
thanks!!

PR #1256, opened 2018-09-25
Add raw image support

basvanmeurs:
When you need to read parsed raw ARGB image data, you need to load the image, then draw it on a canvas and convert it to a buffer:

        var img = new Canvas.Image();
        img.src = data;
        var canvas = new Canvas(img.width, img.height);
        ctx.drawImage(img, 0, 0, img.width, img.height);
        var buffer = canvas.toBuffer('raw');
This is an unnecessary double copy of the data. This pull request contains a new method: Image.getRawData, which returns the buffer directly:

        var img = new Canvas.Image();
        img.src = data;
        var buffer = img.getRawData();
This allows for a more efficient usage of Canvas.Image.

We'd appreciate it if you'd merge this into the node-canvas module!


src/Image.cc
```diff
@@ -189,6 +190,25 @@ NAN_METHOD(Image::GetSource){
...
+ * Get raw data.
+ */
+NAN_GETTER(Image::GetRawData) {
+  Image *img = Nan::ObjectWrap::Unwrap<Image>(info.This());
+  if (img->_surface) {
+    // Return raw ARGB data -- just a memcpy()
+    cairo_surface_t *surface = img->_surface;
+    cairo_surface_flush(surface);
+    const unsigned char *data = cairo_image_surface_get_data(surface);
+     unsigned int nBytes = img->width * img->height * 4;
```
zbjornson: `nBytes = cairo_image_surface_get_stride(surface) * img->height` technically. I don't think images can decode into another surface type now, but this'll be one less thing to change if they can in the future.

zbjornson (commented):
Looks good. Needs a conflict fixed again (sorry), and ideally one trivial change.

I can make the changes real quick if you want.

src/Image.cc
basvanmeurs: Great, thx for the feedback. I've changed the code and merged with master.

src/Image.cc
```diff
@@ -189,6 +190,25 @@ NAN_METHOD(Image::GetSource){
   info.GetReturnValue().Set(Nan::New<String>(img->filename ? img->filename : "").ToLocalChecked());
 }
 
+/*
+ * Get raw data.
+ */
+NAN_GETTER(Image::GetRawData) {
+  Image *img = Nan::ObjectWrap::Unwrap<Image>(info.This());
+  if (img->_surface) {
```
LinusU: Would prefer to have this as a guard instead, e.g.

```c++
if (!img->_surface) {
  info.GetReturnValue().Set(Nan::Undefined());
  return;
}

// ...
```

src/Image.cc
```diff
@@ -189,6 +190,25 @@ NAN_METHOD(Image::GetSource){
...
+    cairo_surface_t *surface = img->_surface;
+    cairo_surface_flush(surface);
+    const unsigned char *data = cairo_image_surface_get_data(surface);
+     unsigned int nBytes = img->width * img->height * 4;
+    Local<Object> buf = Nan::CopyBuffer(reinterpret_cast<const char*>(data), nBytes).ToLocalChecked();
+    info.GetReturnValue().Set(buf);
+  } else {
+    info.GetReturnValue().Set(Nan::Undefined());
+  }
+  return;
```
LinusU: No need to `return` at the end of the function ☺️

src/Image.cc
LinusU: indentation looks a bit off here 🧐

test/image.test.js
```diff
@@ -106,6 +106,14 @@ describe('Image', function () {
     })
   })
 
+  it('returns raw image data', function() {
+      var img = new Image();
+      img.src = png_clock;
+      var buf = img.rawData;
+      assert.ok(buf instanceof Buffer);
+      assert.strictEqual(409600, img.rawData.length);
+  });
```
LinusU: Please follow standard style here (2 spaces indentation, space between `function` and `()`, no semicolons)

test/image.test.js
```diff
@@ -106,6 +106,14 @@ describe('Image', function () {
     })
   })
 
+  it('returns raw image data', function() {
+      var img = new Image();
+      img.src = png_clock;
+      var buf = img.rawData;
+      assert.ok(buf instanceof Buffer);
+      assert.strictEqual(409600, img.rawData.length);
```
LinusU: Could you swap the parameters here? So that it's `assert.strictEqual(actual, expected)`, that way the error message will be correct

test/image.test.js
```diff
@@ -106,6 +106,14 @@ describe('Image', function () {
     })
   })
 
+  it('returns raw image data', function() {
+      var img = new Image();
```
LinusU: use `const` instead of `var`

test/image.test.js
```diff
@@ -106,6 +106,14 @@ describe('Image', function () {
...
+      img.src = png_clock;
+      var buf = img.rawData;
```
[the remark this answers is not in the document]
LinusU: actually, you could just do `assert.ok(img.rawData instanceof Buffer)` on the next line (or also change `img.rawData.length` to `buf.length`)

LinusU:
Personally not a fan of tacking more non-standard things on to the prototype 🤔 I do see the use case though.

Generally, what do we think about having these as standalone exported functions? Not sure that this is a good idea, but it could be something like this:

```js
const { getRawImageBuffer } = require('canvas')

// img = ...
const buffer = getRawImageBuffer(img)
``` 

Just a thought 🤔 💭

basvanmeurs:
Hi guys, can you spare some time to review and consider merge please?

src/Image.cc
zbjornson: Also don't need the `else { set undefined }` block -- that's implicit if you didn't set another return value.

zbjornson:
I don't mind if this goes on the prototype or is an accessory function. It's slightly more convenient on the prototype (one less function to import).

A few of Linus's comments are still pending though, and the last commit should be removed (CI forces a build from source).

LinusU:
> I defer to others' decisions.

I feel the same 😁 

The comments needs to be addressed before we merge though...

NightYoshi370:
I just sent a Pull Request addressing the comments

PR #1268, opened 2018-10-03
Test a different travis config

asturur:
Check if this Travis config is a bit faster

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,4 +23,3 @@ addons:
       - g++-4.9
 env:
   - CXX=g++-4.9
-sudo: false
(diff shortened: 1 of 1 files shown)


asturur:
ok does not change much. Most of the time is spent on apt-get here.
For fabricJS the switch to docker service made test quicker in general.

LinusU:
I recently tried Azure Pipelines, it's free for open source and supports Mac, Linux, and Windows 🙌 

I also found it to be very fast, we could potentially try that!

https://github.com/commitizen/cz-cli/commits/master/azure-pipelines.yml

PR #1281, opened 2018-10-14
Ship 2.0.0

zbjornson:
What do you guys think, is it time finally?

* No known regressions.
* Complete docs present for upgrading from v1.x.
* All open PRs are semver minor/patch, except Linus' #1280, which I think should be 3.x because it drops Node.js 6.x support.
* [Five possible semver-major issues open](https://github.com/Automattic/node-canvas/issues?q=is%3Aopen+is%3Aissue+label%3Asemver-major), but none have activity and another semver-major release can come in the near future.
* Prebuilds are working fairly smoothly. The largest improvement would probably come from uploading the binaries in the npm tarball itself, but the builds are too large right now. Using N-API so we have a stable ABI would require Node.js 10 for non-experimental support. I think the dependencies could be deduped between ABI builds though (should only need one per platform).

Closes #743 
Closes #916

Issue #1063, opened 2017-12-21
Use standard Canvas `ctx.filter` API (blur/grayscale/saturate/contrast)


paramaggarwal:
HTML5 Canvas recently introduced support for filters like `ctx.filter = "blur(5px)";` in the browser, for example to `blur`, `saturate`, `grayscale` and modify `contrast`.

## Issue or Feature
API support for HTML5 Canvas `ctx.filter` from [from MDN web docs](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/filter)

## Steps to Reproduce
```js
var Canvas = require('canvas');
var canvas = Canvas.createCanvas(200, 200);
var ctx = canvas.getContext('2d');
ctx.filter = "blur(5px)";
ctx.drawImage(img,0,0);
```

## Your Environment
* node-canvas: 2.0.0 alpha
* node: v9.3.0
* macOS: High Sierra

asturur:
This api is actually much new also for browsers ( chrome 49+ )

paramaggarwal:
Ah, right. Closing issue for now.

asturur:
I did not mean to make you close it. Just can be expected that is not implemented in  node-canvas since it became a standard non experimental api not long ago.

paramaggarwal:
No, no - I previously thought that this is a bug, but actually it is a feature request. Hence closed it. 😄

zbjornson:
Feature requests are valid issues :-)

binh27112004:
Same problem. I make a canvas with nodejs and use ctx.filter = 'blur(4px)'; to blur canvas then I convert it to png. 
![image](https://user-images.githubusercontent.com/46067234/78055756-380b7480-73ae-11ea-8493-28180f824001.png)
It doesn't work !!
My code:
`
const canvas = createCanvas(900, 900)
const ctx = canvas.getContext('2d')
ctx.filter = 'blur(400px)';
ctx.font = '48px serif';
ctx.fillText('Hello world', 50, 100);
fs.writeFileSync('layer-1.png', canvas.toBuffer());
`

asturur:
i would like to work on this, i wonder if we have the exact specs of the filters. Would be ok to implement those with normal loops over the pixels?

zbjornson:
Reopening because we still don't have this feature.

@asturur I think the full list is here: https://drafts.fxtf.org/filter-effects/#supported-filter-functions. 

> Would be ok to implement those with normal loops over the pixels?

[...]

PR #1282, opened 2018-10-14
Rename 'filter' to 'quality'

zbjornson:
I think this name makes more sense than "filter" anyway.

Ref #1063 and https://github.com/Automattic/node-canvas/pull/1281#issuecomment-429658736

Thanks for contributing!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,11 @@ canvas.createPNGStream() // new
 canvas.syncJPEGStream() // old
 canvas.createSyncJPEGStream() // old
 canvas.createJPEGStream() // new
+
+// (6) Context2d.filter has been renamed to context2d.quality to avoid a
+//     conflict with the new standard 'filter' property.
+context.filter = 'best' // old
+context.quality = 'best' // new
 ```
 
 ### Breaking
@@ -72,6 +77,9 @@ canvas.createJPEGStream() // new
  * See also: *Correct some of the `globalCompositeOperator` types* under
    **Fixed**. These changes were bug-fixes, but will break existing code relying
    on the incorrect types.
+ * Rename `context2d.filter` to `context2d.quality` to avoid a conflict with the
+   new standard 'filter' property. Note that the standard 'filter' property is
+   not yet implemented.
 
 ### Fixed
  * Fix build with SVG support enabled (#1123)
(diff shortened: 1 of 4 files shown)


LinusU:
Since I've already tagged 2.0.0 (sorry about that 🙈), would you mind adding a getter/setter with the old name (perhaps logging a deprecation warning)?

zbjornson:
Hrmmm that sort of defeats the purpose, I think? Could we unpublish 2.0.0 and release it as 2.0.1?

zbjornson:
This could also be deferred to 3.x though. Just nice to get as many breaking changes out of the way as possible. (I wonder if anyone actually uses this property?)

LinusU:
> Hrmmm that sort of defeats the purpose

I think it would be quite nice? Without a deprecation notice for the old setter, code setting the old property will not generate any warning/error (since you can add any property to any object in JS). So I would actually like it if 2.x had a deprecation warning, and 3.x throws an error when trying to set it. Then in 4.x we could remove it completely.

My mind is always open though, so please let me know if you feel otherwise ☺️

asturur:
i do not think anyone is gonna rush to implement ctx.filter ( i would to, but is not my priority ).
It would have been nice to have road unblocked.
Please consider that if jumping to 3.0 means killing node6 while is LTS, that could be less nice.
(since the wasm changes, are targeted for next major bump and required node 8 if i understood correctly)

zbjornson:
> jumping to 3.0

Sorry, I didn't mean that we had to jump to 3.x -- just that it's *not* critical to have this PR landed in 2.x. (I've also been using "3.x" in a few comments when I should have said "some major release after 2.x.")

---

> So I would actually like it if 2.x had a deprecation warning, and 3.x throws an error when trying to set it. Then in 4.x we could remove it completely.

Given the goal to open up that property for the implementation of the standard `filter` property, the two possible scenarios are these, right?

1. 2.x adds warning, 3.x gets standard impl. Downside is if someone is running isomorphic (browser+node) code that uses the standard `filter` attribute, it gets noisy in node instead of just being a noop (as it is in browsers that don't yet support `filter`).
1. 2.x removes attribute, 2.y (minor rel) gets standard impl. Until 2.y, acts the same as the browsers that don't support the property yet.

I agree a message would be friendly to users, but I don't think it's necessary (we aren't doing it for the changed globalCompositeOperators that were renamed, either). Throwing an error would be unusual and would make isomorphic code a pain.

I think this is a very low-impact PR, so I'm fine with any resolution. 🙂 My pref would still be for [semver major after 2.x] or [unpublish 2.0.0 and release 2.0.1], followed by [an alias with no warning].

LinusU:
Ahhhhaaaa, now I see. I thought that our implementation was compatible, just named something else 😄 

Hmmmm, okay. I'll unpublish 2.0.0, merge this, and release 2.0.1...

zbjornson:
Thanks! Sorry for the hassle.

LinusU:
No problem at all, it was I who was a bit fast to publish 2.0 😆

PR #1286, opened 2018-10-21
Update Readme.md to inform about xcode 10 and npm 6.4.1

asturur:
Add a detail about newer xcode and macosx


asturur:
@LinusU @zbjornson  this should be straight forward.

LinusU:
Thanks

PR #1288, opened 2018-10-22
Adapt to V8 7.0

Hakerh400:
The following methods

```
Value::ToObject
Value::ToString
Value::Int32Value
Value::NumberValue
Value::Uint32Value
Value::BooleanValue
```

that take no arguments are deprecated in V8 version 7.0 and will be removed soon.

---

Also reduced code duplication in `NAN_METHOD(ImageBackend::New)`. If the original implementation is better, I'll revert it.


zbjornson:
Awesome 👍 thanks for doing this, hoping to review tonight or tomorrow.

src/Canvas.cc
```diff
@@ -88,9 +88,9 @@ NAN_METHOD(Canvas::New) {
 
   Backend* backend = NULL;
   if (info[0]->IsNumber()) {
-    int width = info[0]->Uint32Value(), height = 0;
+    int width = info[0]->Uint32Value(Nan::GetCurrentContext()).FromMaybe(0), height = 0;
```
LinusU: I think that we should use the `Nan::To<...>(...)` API instead, to better support multiple versions of V8 🤔 

```suggestion
    int width = Nan::To<uint32_t>(info[0]).FromMaybe(0), height = 0;
```

Hakerh400: Fixed.

src/backend/Backend.h
```diff
@@ -14,6 +15,7 @@
 class Canvas;
 
 using namespace std;
+using namespace v8;
```
LinusU: Personally would prefer using `v8::Value` in this case...

LinusU: (somewhat related: https://stackoverflow.com/questions/1452721/why-is-using-namespace-std-considered-bad-practice)

Hakerh400: Done.

PR #1293, opened 2018-10-27
Reset context on resurface

### LINKED ISSUE #1292: Setting canvas width or height does not reset drawing context state
## Steps to Reproduce

```js
const { createCanvas } = require('canvas');

const canvas = createCanvas(500, 500);
context.globalAlpha = 0.5;

console.log('before', context.globalAlpha);

canvas.width = canvas.width;

console.log('after', context.globalAlpha);
```
Results on a browser (expected behaviour)
```
before 0.5
after 1
```
Results in nodejs (using node-canvas)
```
before 0.5
after 0.5
```

## Your Environment
* Version of node-canvas (e.g. 1.4.0): 2.0.1
* Environment (e.g. node 4.2.0 on Mac OS X 10.8): Windows 7, Windows 10, Linux

Hakerh400:
Fixes #1292
Did a lot of refactoring.
If any of these is better to stay as it was implemented before, I can revert it:

- Hide some non-standard context prototype accessors
- Replace deprecated `__defineGetter__` with `defineProperty`
- Add new private method `_updateStateInfo` to track when the surface gets reset
- Hide `lastFillStyle` and similar properties
- Reduce some code duplication

---

- [x] Update the changelog
- [x] Add a test


lib/context2d.js
```diff
@@ -18,6 +18,38 @@ const CanvasPattern = bindings.CanvasPattern
...
+  font: '10px sans-serif',
+  textBaseline: 'alphabetic',
+  textAlign: 'start',
+};
+
+const priv = privateMethodNames.reduce((methods, name) => {
+  const privName = `_${name}`;
+  const original = proto[privName];
+  methods[name] = original.call.bind(original);
+  delete proto[privName];
```
LinusU: Personally not a fan of modifying the prototype, `delete`ing props, and/or running too much logic when `require`ing a package, but I don't really have a better way to propose at the moment 😁 🤔

Hakerh400: That's basically what we have in #1247. However, deleting prototype properties is avoidable by exporting another class which will call the original static methods, but it will probably make all operations a bit slower.

LinusU: > but it will probably make all operations a bit slower.

I'm not sure if this is still true, but I remember V8 being very slow when modifying prototypes, and when using `delete` to remove properties 🤔

Hakerh400: Performed a test:

```js
console.time('Test');
for(let i = 0; i !== 1e6; i++){
  ctx.fillStyle += '';
  ctx.strokeStyle += '';
  ctx.globalAlpha |= 0;
  ctx.lineWidth |= 0;
  ctx.font += '';
}
console.timeEnd('Test');
```

Before PR: `3509.834ms`
After PR: `4650.146ms`

The extra `1100ms` is spent on WeakMap operations. Calls to old prototype methods doesn't seem to be slower than before.

> I remember V8 being very slow when modifying prototypes

I couldn't find anything about it, only found some articles on v8 blog saying that there was a bug in old version of v8 forcing internal objects' hash tables to be reallocated on deleting properties. Not sure if it is related, though. Currently, every lib file in node-canvas modifies a prototype of an internal exported class by defining getters and setters.

Also, from logical point of view, a prototype is just an ordinary object. Adding/deleting properties should be the same as defining a new object. Not sure how it can impact the performance. I performed another test: deleted random (unused) properties from the original implementation and ran the above script: time stays at `~3500ms`, so it seems that deleting prototype properties doesn't affect the performance.

I'm open for any ideas on how to improve the performance without exposing non-standard properties.

zbjornson: I think it would be nice to just do all of this in C++ so that there's no clunky manipulation of the exported C++ class and so the code is all in one spot. `parseFont` will be a pain because that's a lot of string manipulation code to port, but maybe we could move as much as easily possible now and deal with the rest later. Things like `createLinearGradient`, `currentTransform` and `textAlign` are all trivial to port, and I'm happy to help with that effort if you guys agree.

> exporting another class which will call the original static methods, but it will probably make all operations a bit slower.

Invoking a static method with an instance of the object shouldn't be slow. You're talking about something like this?

```js
const {privates} = require("../build/Release/canvas.node")
Object.defineProperty(proto, 'fillStyle', {
  set(val) {
    privates.SetFillStyle(this, val)
  }
})
```

> Adding/deleting properties should be the same as defining a new object. Not sure how it can impact the performance.

On why performance is/can be affected:
* https://developer.mozilla.org/en-US/docs/Web/JavaScript/The_performance_hazards_of__%5B%5BPrototype%5D%5D_mutation
* https://richardartoul.github.io/jekyll/update/2015/04/26/hidden-classes.html

What I don't know is when you're allowed to modify the proto without a penalty. ES5-ish code would set `Fn.prototype.method1 = ...`, then `method2 = ...`, etc. So is it parse-time that modification is allowed without penalty?

zbjornson: (Sorry for dupe. GitHub posted my comment twice and then deleted both copies when I tried to delete one.)

I think it would be nice to just do all of this in C++ so that there's no clunky manipulation of the exported C++ class and so the code is all in one spot. `parseFont` will be a pain because that's a lot of string manipulation code to port, but maybe we could move as much as easily possible now and deal with the rest later. Things like `createLinearGradient`, `currentTransform` and `textAlign` are all trivial to port, and I'm happy to help with that effort if you guys agree.

> exporting another class which will call the original static methods, but it will probably make all operations a bit slower.

Invoking a static method with an instance of the object shouldn't be slow. You're talking about something like this?

```js
const {privates} = require("../build/Release/canvas.node")
Object.defineProperty(proto, 'fillStyle', {
  set(val) {
    privates.SetFillStyle(this, val)
  }
})
```

> Adding/deleting properties should be the same as defining a new object. Not sure how it can impact the performance.

On why performance is/can be affected:
* https://developer.mozilla.org/en-US/docs/Web/JavaScript/The_performance_hazards_of__%5B%5BPrototype%5D%5D_mutation
* https://richardartoul.github.io/jekyll/update/2015/04/26/hidden-classes.html

What I don't know is when you're allowed to modify the proto without a penalty. ES5-ish code would set `Fn.prototype.method1 = ...`, then `method2 = ...`, etc. So is it parse-time that modification is allowed without penalty?

Hakerh400: > I think it would be nice to just do all of this in C++

That seems reasonable. Would something like [this](https://github.com/Hakerh400/node-canvas/commit/272a2ab71589051be2d57641846bf2f73bfbd8cc#diff-d52111b9458a13e196dab8299f822c29R1869) work?

> Things like `createLinearGradient`, `currentTransform` and `textAlign` are all trivial to port

The `currentTransform` rely on the JS implementation of `DOMMatrix`. Constructing it from C++ would probably be slow, wouldn't it?

> `parseFont` will be a pain

Can't we just cache the string user passed to the setter and return it back, like we did in JS land?

zbjornson: > Would something like this work?

Yep!

> The currentTransform rely on the JS implementation of DOMMatrix. Constructing it from C++ would probably be slow, wouldn't it?

It's not too bad. I did something similar the other day by retrieving the `require` function and invoking it with the module/path. The alternative is calling some `init`-ish C++ function with the `DOMMatrix` constructor and storing a persistent reference to it. I doubt anyone is using `currentTransform` in very hot code, so I wouldn't worry much about the extra 10s of nanoseconds here.

> Can't we just cache the string user passed to the setter and return it back, like we did in JS land?

Yeah, but I meant porting parse-font.js would be a bit of a pain. The regexps should port directly to C++ though, so maybe not so bad.

lib/context2d.js
```diff
@@ -97,153 +129,108 @@ Object.defineProperty(Context2d.prototype, 'currentTransform', {
...
-  if (val instanceof CanvasGradient
-    || val instanceof CanvasPattern) {
-    this.lastFillStyle = val;
-    this._setFillPattern(val);
-  } else {
-    this.lastFillStyle = undefined;
-    this._setFillColor(String(val));
+  // Get previous fill style.
+  get(){
+    return getProp(this, 'fillStyle');
```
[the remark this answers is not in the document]
Hakerh400: > The alternative is calling some init-ish C++ function with the DOMMatrix constructor and storing a persistent reference to it.

Good idea. That's what I did now and also moved all prototype accessors (except `createImageData` method) to C++.

zbjornson:
Really like the direction this is headed 👍. Hope to read it more closely tonight.

src/CanvasRenderingContext2d.cc
```diff
@@ -1600,6 +1641,127 @@ NAN_SETTER(Context2d::SetQuality) {
...
+/*
+ * Set current transform.
+ */
+
+NAN_SETTER(Context2d::SetCurrentTransform) {
+  Context2d *context = Nan::ObjectWrap::Unwrap<Context2d>(info.This());
+  Isolate *iso = Isolate::GetCurrent();
+  Local<Context> ctx = Nan::GetCurrentContext();
+
+  // TODO: Check if value is an instance of DOMMatrix
```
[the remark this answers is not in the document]
Hakerh400: There's no `InstanceOf` property of `v8::Object` in Node v6.x

Hakerh400:
Added some commits and rebased onto master. I'll add `createImageData`, but want to make sure everything so far is ok before adding it.

Checking if one `v8::Value` is an instance of another `v8::Value` doesn't seem trivial, unless I'm missing somthing. It can be done by constructing a new function using global.Function constructor that performs JS `instanceof` operator on the arguments and returns the result, but it is very slow.

Two questions: the BMP PR added a line into the changelog section for 2.1.0, but it landed after 2.1.0, right? I can move it in the unreleased section, if that is the case. Another question: tests `registerFont` and `Context2d#measureText().width` are flaky and fail randomly (see [this](https://travis-ci.org/Automattic/node-canvas/jobs/359431230), [this](https://travis-ci.org/Automattic/node-canvas/jobs/359438056), [this](https://travis-ci.org/Automattic/node-canvas/jobs/449739686), [this](https://travis-ci.org/Automattic/node-canvas/jobs/454117694) (before restart), and [this](https://travis-ci.org/Hakerh400/node-canvas/jobs/456007698)). Should we increase their timeout? If yes, let me know, all of that can be done in this PR.

src/CanvasRenderingContext2d.cc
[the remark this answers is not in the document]
zbjornson: Good catch.

> Checking if one v8::Value is an instance of another v8::Value doesn't seem trivial, unless I'm missing something. It can be done by constructing a new function using global.Function constructor that performs JS instanceof operator on the arguments and returns the result, but it is very slow.

From C++ you'd ideally use `FunctionTemplate::HasInstance`, but you can't get the `FunctionTemplate` of a JS function that is passed into C++. Since v6 is EOL in April I'd suggest doing the easy method I suggested above with a preprocessor directive limiting it to Node.js 8+

zbjornson:
> the BMP PR added a line into the changelog section for 2.1.0, but it landed after 2.1.0, right? 

Yeah, appears so. Thanks for finding.

> tests registerFont and Context2d#measureText().width are flaky and fail randomly

I don't think increasing the timeout is the right thing to do (they shouldn't take more than 2000 ms, unless the host has some I/O issue with slow font loading?). I've only seen them be flaky in CI, not locally. Not sure what the answer is there. I guess we could try increasing the timeout in a branch and run CI a bunch of times.

lib/context2d.js
```diff
@@ -18,233 +18,7 @@ const CanvasPattern = bindings.CanvasPattern
...
- * @api public
- */
-
-Context2d.prototype.__defineGetter__('textAlign', function(){
-  return this.lastTextAlignment || 'start';
-});
+bindings.CanvasRenderingContext2dInit(DOMMatrix, parseFont);
 
 /**
  * Create `ImageData` with the given dimensions or
```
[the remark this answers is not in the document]
Hakerh400: Done.

src/CanvasRenderingContext2d.cc
```diff
@@ -1083,6 +1125,48 @@ NAN_METHOD(Context2d::GetImageData) {
...
+    width = Nan::To<int32_t>(obj->Get(ctx, Nan::New("width").ToLocalChecked()).ToLocalChecked()).FromMaybe(0);
+    height = Nan::To<int32_t>(obj->Get(ctx, Nan::New("height").ToLocalChecked()).ToLocalChecked()).FromMaybe(0);
+  } else {
+    width = Nan::To<int32_t>(info[0]).FromMaybe(0);
+    height = Nan::To<int32_t>(info[1]).FromMaybe(0);
+  }
+
+  int stride = canvas->stride();
+  double Bpp = static_cast<double>(stride) / canvas->getWidth();
+  int nBytes = static_cast<int>(Bpp * width * height + .5);
```
zbjornson: Looks like this also fixes a bug in the original JS impl. 👍

CHANGELOG.md
```diff
@@ -9,13 +9,15 @@ project adheres to [Semantic Versioning](http://semver.org/).
 ==================
 ### Changed
 ### Added
+* BMP support
```
LinusU: I'm guessing it wasn't intentional to move this? ☺️

lib/context2d.js
```diff
@@ -10,264 +10,10 @@
...
-const bindings = require('./bindings')
-const parseFont = require('./parse-font')
-const Context2d = module.exports = bindings.CanvasRenderingContext2d
-const CanvasGradient = bindings.CanvasGradient
-const CanvasPattern = bindings.CanvasPattern
-const ImageData = bindings.ImageData
-const DOMMatrix = require('./DOMMatrix').DOMMatrix
+const bindings = require('./bindings');
+const parseFont = require('./parse-font');
+const {DOMMatrix} = require('./DOMMatrix');
```
LinusU: It would be nice to follow [standard](https://github.com/feross/standard) on all new code that we add ☺️ 

```suggestion
const { DOMMatrix } = require('./DOMMatrix')
```

Since this entire file is rewritten you could add it to [the list here](https://github.com/Automattic/node-canvas/blob/cc6c173bfd1d77ef1fd3552d655c0a4545218a2a/package.json#L28) as well 🎉 

(btw. I know this is a bit of a complicated way to convert a codebase 😄 but it's the one being used by Express.js as well. The goal of it is to not destroy the diffs and ability to backport stuff by rewriting all files at once, rather we just fix the lines that are already touched by a diff ☺️)

CHANGELOG.md
Hakerh400: BMP landed after 2.1.0

LinusU: Ah, sorry about that!

package.json
```diff
@@ -25,7 +25,7 @@
   "scripts": {
     "prebenchmark": "node-gyp build",
     "benchmark": "node benchmarks/run.js",
-    "pretest": "standard examples/*.js test/server.js test/public/*.js benchmark/run.js util/has_lib.js browser.js index.js && node-gyp build",
+    "pretest": "standard examples/*.js test/server.js test/public/*.js benchmark/run.js util/context2d.js util/has_lib.js browser.js index.js && node-gyp build",
```
LinusU:
```suggestion
    "pretest": "standard examples/*.js test/server.js test/public/*.js benchmark/run.js lib/context2d.js util/has_lib.js browser.js index.js && node-gyp build",
```

small typo ☺️

LinusU:
Awesome 👏

LinusU:
Thank you for this 🚀

PR #1307, opened 2018-11-20
Fix linting

piranna:
Thanks for contributing!

### DIFF
--- a/benchmarks/run.js
+++ b/benchmarks/run.js
@@ -65,8 +65,8 @@ function done (benchmark, times, start, isAsync) {
 // node-canvas
 
 bm('fillStyle= name', function () {
-  ctx.fillStyle = "transparent";
-});
+  ctx.fillStyle = 'transparent'
+})
 
 bm('lineTo()', function () {
   ctx.lineTo(0, 50)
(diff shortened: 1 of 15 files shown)


LinusU:
The problem with fixing all the linting at once is that it will make backporting stuff to 1.x harder, and it will cause pretty much every open pull request to have conflicts.

Without having formally written it down somewhere, we have been trying to convert all the lines that we touch when making other changes. This approach is taken from Express.js, unfortunately I haven't seen it being written down somewhere there either, so I don't have any good links to read up more on this.

I'm going to open up an issue in the discussion repo there, and then we can link to their policy when it's written down.

I do think that it's a good approach though, and unless any other maintainers feel strongly that it's a bad approach I would like for us to do it that way...

LinusU:
https://github.com/expressjs/discussions/issues/70

piranna:
I've done this pull-request because currently when you exec `npm test` locally there's a lot of errors regarding `standard`, so I've fixed them so they don't appear anymore. What other alternative I can do to don't get these errors without fixing the linting first?

LinusU:
That seems very strange, we are running `npm test` on Travis for every commit, and I can run it locally without errors as well.

If you look in the `test` script in the `package.json` you should see that it only lints files which are 100% converted?

piranna:
> If you look in the `test` script in the `package.json` you should see that it only lints files which are 100% converted?

Well, don't know why, but it does it to me for all the repo files from a freshly downloaded repo, including the examples... I've seen this in two different machines, with Node.js 10.x and 11.x.

LinusU:
Was that also with a fresh `node_modules`? If you didn't run `npm install` it might have picked up some globally installed `standard` command? 🤔 

I don't see how it could be possible that it runs on every file for you, especially since we are running the tests on Travis at every commit 🤔

LinusU:
I tried running `npm test --build-from-source` and that works for me as well, is that the exact command that is failing for you?

zbjornson:
(I get the same behavior as Linus locally, standard is only running on the specified subset of files.)

Issue #122, opened 2011-10-15
PNG with cHRM chunk fails to render


kangax:
Here's a simplified test case, where one image renders and another one fails (`url` vs `url2`). Both images open fine in a browser (and can be downloaded and viewed locally).

Image fails with "Error: Image given has not completed loaded".

I'm not 100% sure if the problem is with node-canvas. Would appreciate any pointers!

```
var Canvas = require('canvas')
  , Image = Canvas.Image
  , canvas = new Canvas(500, 500)
  , ctx = canvas.getContext('2d')
  , fs = require('fs')
  , URL = require('url')
  , HTTP = require('http')
  , url = 'http://s3-eu-west-1.amazonaws.com/printio/assets/images/large/9c6af4fc958a1967eaa4dba1200fef75ca2f68e9.png'
  , url2 = 'https://www.google.com/intl/en_com/images/srpr/logo3w.png';

function request(url, encoding, callback) {
  var oURL = URL.parse(url),
      client = HTTP.createClient(80, oURL.hostname),
      request = client.request('GET', oURL.pathname, { 'host': oURL.hostname });

  request.end();
  request.on('response', function (response) {
    var body = "";
    if (encoding) {
      response.setEncoding(encoding);
    }
    response.on('end', function () {
      callback(body);
    });
    response.on('data', function (chunk) {
      if (response.statusCode == 200) {
        body += chunk;
      }
    });
  });
}

request(url2, 'binary', function(body) {
  var img = new Image();
  img.src = new Buffer(body, 'binary');

  ctx.drawImage(img, 0, 0);

  var out = fs.createWriteStream(__dirname + '/output.png')
    , stream = canvas.createPNGStream();

  stream.on('data', function(chunk){
    out.write(chunk);
  });
});
```

missdora:
@choyongjoon It works for me! thank you!

zbjornson:
The `cHRM` chunk in http://s3-eu-west-1.amazonaws.com/printio/assets/images/large/9c6af4fc958a1967eaa4dba1200fef75ca2f68e9.png is causing the failure. I think we need to use libpng directly instead of cairo's png functions (which it says are toys). That would let us fix #73 as well (which is due to a gAMA chunk bomb).

zbjornson:
This works with Cairo 1.16.0 🎉 .

PR #1324, opened 2018-12-12
Detect invalid PNG and prevent segfault

Hakerh400:
IIUC, the cairo PNG parser reads data until the IEND chunk is reached. It doesn't perform any overrun checks, and if there is no IEND chunk, it will cause a segfault.

### DIFF
--- a/test/image.test.js
+++ b/test/image.test.js
@@ -84,6 +84,12 @@ describe('Image', function () {
     })
   })
 
+  it('detects invalid PNG', function (done) {
+    const img = new Image()
+    img.onerror = () => done()
+    img.src = Buffer.from('89504E470D', 'hex')
+  })
+
   it('loads SVG data URL base64', function () {
     const base64Enc = fs.readFileSync(svg_tree, 'base64')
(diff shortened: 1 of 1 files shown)


zbjornson:
Good find. That seems like a pretty gross bug in cairo...

We also have #73 and #122 (and others?) open that AFAIK are unfixable without switching to libpng for loading images (instead of just writing). Any chance you're interested in doing that instead? Would be a cleaner fix...

Hakerh400:
Is this PR still relevant? None of the mentioned issues reproduce with Cairo 1.16.0.

zbjornson:
Sorry for the delay. With cairo 1.16.0:

- the test case in this PR passes
- the gAMA bomb in #73 loads fine
- the image in the OP in #122 loads fine. Still get stdout as TJ described (https://github.com/Automattic/node-canvas/issues/122#issuecomment-2423574).

Glad you checked with newer Cairo 👍 . We could merge just the test from this PR if you want to drop the other changes.

Hakerh400:
Thanks for verifying. Left the test only.

LinusU:
Nice!

PR #1328, opened 2018-12-13
Catch font errors

### LINKED ISSUE #1327: Hard crash when setting font with invalid size
## Steps to Reproduce
Since 96bdd4f9d883b5ff11d0409d9a24a6879db579d2, this will cause a hard crash from an empty MaybeLocal because parseFont returns undefined.

```js
const {createCanvas} = require(".");
const ctx = createCanvas(10,10).getContext("2d");
ctx.font = "bold undefinedpx Arial";
```

Crashing at
https://github.com/Automattic/node-canvas/blob/7de2441ee517a3681c5898a9a8520f4bed9619cc/src/CanvasRenderingContext2d.cc#L2455

## Your Environment
* Version of node-canvas (output of `npm list canvas` or `yarn list canvas`): 96bdd4f9d883b5ff11d0409d9a24a6879db579d2
* Environment (e.g. node 4.2.0 on Mac OS X 10.8): Win10, Node 10

Hakerh400:
Fix #1327

### DIFF
--- a/src/CanvasRenderingContext2d.cc
+++ b/src/CanvasRenderingContext2d.cc
@@ -2451,8 +2451,12 @@ NAN_SETTER(Context2d::SetFont) {
 
   const int argc = 1;
   Local<Value> argv[argc] = { value };
-  Local<Value> parsed = _parseFont.Get(iso)->Call(ctx, ctx->Global(), argc, argv).ToLocalChecked();
-  Local<Object> font = Nan::To<Object>(parsed).ToLocalChecked();
+
+  MaybeLocal<Value> mparsed = _parseFont.Get(iso)->Call(ctx, ctx->Global(), argc, argv);
+  if (mparsed.IsEmpty()) return;
+  MaybeLocal<Object> mfont = Nan::To<Object>(mparsed.ToLocalChecked());
+  if (mfont.IsEmpty()) return;
+  Local<Object> font = mfont.ToLocalChecked();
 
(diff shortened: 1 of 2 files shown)


src/CanvasRenderingContext2d.cc
```diff
@@ -2451,8 +2451,19 @@ NAN_SETTER(Context2d::SetFont) {
...
+  if (try_catch.HasCaught()){
+    try_catch.ReThrow();
+    return;
+  }
+  MaybeLocal<Object> maybeFont = Nan::To<Object>(maybeParsed.ToLocalChecked());
+  if (try_catch.HasCaught()){
+    try_catch.ReThrow();
+    return;
+  }
+  Local<Object> font = maybeFont.ToLocalChecked();
```
zbjornson: Thanks for the fast fix :).

A bit simpler: you can explicitly check if the `MaybeLocal` is empty with something like:

```cpp
  MaybeLocal<Object> mlfont = Nan::To<Object>(parsed);
  if (mlfont.IsEmpty()) return;
  Local<Object> font = mlfont.ToLocalChecked();
```

The error will still be thrown appropriately when the handle is empty.

Hakerh400: Done. Is there a need to test if `mparsed` is empty (can `parseFont` itself throw an error)?

zbjornson: Technically I'm not sure if there's any v8 operation that returns a `MaybeLocal` that is guaranteed not to fail. I'm not sure how unwise it is to assume simple operations don't fail. Having it doesn't hurt.

Issue #1218, opened 2018-08-06
Image is missing inherited methods


zbjornson:
Image is missing at least `addEventListener` and `removeEventListener`. See https://github.com/Automattic/node-canvas/issues/730#issuecomment-410588856.

demurgos:
I'm just posting this quote here because it's relevant:

> I'm not sure if we should (a) use the Node.js EventEmitter API (`on()`, `addListener()`, `off()`, `removeListener()`), (b) use the HTML API (`addEventListener()`, `removeEventListener()`), or (c) use [a] along with aliases for [b]. (If you have any thoughts on it, feel free to comment in the issue!)

_Originally posted by @zbjornson in https://github.com/Automattic/node-canvas/pull/1330_

---

My opinion is that the lib is intended to be used universally and bring a browser API to Node. As such, it should follow the browser API with `addEventListener` and `removeListener`. Node's `EventEmitter` is pretty complex, adding it on top of the browser API would be very confusing.

makc:
It is 2021 and I still have to do
```js
		const img = new window.Image();
		img.addEventListener = (type, handler) => {
			img['on' + type] = handler;
		};
		img.removeEventListener = (type) => {
			img['on' + type] = null;
		};
```

PR #1330, opened 2018-12-15
Add typescript definitions

### LINKED ISSUE #1303: Add Typescript typings
There are currently no Typescript typings available for this repo. This makes using this package from a Typescript library very difficult and inconvenient.

Here is a start (not my own work):
https://github.com/gmarty/nesCompiler/blob/master/src/declarations/node-canvas.d.ts

zbjornson:
Thanks for contributing!

I'm not a typescript expert, so if anyone sees any better ways to do things, please speak up! :) @varHarrie @adamjking3

Note that typescript between 2.9ish and 3.1ish renamed `Canvas2DContextAttributes` to `CanvasRenderingContext2DSettings`. People using older typescript won't have correct definitions.

Fixes #1303

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,3 +19,6 @@ addons:
 env:
   - CXX=g++-4.9
 sudo: false
+script:
+  - node_version=$(node -v); if [ ${node_version:1:2} = 10 ]; then npm run dtslint; fi
+  - npm test
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -1,0 +1,338 @@
+// TypeScript Version: 3.0
+/// <reference lib="dom" />
+
+import { Readable } from 'stream'
+
+export interface PngConfig {
+	/** Specifies the ZLIB compression level. Defaults to 6. */
+	compressionLevel?: 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9
+	/**
+	 * Any bitwise combination of `PNG_FILTER_NONE`, `PNG_FITLER_SUB`,
+	 * `PNG_FILTER_UP`, `PNG_FILTER_AVG` and `PNG_FILTER_PATETH`; or one of
+	 * `PNG_ALL_FILTERS` or `PNG_NO_FILTERS` (all are properties of the canvas
+	 * instance). These specify which filters *may* be used by libpng. During
+	 * encoding, libpng will select the best filter from this list of allowed
+	 * filters. Defaults to `canvas.PNG_ALL_FITLERS`.
+	 */
+	filters?: number
+	/**
+	 * _For creating indexed PNGs._ The palette of colors. Entries should be in
+	 * RGBA order.
+	 */
+	palette?: Uint8ClampedArray
+	/**
+	 * _For creating indexed PNGs._ The index of the background color. Defaults
+	 * to 0.
+	 */
+	backgroundIndex?: number
+	/** pixels per inch */
+	resolution?: number
+}
+
+export interface JpegConfig {
+	/** Specifies the quality, between 0 and 1. Defaults to 0.75. */
+	quality?: number
+	/** Enables progressive encoding. Defaults to `false`. */
+	progressive?: boolean
+	/** Enables 2x2 chroma subsampling. Defaults to `true`. */
+	chromaSubsampling?: boolean
+}
+
+export interface PdfConfig {
+	title?: string
+	author?: string
+	subject?: string
+	keywords?: string
+	creator?: string
+	creationDate?: Date
+	modDate?: Date
+}
+
+export interface NodeCanvasRenderingContext2DSettings {
+	alpha?: boolean
+	pixelFormat?: 'RGBA32' | 'RGB24' | 'A8' | 'RGB16_565' | 'A1' | 'RGB30'
+}
+
+export class Canvas {
+	width: number
+	height: number
+
+	/** _Non standard._ The type of the canvas. */
+	readonly type: 'image'|'pdf'|'svg'
+
+	// This is a getter, but it's non-standard and I don't know why we export it.
+	// readonly stride: number;
+
+	/** Constant used in PNG encoding methods. */
+	readonly PNG_NO_FILTERS: number
+	/** Constant used in PNG encoding methods. */
+	readonly PNG_ALL_FILTERS: number
+	/** Constant used in PNG encoding methods. */
+	readonly PNG_FILTER_NONE: number
+	/** Constant used in PNG encoding methods. */
+	readonly PNG_FILTER_SUB: number
+	/** Constant used in PNG encoding methods. */
+	readonly PNG_FILTER_UP: number
+	/** Constant used in PNG encoding methods. */
+	readonly PNG_FILTER_AVG: number
+	/** Constant used in PNG encoding methods. */
+	readonly PNG_FILTER_PAETH: number
+
+	constructor(width: number, height: number, type?: 'pdf'|'svg')
+
+	getContext(contextId: '2d', contextAttributes?: NodeCanvasRenderingContext2DSettings): NodeCanvasRenderingContext2D
+
+	/**
+	 * For image canvases, encodes the canvas as a PNG. For PDF canvases,
+	 * encodes the canvas as a PDF. For SVG canvases, encodes the canvas as an
+	 * SVG.
+	 */
+	toBuffer(cb: (err: Error|null, result: Buffer) => void): void
+	toBuffer(cb: (err: Error|null, result: Buffer) => void, mimeType: 'image/png', config?: PngConfig): void
+	toBuffer(cb: (err: Error|null, result: Buffer) => void, mimeType: 'image/jpeg', config?: JpegConfig): void
+
+	/**
+	 * For image canvases, encodes the canvas as a PNG. For PDF canvases,
+	 * encodes the canvas as a PDF. For SVG canvases, encodes the canvas as an
+	 * SVG.
+	 */
+	toBuffer(): Buffer
+	toBuffer(mimeType: 'image/png', config?: PngConfig): Buffer
+	toBuffer(mimeType: 'image/jpeg', config?: JpegConfig): Buffer
+	toBuffer(mimeType: 'application/pdf', config?: PdfConfig): Buffer
+
+	/**
+	 * Returns the raw ARGB data without encoding in native-endian byte order,
+	 * top-to-bottom.
+	 */
+	toBuffer(mimeType: 'raw'): Buffer
+
+	createPNGStream(config?: PngConfig): PNGStream
+	createJPEGStream(config?: JpegConfig): JPEGStream
+	createPDFStream(config?: PdfConfig): PDFStream
+
+	/** Defaults to PNG image. */
+	toDataURL(): string
+	toDataURL(mimeType: 'image/png'): string
+	toDataURL(mimeType: 'image/jpeg', quality?: number): string
+	/** _Non-standard._ Defaults to PNG image. */
+	toDataURL(cb: (err: Error|null, result: string) => void): void
+	/** _Non-standard._ */
+	toDataURL(mimeType: 'image/png', cb: (err: Error|null, result: string) => void): void
+	/** _Non-standard._ */
+	toDataURL(mimeType: 'image/jpeg', cb: (err: Error|null, result: string) => void): void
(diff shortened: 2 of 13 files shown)


rauchg:
This would be really nice!

varHarrie:
It seems good, and we should also specify `types` in `package.json`.

krzysztof-miemiec:
After copy-pasting it into my node project, I see that the typings depend on `lib.dom.d.ts` typings, and `"lib": ["dom"]` has to be present in `tsconfig.json` file - I think that it's worth mentioning in `README`. I also got the following TypeScript errors:
```
Error:(130, 5) TS2661: Cannot export 'CanvasGradient'. Only local declarations can be exported from a module.
Error:(131, 5) TS2661: Cannot export 'CanvasPattern'. Only local declarations can be exported from a module.
Error:(140, 5) TS2416: Property 'src' in type 'Image' is not assignable to the same property in base type 'HTMLImageElement'.
  Type 'string | Buffer' is not assignable to type 'string'.
    Type 'Buffer' is not assignable to type 'string'.
Error:(149, 15) TS2702: 'Image' only refers to a type, but is being used as a namespace here.
Error:(149, 34) TS2702: 'Image' only refers to a type, but is being used as a namespace here.
Error:(216, 12) TS2661: Cannot export 'DOMMatrix'. Only local declarations can be exported from a module.
Error:(216, 23) TS2661: Cannot export 'DOMPoint'. Only local declarations can be exported from a module.
Error:(218, 12) TS2661: Cannot export 'ImageData'. Only local declarations can be exported from a module.
```

I think that reexports can be changed to something like: `export const DOMMatrix: DOMMatrix;`, however I'm not sure what to do with `HTMLImageElement` error.

krzysztof-miemiec:
Oh, and when I started to use your typings, I saw the need to export `NodeCanvasRenderingContext2D` from them. I think it's a good practice to export all interfaces that can be returned by functions, so that library user can do something like this:

```typescript
class MyCanvasWrapper {
  private readonly canvas: Canvas;
  private readonly context: NodeCanvasRenderingContext2D;

  constructor(width: number, height: number) {
    this.canvas = createCanvas(width, height);
    this.context = this.canvas.getContext('2d');
  }
}

```

index.d.ts
```diff
@@ -0,0 +1,223 @@
+import { Readable } from "stream";
```
krzysztof-miemiec: Actually, I realized that `Readable` isn't used anywhere in the code, I think it should be `ReadableStream`.
```suggestion
/// <reference lib="dom" />
import { ReadableStream } from "stream";
```

j:
Status of this?

zbjornson:
Got busy working on bugfixes -- will update this this weekend.

index.d.ts
zbjornson: (Other way around actually -- import was correct, usage in code should have been `Readable`)

zbjornson:
Ready for another review/testing.

Used the instructions in https://github.com/Microsoft/dtslint#add-types-for-a-library-not-on-definitelytyped for setup.

bmamouri:
@zbjornson Thank you for the great work. Wondering when this will be merged?

zbjornson:
Just tested this in a module where I'm using node-canvas. I forgot the PNG constants, `drawImage(Canvas.Canvas, ...)` specialization and `canvas.width/height`, but I just fixed those and everything seems to work well.

https://github.com/Automattic/node-canvas/pull/1330/files#diff-88bec6beae84369c0376b56b3bb88fe1R200 this is sort of ugly but I don't see a way around it aside from providing all of our own Context2D types. That might not be a bad idea since we don't have full support yet for the entire HTML Canvas API (see https://github.com/Automattic/node-canvas/wiki/Compatibility-Status).

**What do folks think we should do?**

1. Use this PR as-is and have ~95% accurate types. The types will claim that `ctx.direction`, `ctx.filter`, `scrollPathIntoView()` and a handful of other properties/methods work, but they don't. Some of them never will (like `scrollPathIntoView()`).
2. Copy from lib.dom.d.ts into this library so that we can remove the things we don't provide and make them ~100% accurate?

LinusU:
Hmm, seems like `OffscreenCanvas` gives back a normal `CanvasRenderingContext2D` when calling `getContext()`, so I wonder how browsers handle calling `scrollPathIntoView()` from a web worker 🤔 

I think I'm leaning towards option `1`, but I could be convinced either way ☺️

types/index.d.ts
```diff
@@ -0,0 +1,329 @@
...
+	 * Applies to JPEG images drawn to PDF canvases only. Setting
+	 * `img.dataMode = Image.MODE_MIME` or `Image.MODE_MIME|Image.MODE_IMAGE`
+	 * enables image MIME data tracking. When MIME data is tracked, PDF canvases
+	 * can embed JPEGs directly into the output, rather than re-encoding into
+	 * PNG. This can drastically reduce filesize and speed up rendering.
+	 */
+	dataMode: number
+
+	onload(): void
+	onerror(err: Error): void
```
demurgos: Since you can set and unset the handlers, they should be marked as optional:

```diff
-	onload(): void
-	onerror(err: Error): void
+	onload?(): void
+	onerror?(err: Error): void
```

zbjornson: Will make the declaration fix, thanks.

That shouldn't be necessary to avoid leaks since https://github.com/Automattic/node-canvas/pull/1003. If you can repro a leak with the listeners, can you please open an issue for it?

demurgos: Thanks for the link, I wasn't aware that `node-canvas` can avoid these kinds of leaks. I still prefer to manually unset the handler: not all libraries handle this properly so I prefer to always do it rather than remembering which libraries are smarter.

By the way: I always use the documented handlers but does `Image` supports an event-emitter based API? (`addEventListener`, `removeEventListener`, etc.)? If so, it should also be typed.

zbjornson: Good question -- no, that's #1218. The only reason I haven't worked on that is because I'm not sure if we should (a) use the Node.js EventEmitter API (`on()`, `addListener()`, `off()`, `removeListener()`), (b) use the HTML API (`addEventListener()`, `removeEventListener()`), or (c) use [a] along with aliases for [b]. (If you have any thoughts on it, feel free to comment in the issue!)

icfantv:
Hey guys, anything I can do to get the ball rolling on this again?  Any more updates to the typings that need to happen?  Happy to help.  Cheers.

playground:
Will this be released soon?

zbjornson:
I'll finish up the last set of changes this weekend, sorry for the delay.

playground:
> I'll finish up the last set of changes this weekend, sorry for the delay.

Thanks @zbjornson, the timing is perfect as I'm working on porting the existing repo to typescript.  Can't wait for it to be ready.  I will check back on Monday have a good weekend.

1999:
Wow, nice work @zbjornson! I didn't notice that when I raised a PR for DefinitelyTyped repo (https://github.com/DefinitelyTyped/DefinitelyTyped/pull/33555), so I will close the PR and wait for this one to get merged.

types/index.d.ts
```diff
@@ -0,0 +1,329 @@
...
+	 * _Non-standard_. Defaults to 'good'. Affects pattern (gradient, image,
+	 * etc.) rendering quality.
+	 */
+	patternQuality: 'fast' | 'good' | 'best' | 'nearest' | 'bilinear'
+
+	/**
+	 * _Non-standard_. Defaults to 'good'. Like `patternQuality`, but applies to
+	 * transformations affecting more than just patterns.
+	 */
+	quality: 'fast' | 'good' | 'best' | 'nearest' | 'bilinear'
```
septs: **quality** and **patternQuality** share typing definition.

```typescript
type Quality = 'fast' | 'good' | 'best' | 'nearest' | 'bilinear';

quality: Quality
patternQuality: Quality
```

types/index.d.ts
```diff
@@ -0,0 +1,329 @@
...
+	toBuffer(mimeType: 'image/jpeg', config?: JpegConfig): Buffer
+	toBuffer(mimeType: 'application/pdf', config?: PdfConfig): Buffer
+
+	/**
+	 * Returns the raw ARGB data without encoding in native-endian byte order,
+	 * top-to-bottom.
+	 */
+	toBuffer(mimeType: 'raw'): Buffer
+
+	createPNGStream(config?: PngConfig): Readable
```
septs: https://github.com/Automattic/node-canvas/pull/1330/files#diff-c472f8ca9fb190272d5098f1ccd7bd41R32

requires definition `pngStream` , `jpegStream` , `pdfStream` typing

demurgos:
Would it be possible to publish this PR and iterate? It's been pending for some time now.

zbjornson:
*grumble* a dependency of dtslint added an async function, which breaks our CI that runs with node 6. Thinking of better workarounds, but if I can't think of one today I'll remove dtslint (sadly; it's been critical to testing this) and merge.

zbjornson:
There we go.

@LinusU since you have to do releases anyway, I'll let you merge. I just changed .travis.yml to only run dtslint with Node v10.

types/index.d.ts
```diff
@@ -0,0 +1,344 @@
...
+	toBuffer(mimeType: 'image/jpeg', config?: JpegConfig): Buffer
+	toBuffer(mimeType: 'application/pdf', config?: PdfConfig): Buffer
+
+	/**
+	 * Returns the raw ARGB data without encoding in native-endian byte order,
+	 * top-to-bottom.
+	 */
+	toBuffer(mimeType: 'raw'): Buffer
+
+	createPNGStream(config?: PngConfig): PngStream
```
septs: https://github.com/Automattic/node-canvas/pull/1330/files#diff-c472f8ca9fb190272d5098f1ccd7bd41R32

requires definition pngStream , jpegStream , pdfStream method typing

zbjornson: Ahh I thought you were asking for those classes to be defined, now I see you meant the method aliases. Those method aliases are present for backwards compatibility; we standardized on Node.js core's idiom of `create____` in #1152. I'd rather not type them since they're not otherwise documented, but can add them if you think it's important.

septs: I think it should be backwards compatible.

Maybe you can learn how to React.

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/e1f9647734bd990f49b0aca0f44acd4d55fe1e93/types/react/index.d.ts#L456-L462

.travis.yml
```diff
@@ -19,3 +19,6 @@ addons:
 env:
   - CXX=g++-4.9
 sudo: false
+script:
+  - node_version=$(node -v); if [ ${node_version:1:2} = 10 ]; then yarn dtslint; fi
+  - yarn test
```
LinusU: Maybe `npm test` (and `npm run dtslint` above)? I don't think we are using yarn anywhere else? 🤔

LinusU:
Nice! I'm out and about at the moment but will try to merge and release this asap!

Feel free to address my Yarn comment, otherwise I'll fix it myself before merging 👍

playground:
@zbjornson will this be ready soon?

zbjornson:
It's ready. @LinusU I forgot to post that I addressed your comment.

playground:
Thanks @zbjornson, can't wait, I'm currently blocked by this :-) as I'm porting existing codebase to Typescript.

LinusU:
Awesome work as always @zbjornson ❤️

playground:
@zbjornson @LinusU I'm getting error with tsc

ERROR in ./node_modules/canvas/build/Release/canvas.node 1:0
Module parse failed: Unexpected character '�' (1:0)
You may need an appropriate loader to handle this file type.
(Source code omitted for this binary file)

icfantv:
@zbjornson @LinusU hey guys, when do you think you'll be able to cut a new release to NPM?  Thanks.

LinusU:
was hoping to get #1255 in after @zbjornson have taken a look at it, but otherwise we can ship that later...

Will try to cut a release tonight or tomorrow, regardless of if it's merged or not ☺️

zbjornson:
Good plan; I'll try to get to that tonight, but lots of people are waiting on typescript defs, so no need to delay it.

@playground that means something (usually webpack) is trying to load the `.node` binary as a JS file. If you search for "webpack native module" you should find some advice. One option I've seen people use is https://github.com/webpack-contrib/node-loader, or just not use webpack for Node.js/server-side.

PR #1331, opened 2018-12-17
Fix memory leaks in toBuffer

### LINKED ISSUE #1296: Memory leaks in asynchronous image saving methods
## Issue
Memory leaks appear in asynchronous version of `Canvas.toBuffer` method and `Canvas.createPNGStream` method

- `Canvas.toBuffer` causes large memory leak
- `Canvas.createPNGStream` causes small memory leak

## Steps to Reproduce

```js
const fs = require('fs');
const { createCanvas, Image } = require('canvas');

const image = new Image();
image.src = fs.readFileSync('some_test_image.png');

const canvas = createCanvas(image.width, image.height);
canvas.getContext('2d').drawImage(image, 0, 0);

function saveUsingToBuffer(path, canvas) {
		canvas.toBuffer((err, buffer) => {
			if (err) {
				reject(err);
			} else {
				fs.writeFile(path, buffer, err => {
					if (err) {
						reject(err);
					} else {
						resolve();
					}
				});
			}
		});
	});
}

function saveUsingSyncToBuffer(path, canvas) {
		const buffer = canvas.toBuffer();
		fs.writeFile(path, buffer, err => {
			if (err) {
				reject(err);
			} else {
				resolve();
			}
		});
	});
}

function saveUsingStream(path, canvas) {
		const out = fs.createWriteStream(path);
		const stream = canvas.createPNGStream();
		stream.pipe(out);
		out.on('finish', resolve);
		out.on('error', reject);
	});
}

async function test() {
	while (true) {
		// uncomment one test to run
		// await saveUsingSyncToBuffer('test.png', canvas); // synchronous toBuffer() (no leak)
		await saveUsingToBuffer('test.png', canvas); // asynchronous toBuffer() (large leak)
		// await saveUsingStream('test.png', canvas); // createPNGStream() (small leak)
		gc();
	}
}

test();
```

You need to provide some test png image to load `some_test_image.png`
The script will keep saving `test.png` image in current path
Run with `--expose-gc` flag (or comment out call to `gc()`

## Your Environment
* Version of node-canvas: 2.0.1
* Environment: node 9.11.2 on Windows 10 x64

zbjornson:
free(data) was apparently omitted from the code because it caused a crash with JPEG export. That crash turns out to be because at least on Windows using prebuilt libjpeg-turbo, the malloc() that jpeg_mem_dest uses doesn't work with the free() that MSVC provides (different heap). Since libjpeg doesn't provide a way to call its particular free() function, we need to use our own version of jpeg_mem_dest. That omission of free() also caused a leak in toBuffer("image/png").

Replaced the toBuffer.cc code with std::vector. Also finished a TODO to cleanup the two callback functions.

image

Sorta hard to tell if that's stable. If there are leaks left, they're small.

@chearon FYI this PR changes binding.gyp.

Fixes #1296


bevand10:
Hi.

Anyone come across a similar memory leak issues with use of `toBuffer('raw')`?

Currently using v2.8.0 on Node 14 (execution context is AWS Lambda).

Multiple calls to toBuffer on a changing canvas cause memory-used to escalate. Canvas dimensions are 1920x1080.

PR #1335, opened 2018-12-20
add test demonstrating drawImage reflection bug

bennlich:
screenshot from 2018-12-20 15-47-43

Maybe related to https://github.com/Automattic/node-canvas/issues/1249 ?

This seems to occur when you use `context.drawImage` with a source image size that is less than the destination size.


asturur:
i think my open pr would fix that. i ll try and in case finish it.

bennlich:
Woo! Y'all are so speedy!

PR #1341, opened 2018-12-22
Clean up #includes

zbjornson:
This has been bugging me for ages. Each header can now stand alone and has no `using namespace`s.

I don't think this creates any conflicts with open PRs that are close to landing, except my #1340.

(Can we remove the comments at the top of the files from 2010 with the file name and copyright notice?)

### DIFF
--- a/binding.gyp
+++ b/binding.gyp
@@ -105,7 +105,7 @@
                   'WarningLevel': 4,
                   'ExceptionHandling': 1,
                   'DisableSpecificWarnings': [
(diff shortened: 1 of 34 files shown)


zbjornson:
Looks like I'm missing a header GCC -- will fix when I fix the conflicts after #1340 goes in.

zbjornson:
Ready to go! I tacked on a few more small fixups. Tested with GCC 4.8.5, GCC 8.1.0, MSVC 2015 and MSVC 2017. Can someone please check with Apple Clang?

LinusU:
I'll test with Apple Clang today 🙌

PR #1344, opened 2018-12-30
Hex 8 backport

asturur:
Backport color hex with 8 or 4 digits to branch 1.x


zbjornson:
I'm not sure if we should add any new features to 1.x, or keep 1.x to critical bug fixes only. Is it possible for fabric.js to update to 2.x? It looks like quite a few open issues in fabric would be resolved by requiring 2.x.

asturur:
yes fabric will move to 2 soon. But version 2.x will stay on node canvas 1.x and is widespread.

This would be a fetaure or a bugfix?

zbjornson:
I'd consider this a feature (it's a feature of the newer CSS Color Module Level 4). Personally I'd prefer to only add critical bug fixes to 1.x, in part to encourage people to upgrade.

asturur:
So you should close this.

I thought it could be useful to others users other than me!

zbjornson:
Happy to reconsider if others feel otherwise. I think it'll be useful, but that also means it's motivation for people to upgrade to 2.x. :)

PR #1382, opened 2019-03-15
Fix missing encoding options (quality) parameter in `canvas.toDataURL()`

kayahr:
I noticed that toDataURL always created the same URL, no matter what encoding options I specified. So I looked into the source code and found out that the `opts` variable is not always passed to `toBuffer`. This PR fixes this.


LinusU:
Thanks!

PR #1383, opened 2019-03-15
Fix Electron 5 and NodeJS 12 compatibility

Rosga:
Electron 5 is still in beta and will be shipped with Node 12. There are some breaking changes to native modules: https://electronjs.org/blog/nodejs-native-addons-and-electron-5

With latest version the command `node-gyp rebuild --target=5.0.0-beta.5 --runtime=electron` fails. This PR fixes build, and it works fine with my Electron application. 

Unfortunately, some new deprecated warnings have appeared. I'm not so experienced in writing native addons and I'd appreciate help with that. Or, if it is necessary, I could give it a try, but need more time with that. 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 ### Changed
 ### Added
 ### Fixed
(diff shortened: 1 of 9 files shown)


LinusU:
Thanks!

PR #1384, opened 2019-03-15
Use size_t where appropriate and guard against overflow

LinusU:
ping @zbjornson, @chearon 

I think that `size_t` might be the correct type here since it's a "type which is used to represent sizes of objects in bytes", and "it is guaranteed to be big enough to contain the size of the biggest object that system can handle".

### DIFF
--- a/src/Image.cc
+++ b/src/Image.cc
@@ -10,6 +10,11 @@
 #include <node_buffer.h>
 #include "Util.h"
 
+/* Cairo limit:
+  * https://lists.cairographics.org/archives/cairo/2010-December/021422.html
+  */
+static constexpr const size_t canvas_max_side = (1 << 15) - 1;
+
 #ifdef HAVE_GIF
 typedef struct {
   uint8_t *buf;
@@ -607,10 +612,7 @@ Image::loadGIFFromBuffer(uint8_t *buf, unsigned len) {
   width = naturalWidth = gif->SWidth;
   height = naturalHeight = gif->SHeight;
 
-  /* Cairo limit:
-   * https://lists.cairographics.org/archives/cairo/2010-December/021422.html
-   */
-  if (width > 32767 || height > 32767) {
+  if (width > canvas_max_side || height > canvas_max_side) {
     GIF_CLOSE_FILE(gif);
     return CAIRO_STATUS_INVALID_SIZE;
   }
@@ -1097,6 +1099,12 @@ Image::loadJPEG(FILE *stream) {
 
     jpeg_read_header(&args, 1);
     jpeg_start_decompress(&args);
+
+    if (args.output_width > canvas_max_side || args.output_height > canvas_max_side) {
+      jpeg_destroy_decompress(&args);
+      return CAIRO_STATUS_INVALID_SIZE;
+    }
+
     width = naturalWidth = args.output_width;
     height = naturalHeight = args.output_height;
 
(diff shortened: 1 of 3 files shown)


src/Image.cc
```diff
@@ -799,8 +800,26 @@ void Image::jpegToARGB(jpeg_decompress_struct* args, uint8_t* data, uint8_t* src
...
+  }
+
+  size_t size = naturalHeight * stride;
+  if (naturalHeight != 0 && size / naturalHeight != stride) {
+    // The integer overflowed, which means we cannot allocate that much memory.
+    jpeg_abort_decompress(args);
+    jpeg_destroy_decompress(args);
+    this->errorInfo.set(NULL, "malloc", errno);
+    return CAIRO_STATUS_NO_MEMORY;
+  }
```
zbjornson: You could consider combining these two `if` bodies into one since you don't have to check overflow *immediately* after doing the mul.

src/Image.h
```diff
@@ -37,8 +37,8 @@ using JPEGDecodeL = std::function<uint32_t (uint8_t* const src)>;
 class Image: public Nan::ObjectWrap {
   public:
     char *filename;
-    int width, height;
-    int naturalWidth, naturalHeight;
+    size_t width, height;
+    size_t naturalWidth, naturalHeight;
```
zbjornson: I'm trying to find any cases where `int` isn't big enough. The only one I found is reading a jpeg, where it's `unsigned`. GifLib, Cairo and librsvg all use int as far as I can tell. Was this type itself responsible for overflow?

One concern making this `size_t` is that `width` and `height` are passed directly to e.g. `cairo_image_surface_create(format, int width, int height)`.

src/Image.cc
```diff
@@ -799,8 +800,26 @@ void Image::jpegToARGB(jpeg_decompress_struct* args, uint8_t* data, uint8_t* src
 cairo_status_t
 Image::decodeJPEGIntoSurface(jpeg_decompress_struct *args) {
   cairo_status_t status = CAIRO_STATUS_SUCCESS;
-  
-  uint8_t *data = new uint8_t[naturalWidth * naturalHeight * 4];
+
+  size_t stride = naturalWidth * 4;
+  if (naturalWidth != 0 && stride / naturalWidth != 4) {
```
zbjornson: We've got a host of these unchecked muls... might be worth copying a version of this function so it's easier to continue addressing them later.

https://github.com/nodejs/node/blob/master/src/util-inl.h#L297-304

(not sure if that link works, github is acting up. Line 297-304.)

src/Image.h
LinusU: If the image is `32798x32798` it will overflow and allocate way to little bytes since:

Screenshot 2019-03-16 at 10 12 57

Good point about passing to cairo, I think that we already check for some specific value since cairo only accepts a maxiumum width or height of `2^15 - 1` (I think, ref: https://github.com/Automattic/node-canvas/blob/3d81e0ae45ad205540060d3f58ebad463f9d89a0/src/Image.cc#L610-L613)

Okay, I just realized that the image I'm loading has a larger width & height than that, maybe we should just check that before even setting `naturalWidht` and `naturalHeight` at all 🤔

zbjornson: I meant to propose: keep the members as `int` (or `unsigned`) since that's the largest Cairo can take, and only cast to a larger type where you multiply them and risk overflow.

zbjornson:
@LinusU do you still want to move forward with this since #1385 was merged?

LinusU:
Hmm, I don't know 😄 

If Cairo is using `int` I don't really think that there is much we can gain by changing anything, let's just keep it as it is?

zbjornson:
Makes senses to me! 😃

PR #1385, opened 2019-03-16
Guard JPEG width/height against maximum supported

LinusU:


src/Image.cc
```diff
@@ -10,6 +10,11 @@
 #include <node_buffer.h>
 #include "Util.h"
 
+/* Cairo limit:
+  * https://lists.cairographics.org/archives/cairo/2010-December/021422.html
+  */
+static constexpr const size_t canvas_max_side = (1 << 15) - 1;
```
zbjornson: Don't need `const` with `constexpr` (`constexpr` is enough).

This can also definitely be an `int` instead of `size_t`. Anywhere it's compared against a larger type, it'll get implicitly promoted. (Think the only case of that is when reading a jpeg.)

(I'm only dwelling on the dimension types out of interest of minimizing the number of conversions.)

src/Image.cc
```diff
@@ -604,17 +609,14 @@ Image::loadGIFFromBuffer(uint8_t *buf, unsigned len) {
...
   }
 
-  width = naturalWidth = gif->SWidth;
-  height = naturalHeight = gif->SHeight;
-
-  /* Cairo limit:
-   * https://lists.cairographics.org/archives/cairo/2010-December/021422.html
-   */
-  if (width > 32767 || height > 32767) {
+  if (static_cast<size_t>(gif->SWidth) > canvas_max_side || static_cast<size_t>(gif->SHeight) > canvas_max_side) {
```
zbjornson: If you're okay with implicit conversions, you don't need the `static_cast`s here

LinusU:
@zbjornson updated, care to review again? ☺️

Issue #1389, opened 2019-03-20
Issues for URLs with 302 redirect


loretoparisi:
There are some issues when setting a `Image.src` url that actually does a 302 HTTP redirect
For the images being downloaded the api `ClientRequest.type.get` returns an error

```
 [Wed Mar 20 2019 12:48:48 GMT+0100 (CET)] trapped error code: message:Server responded with 302 stack:Error: Server responded with 302
    at ClientRequest.type.get.res (/node_modules/canvas/lib/image.js:52:28)
```

This happens for both

```javascript
loadImage( imageURL )
    .then((image) => {
        console.log("---->", image);
    })
```

and 

```javascript
// feed images
        const img = new Image()
        img.onload = () => {
            const canvas = createCanvas(width, height);
            const ctx = canvas.getContext('2d');
            ctx.drawImage(img, 0, 0, width, height);
        }
        img.onerror = err => { throw err }
        img.src = imageURL;
```

PR #1390, opened 2019-03-20
Don't rely strictly on code 200 for success

InventivetalentDev:
Thanks for contributing!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 (Unreleased)
 ==================
 ### Changed
+* Changed Image download success check to rely on a success-code range instead of strictly code 200
 ### Added
 ### Fixed
 
--- a/lib/image.js
+++ b/lib/image.js
@@ -48,7 +48,7 @@ Object.defineProperty(Image.prototype, 'src', {
 
         const type = /^\s*https:\/\//.test(val) ? https : http
         type.get(val, res => {
-          if (res.statusCode !== 200) {
+          if (res.statusCode < 200 || res.statusCode > 230) {
             return onerror(new Error(`Server responded with ${res.statusCode}`))
           }
           const buffers = []


loretoparisi:
@InventivetalentDev could this strict HTTP Code handling related to https://github.com/Automattic/node-canvas/issues/1389?
Thank you.

InventivetalentDev:
@loretoparisi Yeah, redirects shouldn't be throwing errors either... Not sure if the internal request handling supports redirects though, so I'm not gonna change anything regarding that for now.

loretoparisi:
@InventivetalentDev so the problem is the internal module making HTTP requests?

loretoparisi:
@zahhakrnch yes I get your point. So in the meanwhile I just did like this

```javascript
self.download(pathOrUrl, temp)
                        .then(path => {
                            // feed images
                            const img = new Image()
                            img.onload = () => {

                                // create Canvas from Image
                                const width = self._options.model.imageWidth;
                                const height = self._options.model.imageHeight;
                                const canvas = createCanvas(width, height);
                                ctx.drawImage(img, 0, 0, width, height);
                            }
                            img.onerror = err => { throw err }
                            img.src = path;
                        })
                        .catch(error => {
                            return reject(error);
                        })
```

where

```javascript
this.download = (uri, filename) => {
                    request.head(uri, (err, res, body) => {
                        request(uri).pipe(fs.createWriteStream(filename))
                        .on('close', function () {
                            return resolve(filename);
                        })
                        .on('error', function (error) {
                            return reject(error);
                        });
                    });
                });
```

using `request` that will handle redirects seamless.

LinusU:
This was addressed in #1398, please let me know if there is anything more missing :)

loretoparisi:
@LinusU just to be clear, this means that there is no need anymore to manually fetch so now I can just do ` img.src = path;`?

LinusU:
Yeah, it should handle redirects and all 2xx status codes now!

PR #1393, opened 2019-03-22
Fix ES6 syntax usage for IE

mhuggins:
IE11 throws an error due to the object literal, arrow function, and string interpolation syntax used by `browser.js`.  This fixes those usages so that the canvas package can run properly.

Screen Shot 2019-03-22 at 11 01 09 AM

I tested these changes locally by directly applying these changes within my project's node_modules folder, and it fixed all IE errors, allowing the page to load.

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -5,7 +5,7 @@ const parseFont = require('./lib/parse-font')
 exports.parseFont = parseFont
 
 exports.createCanvas = function (width, height) {
-  return Object.assign(document.createElement('canvas'), { width, height })
+  return Object.assign(document.createElement('canvas'), { width: width, height: height })
 }
 
 exports.createImageData = function (array, width, height) {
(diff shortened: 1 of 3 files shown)


browser.js
```diff
@@ -19,16 +19,16 @@ exports.createImageData = function (array, width, height) {
...
 
     function cleanup () {
       image.onload = null
       image.onerror = null
     }
 
-    image.onload = () => { cleanup(); resolve(image) }
-    image.onerror = () => { cleanup(); reject(new Error(`Failed to load the image "${src}"`)) }
+    image.onload = function () { cleanup(); resolve(image) }
+    image.onerror = function () { cleanup(); reject(new Error('Failed to load the image ' + src + '"')) }
```
thisconnect: missing double quote "
should be `image "' +`

mhuggins: @thisconnect Fixed, thank you!

mhuggins:
@zbjornson I fixed the quote issue that was called out and rebased against master.

LinusU:
Thanks!

PR #1398, opened 2019-04-01
Use simple-get for fetching images

### LINKED ISSUE #1389: Issues for URLs with 302 redirect

LinusU:
Thanks for contributing!

This should fix #1389 and replace #1390

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 (Unreleased)
 ==================
 ### Changed
+* Use [simple-get](https://github.com/feross/simple-get) for fetching images, now supports redirects (#1398)
 ### Added
 ### Fixed
 
--- a/lib/image.js
+++ b/lib/image.js
@@ -12,8 +12,7 @@
 
 const bindings = require('./bindings')
 const Image = module.exports = bindings.Image
-const http = require("http")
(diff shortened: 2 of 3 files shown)


LinusU:
I used simple-get because it was the absolute smallest library out of those I looked at, and I've used it before with great success.

Screenshot 2019-04-01 at 18 27 59

(ref: https://www.npmtrends.com/got-vs-needle-vs-request-vs-simple-get-vs-superagent-vs-axios-vs-node-fetch)

Also have a branch up using `got` here: https://github.com/Automattic/node-canvas/compare/master...LinusU:got

PR #1401, opened 2019-04-05
lazy-load simple-get

zbjornson:
Appears to save ~1.5-2 MB of memory and improves startup speed.

Given that the ability to use a URL for an image source was added only recently, I'm guessing it's not super widely used, so I think it makes sense to lazy-load it.

- [ ] Have you updated CHANGELOG.md? - no, not notable/no release since simple-get was added

### DIFF
--- a/lib/image.js
+++ b/lib/image.js
@@ -12,8 +12,10 @@
 
 const bindings = require('./bindings')
 const Image = module.exports = bindings.Image
-const get = require('simple-get')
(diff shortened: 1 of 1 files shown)


zbjornson:
@LinusU are you okay with this change?

With this and potentially #1393 I'd like to propose a release be made, since Node 12 is out.

LinusU:
Sounds good 👍

PR #1402, opened 2019-04-05
adds crossOrigin attribute to loadImage

thisconnect:
without that using the image later with webgl errors
 DOMException: Failed to execute 'texImage2D' on 'WebGLRenderingContext': The image element contains cross-origin data, and may not be loaded.

Not sure if there is any reason why to not have this attribute. Please let me know

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -21,8 +21,8 @@ exports.createImageData = function (array, width, height) {
 exports.loadImage = function (src) {
   return new Promise((resolve, reject) => {
     const image = document.createElement('img')
-    image.crossOrigin = "anonymous"
-    
+    image.crossOrigin = 'anonymous'
+
     function cleanup () {
       image.onload = null
       image.onerror = null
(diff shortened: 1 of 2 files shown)


thisconnect:
sorry I dont know how to correctly write a test for this issue in node-canvas project. I mainly open the PR to discuss a possible change

zbjornson:
Thanks for the PR! I'm not sure if this should be hard-coded; some users might want "use-credentials" and others might not want CORS at all. Maybe we could add an optional second argument that is an object with a `crossOrigin` property to let users set it. @LinusU or @chearon might have a preference.

LinusU:
> Maybe we could add an optional second argument that is an object with a crossOrigin property to let users set it. 

This would be great, maybe even implement it as such:

```diff
- exports.loadImage = function (src) {
+ exports.loadImage = function (src, options) {
    return new Promise((resolve, reject) => {
-     const image = document.createElement('img')
+     const image = Object.assign(document.createElement('img'), options)
```

That way any properties that you can set on an HTMLImageElement can be passed 🙌

zbjornson:
@thisconnect That looks good! Can you please update typings/index.d.ts to reflect that new option? (It can just be `options?: any` I think.)

thisconnect:
@zbjornson  I added `options?: any` please check

PR #1415, opened 2019-05-01
Properly check return value from Set and Call

LinusU:
Thanks for contributing!

### DIFF
--- a/src/Backends.cc
+++ b/src/Backends.cc
@@ -14,6 +14,5 @@ void Backends::Initialize(Local<Object> target) {
   PdfBackend::Initialize(obj);
   SvgBackend::Initialize(obj);
 
-  Local<Context> v8ctx = Nan::GetCurrentContext();
-  target->Set(v8ctx, Nan::New<String>("Backends").ToLocalChecked(), obj);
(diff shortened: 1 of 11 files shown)


LinusU:
Blocked on https://github.com/nodejs/nan/issues/851

src/Backends.cc
```diff
@@ -14,6 +14,5 @@ void Backends::Initialize(Local<Object> target) {
   PdfBackend::Initialize(obj);
   SvgBackend::Initialize(obj);
 
-  Local<Context> v8ctx = Nan::GetCurrentContext();
-  target->Set(v8ctx, Nan::New<String>("Backends").ToLocalChecked(), obj);
+  Nan::Set(target, Nan::New<String>("Backends").ToLocalChecked(), obj).ToChecked();
```
LinusU: todo: replace `ToChecked()` call that doesn't use the return value with `Check()`

zbjornson:
I think this can land now with Nan 2.14.0. :shipit:

src/Backends.cc
LinusU: Fixed 👌

LinusU:
Rebased on master and upgraded nan 🚀

PR #1421, opened 2019-05-11
Clarify toBuffer('raw') byte ordering

### LINKED ISSUE #1416: toBuffer() returns bgra
## Issue or Feature
I'm seeing something weird with the `toBuffer` method: The documentation says canvas.toBuffer('raw') returns `argb` data, but it is actually returning `bgra`.

That's not a big deal: once I found where this came from, I simply had to change a parameter in my ffmpeg command... But maybe the doc could be updated?

By the way, thanks a lot. This tool is awesome.

## Steps to Reproduce
```js
const { createCanvas } = require('canvas');

const canvas = createCanvas(2, 2);

ctx.fillStyle = 'red';
ctx.fillRect(0, 0, 1, 1);
ctx.fillStyle = 'green';
ctx.fillRect(1, 0, 1, 1);
ctx.fillStyle = 'blue';
ctx.fillRect(0, 1, 1, 1);
ctx.fillStyle = 'black';
ctx.fillRect(1, 1, 1, 1);

const buffer = canvas.toBuffer('raw');

console.log(buffer);
// expected (argb): ff,ff,0,0, ff,0,80,0, ff,0,0,ff, ff,0,0,0
// received (bgra): 0,0,ff,ff, 0,80,0,ff, ff,0,0,ff, 0,0,0,ff
```

## Environment
node-canvas 2.4.1
node 10.0 on Mac OS X 10.14

zbjornson:
Fixes #1416

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -102,8 +102,9 @@ export class Canvas {
 	toBuffer(mimeType: 'application/pdf', config?: PdfConfig): Buffer
 
 	/**
-	 * Returns the raw ARGB data without encoding in native-endian byte order,
-	 * top-to-bottom.
+	 * Returns the unencoded pixel data, top-to-bottom. On little-endian (most)
+	 * systems, the array will be ordered BGRA; on big-endian systems, it will
+	 * be ARGB.
 	 */
 	toBuffer(mimeType: 'raw'): Buffer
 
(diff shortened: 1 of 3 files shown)


types/index.d.ts
```diff
@@ -102,8 +102,8 @@ export class Canvas {
 	toBuffer(mimeType: 'application/pdf', config?: PdfConfig): Buffer
 
 	/**
-	 * Returns the raw ARGB data without encoding in native-endian byte order,
-	 * top-to-bottom.
+	 * Returns the raw ARGB data without encoding, top-to-bottom. On big-endian
+	 * platforms, the data will instead be BGRA.
```
legraphista: This comment is a bit confusing.
Shouldn't `LE` be `BGRA`, and `BE` be `ARGB`?
As I read the text, it gives me the impression that it's the other way around

zbjornson: Good catch! thank you

PR #1423, opened 2019-05-11
Fix crash when resurfacing SVG canvases

### LINKED ISSUE #1380: Segfault when resizing and rendering SVG canvas
We recently upgraded from 1.x to 2.x; now with the new version, the node process terminates with the following error when rendering SVG:

```
node(45011,0x10d46d5c0) malloc: *** error for object 0x102332710: pointer being freed was not allocated
node(45011,0x10d46d5c0) malloc: *** set a breakpoint in malloc_error_break to debug
```

Other formats (PNG, PDF) work perfectly fine. I can also render an empty canvas as SVG; the error seems to related to the contents / issues rendering commands.

I did what the error message suggested, and came up with the following stacktrace. The last relevant frame here seems to be #15 ``Closure::writeVec [...] at closure.h:29``; then the actual crash happens somewhere inside some stdlib routines for string processing.

Stacktrace: https://gist.github.com/Tharit/670228ebc114c97231ff6f71368d5696

It was created locally on OSX Mojave using node v8.9.4 and a debug build of node-canvas v2.3.1
I also tried node v10.15.3 in combination with node-canvas v2.3.1 and v2.0.0, to make sure the error was not related to the node version, or see if it was introduced recently. I could reproduce the crash with ALL versions.

The same crash is observed on our debian based live systems.

We were using a really old version of node-canvas before (v1.3.0), so at the moment I can only say that the issue appeared between v1.3.0 and v2.0.0, which is probably not super helpful.

I will try to see if I can isolate the issue down to a specific drawing command, but that could take some time. In the meantime, if there is some other information that I could provide please let me know.

zbjornson:
Fixes #1380

---

The backend code is sorely in need of more cleanup. Right now changing either the width or height causes the backend to be recreated twice (and the common case of changing both width and height thus causes four recreations).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 * Fix crash when changing canvas width/height while `fillStyle` or `strokeStyle`
   was set to a `CanvasPattern` or `CanvasGradient` (#1357).
+* Fix crash when changing width/height of SVG canvases (#1380).
 
 2.5.0
 ==================
--- a/src/backend/SvgBackend.cc
+++ b/src/backend/SvgBackend.cc
@@ -3,6 +3,7 @@
 #include <cairo-svg.h>
 #include "../Canvas.h"
 #include "../closure.h"
+#include <cassert>
 
 using namespace v8;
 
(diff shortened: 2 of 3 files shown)


asturur:
one way to handle the resurfacing could be to recreate the surface only the first time it get used. I gave a stab at refactoring the code to do that but i realized i do not know the code enough to handle it.

PR #1432, opened 2019-06-04
Update package.json typings field

diesal11:
'typings' in package.json needs to point to the Declaration file. It currently points to the just the folder

As a result IDEs like VSCode will auto import from `'canvas/typings'` instead of `canvas`, This will fix that issue

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 * Fix crash when changing width/height of SVG canvases (#1380).
 * Fix crash when using `toBuffer('raw')` with large canvases (#1158).
 * Clarified meaning of byte ordering for `toBuffer('raw')` in readme. (#1416)
-* Fixed package.json Typings field (#1432)
+* Fix package.json Typings field to point to Declaration file (#1432)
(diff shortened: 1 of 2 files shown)


zbjornson:
Are you using Typescript 3.5.1 by chance? It looks like https://github.com/microsoft/TypeScript/commit/5d188a8c68d1a54d6933ad0be5e1375b7be3e4db#diff-0049ad7158cd6170295446b6b60cb8a7 perhaps inadvertently changed the behavior; "index.d.ts" used to be [implied](https://github.com/microsoft/TypeScript/commit/5d188a8c68d1a54d6933ad0be5e1375b7be3e4db#diff-0049ad7158cd6170295446b6b60cb8a7L1039). I'm good with this change (being explicit is good) but if the above is true, then I'm sure lots of packages will break once 3.5.1 becomes mainstream.

zbjornson:
Hm, no. I can't reproduce the issue you described with typescript 3.5.1 either.

diesal11:
Nope, my VSCode & project are currently on TS version 3.4.5. This may have been an error specific to VSC? I just noticed the incorrect imports and this seemed to fix the problem :)

PR #1455, opened 2019-07-08
Ignore maxWidth in fillText and strokeText if it is undefined

### LINKED ISSUE #1454: fillText fails if maxWidth set undefined
When I explicitly set the `maxWidth` option to `undefined` in `fillText`, the text doesn't render.

## Steps to Reproduce

```typescript
import { createCanvas } from 'canvas';
ctx.fillText('You can read this!', 50, 50);
ctx.fillText('You cannot read this!', 50, 50, undefined);
```

## Your Environment
* Version of node-canvas: 2.5.0
* Environment: Node v11.13, MacOS 10.14

Hakerh400:
Fixes #1454

As it turned out that Chrome's behavior is correct, we can now safely update our implementation. 

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,9 +2,13 @@ language: node_js
 install:
   - npm install --build-from-source
 node_js:
+  - '12'
   - '10'
   - '8'
   - '6'
+matrix:
+  allow_failures:
+    - node_js: '12'
 addons:
   apt:
     sources:
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,6 @@ language: node_js
 install:
   - npm install --build-from-source
 node_js:
-  - '12'
   - '10'
   - '8'
   - '6'
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 ### Changed
 ### Added
 ### Fixed
+* Ignore `maxWidth` in `fillText` and `strokeText` if it is undefined
 
 2.6.0
 ==================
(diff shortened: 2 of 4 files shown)


Hakerh400:
The Travis failure is not related to this PR (it also fails on master with the same error).

Hakerh400:
Can't build locally at the moment, so I'm not sure if it would reproduce the crash. But I think it is a bug with cairo pattern allocation/deallocation, which was partially fixed by #1422, but after new V8's GC implementation in Node.js v12.5.0 the order of destructing V8 objects changed, revealing the bug (the bug was probably present earlier too, but didn't consistently cause crash).

Hakerh400:
Confirmed that the bug is with cairo pattern deallocation.
The following change makes the test successfull (see [Travis build](https://travis-ci.org/Hakerh400/node-canvas/builds/557793617)):

```diff
diff --git a/index.js b/index.js
index 3c9c826..0a4b177 100644
--- a/index.js
+++ b/index.js
@@ -11,8 +11,24 @@ const JPEGStream = require('./lib/jpegstream')
 const DOMMatrix = require('./lib/DOMMatrix').DOMMatrix
 const DOMPoint = require('./lib/DOMMatrix').DOMPoint

+const set = new Set()
+
+CanvasRenderingContext2D.prototype.createLinearGradient = ((f) => function (...args) {
+  const a = f.apply(this, args)
+  set.add(a)
+  return a
+})(CanvasRenderingContext2D.prototype.createLinearGradient)
+
+CanvasRenderingContext2D.prototype.createPattern = ((f) => function (...args) {
+  const a = f.apply(this, args)
+  set.add(a)
+  return a
+})(CanvasRenderingContext2D.prototype.createPattern)
+
 function createCanvas (width, height, type) {
-  return new Canvas(width, height, type)
+  const canvas = new Canvas(width, height, type)
+  set.add(canvas)
+  return canvas
 }
```

Of course this patch is not a good solutions, as it keeps references to all patterns to prevent them from deallocating. The real issue is, I think, related to canvas resurfacing. It deserves more investigation and a new PR, because it is not relevant to this PR.

zbjornson:
Ah thanks for digging into that. When I was fixing segfaults a few weeks ago I left some existing code in the destructors that looked wrong and that I expected to cause crashes because for the life of me I couldn't get the destructors to run, and wanted the noisy crashes to reveal if they ever did start running. I'll try to repro it again with that in mind.

PR #1457, opened 2019-07-13
fix assertion failure in Node 12

zbjornson:
These are destroyed in their class destructors. Destroying them here is a double-free.

See https://github.com/Automattic/node-canvas/pull/1455#issuecomment-510852554

I want to mull this over a little bit before merging... @Hakerh400 lmk if you have thoughts as well please.

This is the first time I've witnessed the destructors running, so if something indeed changed in v8/Node.js' GC in 12.x, that would make sense that we're only now seeing a double-free.

Minimal test case:

```js
// run with --expose-gc
const {createCanvas, loadImage} = require(".");

const canvas = createCanvas(100, 200);

loadImage(`./test/fixtures/checkers.png`).then(img => {
	const pattern = ctx.createPattern(img, 'repeat');
	ctx.fillStyle = pattern;
	canvas.width = 200; // cause canvas to resurface
	process.nextTick(gc);
});
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,9 +6,6 @@ node_js:
   - '10'
   - '8'
   - '6'
-matrix:
-  allow_failures:
-    - node_js: '12'
 addons:
   apt:
     sources:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 ### Fixed
 * Ignore `maxWidth` in `fillText` and `strokeText` if it is undefined
+* Fix crash (assertion failure) in Node 12.x when patterns or gradients are used
 
 2.6.0
 ==================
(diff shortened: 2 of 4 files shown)


Hakerh400:
The change looks obviously correct, but I'm a bit confused now: do pattern destructors run in Node.js below v12.x? If yes, why don't we observe crashes? If not, does it mean we have a memory leak after this change?

zbjornson:
That confused me also. I can repro the crash using the above code with Node.js v10.15.0, and I don't see signs of a leak with this PR applied. I can also repro the crash going back to a Dec 2018 commit (didn't try earlier than that). So I guess the mystery is why this wasn't discovered much earlier.

zbjornson:
@LinusU @chearon unless you have any objections, I'll merge this in a few days.

LinusU:
Sorry for late response, been away on long vacation...

PR #1509, opened 2020-01-08
Updating Types

jpinkster:
There are two missing types that I am requesting to add. `jpg` support and on the NodeCanvasRenderingContext2D

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -78,7 +78,7 @@ export class Canvas {
 	/** Constant used in PNG encoding methods. */
 	readonly PNG_FILTER_PAETH: number
 
-	constructor(width: number, height: number, type?: 'pdf'|'svg')
+	constructor(width: number, height: number, type?: 'pdf'|'svg'|'jpg')
 
 	getContext(contextId: '2d', contextAttributes?: NodeCanvasRenderingContext2DSettings): NodeCanvasRenderingContext2D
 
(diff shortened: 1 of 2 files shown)


types/index.d.ts
```diff
@@ -212,6 +212,10 @@ declare class NodeCanvasRenderingContext2D extends CanvasRenderingContext2D {
 	createPattern(...args: any[]): NodeCanvasCanvasPattern
 	createLinearGradient(x0: number, y0: number, x1: number, y1: number): NodeCanvasCanvasGradient;
 	createRadialGradient(x0: number, y0: number, r0: number, x1: number, y1: number, r1: number): NodeCanvasCanvasGradient;
+	/**
+	 * An additional method .addPage() is then available to create multiple page PDFs
+	 */
+	addPage(width?: number, height?: number): void
```
zbjornson: Oops, good catch. Can you delete the incorrect declaration from line 130-134 (I had it on the Canvas instance instead of the rendering context) please? You could also copy the text from it that describes the default width/height values.

types/index.d.ts
```diff
@@ -78,7 +78,7 @@ export class Canvas {
 	/** Constant used in PNG encoding methods. */
 	readonly PNG_FILTER_PAETH: number
 
-	constructor(width: number, height: number, type?: 'pdf'|'svg')
+	constructor(width: number, height: number, type?: 'pdf'|'svg'|'jpg')
```
zbjornson: `'jpg'` isn't a valid Canvas type, just `'pdf'` and `'svg'`. Anything else is ignored and creates an image canvas. If you want you could add `'image'` here to maybe improve clarity...

https://github.com/Automattic/node-canvas/blob/37d556c3cd9704d8acc0071d991e4bad4f69e10e/src/Canvas.cc#L105-L110

types/index.d.ts
jpinkster: Updated!

types/index.d.ts
jpinkster: Ahh thanks for that clarity! Updated

eventualbuddha:
Could you do a release to include this? Thanks!

PR #1539, opened 2020-03-21
Protected surfaces

piranna:
This abstract and protect the actual Cairo surfaces, making them an implementation detail of the backends, preventing to be created or destroyed from outside. This allow for greater isolation of concepts, allowing to access to the actual Cairo surface just only by using the `getSurface()` method, preventing any other direct access.


piranna:
Can this be reviewed and merged?

piranna:
Code sync'ed with latest `master` branch and all tests are fixed and passing in all platforms. Can this pull-request be reviewed and merged? :-)

piranna:
Can this be reviewed and merged?

src/Canvas.cc
```diff
@@ -914,8 +914,6 @@ Canvas::resurface(Local<Object> canvas) {
   Nan::HandleScope scope;
   Local<Value> context;
 
-  backend()->recreateSurface();
```
LinusU, line 917: Would you mind explaining this change? I'm not to familiar with this part of the code 😅

LinusU (commented):
Sorry for not getting to this earlier! Just one small comment, otherwise it looks very good 👍

piranna:
> otherwise it looks very good

Thank you! :-)

src/Canvas.cc
piranna: I don't remember the exact details, but reviewing the PR, seems to me like a bug, because `recreateSurface()` it's not being called anywhere else...

LinusU: Maybe best to add it back?

piranna: Maybe, or test it first as is, to be sure if it crash or fails.

chearon:
Backends have been removed

PR #1540, opened 2020-03-21
Abstract `Backend` class for screen backends with double buffering support

piranna:
Thanks for contributing!


piranna:
Can this be reviewed and merged? It only abstracts the creation of screen-based drawing backends allowing to be as basis to draw directly on Linux framebuffer of X11 o Windows GDI or macOS Cocoa, there's no conflicts at all.

Only thing that maybe would have some concerns are the dynamically created objects at https://github.com/Automattic/node-canvas/pull/1540/files#diff-5c8e889223e6698e689baa171f92bcd9R38-R40, but since they are running in their own short-lived thread they will not create a memory leak. Can somebody confirm it, of tell me what could be a better aproach to explicitly destroy them?

piranna:
Code sync'ed with latest `master` branch and all tests are fixed and passing in all platforms. Can this pull-request be reviewed and merged? :-)

piranna:
Can this pull-request be reviewed and merged? :-)

chearon:
We did have a backend class for a while but none of the other PRs made it in. It got removed as it was difficult to maintain. Not sure what this PR is doing, but I'd rather use a thinner abstraction if we're to support drawing to other surfaces (I don't even know if maintainers (which right now is just me) have the capacity to support that).

piranna:
The other PRs didn't do it because they got ignored for years, although I updated them and also I got contracted by third parties to implement that functionality and got used on production on embeded systems. I can do a final update if they are going to be merged.

The Backend abstraction (that I originally implemented and got merged) provides a cleaner and easier to extend and maintain Architecture. My only fault was not do the Split of Backend and ScreenBackend on the original commits.

chearon:
@piranna I am sorry about that. I have no doubt this is useful for some people, and it's cool that you got to do some contract work for it! I just don't have the capacity to maintain stuff that I don't think many people need, especially when it has such a high maintenance cost. We had one memory leak caused by the `Backend` class, for example, meanwhile the prebuilds and other stuff is a nightmare for almost everyone. I really do think it works well as a fork and I'm sorry that that doesn't work out for you, but I don't know what else to say :/

PR #1562, opened 2020-04-25
Text baseline alignment issues

asturur:
This PR tries to improve the precision with which different text baselines are calculated.
With node-canvas the alphabetic and bottom baseline is always perfect, it works good.

top, middle, ideographic and hanging are a bit off.

The tweak is a 2 part change.
In the function `getBaselineAdjustment` we copy the pangoLayout and we swap the text with a string that uses all the extent of the selected font, this improve automatically the `middle` use case.

Then a correction factor is applied to the positions that do not seem correct, this correction factor is found with trial and error.

### FIREFOX

image
image
image

### CHROME

image
image
image

### FIREFOX
image
image
image

### CHROME
image
image
image


asturur:
I also extended the tests with differenf fonts, to show this fix CAN'T work perfectly with all fonts ( middle seems an improvement anyway );

So this is the fix with the different text string use for measurement, with NO correction factor:
image

and this is with ALSO the correction factor
image

test/canvas.test.js
```diff
@@ -947,7 +947,7 @@ describe('Canvas', function () {
       assert.ok(metrics.alphabeticBaseline > 0) // ~4-5
       assert.ok(metrics.actualBoundingBoxAscent > 0)
       // On the baseline or slightly above
-      assert.ok(metrics.actualBoundingBoxDescent <= 0)
```
asturur, line 950: i tweaked the baseline offset, so i expect some text measurement to change slightly. Unsure if this is ok or not. Any thought anyone?

src/CanvasRenderingContext2d.cc
```diff
@@ -2684,8 +2692,10 @@ NAN_METHOD(Context2d::MeasureText) {
       x_offset = 0.0;
   }
 
+  // are those two line useful?
   cairo_matrix_t matrix;
   cairo_get_matrix(ctx, &matrix);
+
```
asturur, line 2698: matrix does not seem to be used in the function. Is `cairo_get_matrix` having some side effect that is necessary for the code below to work?

asturur:
@zbjornson @LinusU any comment / recomendation?

LinusU:
This is way outside my area of expertise 😅 

I'm always a bit afraid of magic numbers, but since this clearly makes it better I think it's a good change 👍 

Hopefully @zbjornson & @chearon will have some better input 😄

chearon:
You should be able to get the right X-height via [pango_font_metrics_get_strikethrough_position](https://developer.gnome.org/pango/stable/pango-Fonts.html#pango-font-metrics-get-strikethrough-position) (for the "middle" baseline).

For the other ones I guess some kind of hack is reasonable. I don't quite understand it yet and I'll try to give a fuller review later. Ideally we would extract the FreeType face and look at ideographic/hanging/etc from the font tables. But Pango doesn't use FreeType on macOS/Windows so there'd have to be platform-specific code. More reasons I should remove Pango and use lower-level libraries...

asturur:
i m willing to write more code if it does not hinder performances.
Totally looking at pango_font_metrics_get_striketrough_position asap and remove the `different text` hack i made if it works, definitely a cleaner approach.

asturur:
sorry i forgot about this PR completly.

LinusU:
@asturur no worries, if you feel that you don't have time to improve further, I'm open to merging it as is! Thank you for spending time on this project ☺️

asturur:
I'll give a stab tomorrow!

asturur:
@charon, i tried this:

```c
  PangoRectangle logical_rect;
  PangoLayout* measureLayout = pango_layout_copy(layout);
  PangoFontMetrics *metrics;
  pango_layout_set_text(measureLayout, "gjĮ測試ÅÊ", -1);
  pango_layout_line_get_extents(pango_layout_get_line(measureLayout, 0), NULL, &logical_rect);
  metrics = PANGO_LAYOUT_GET_METRICS(measureLayout);
  int strike_thickness = pango_font_metrics_get_strikethrough_thickness(metrics);
  int strike_position = pango_font_metrics_get_strikethrough_position(metrics);
  double scale = 1.0 / PANGO_SCALE;
  double ascent = scale * pango_layout_get_baseline(measureLayout);
  double descent = scale * logical_rect.height - ascent;
  double middle = ascent - (scale * strike_position) + (scale * strike_thickness / 2);
  // 0.072 is a constant that has been chosen comparing the canvas output
  // if some code change, this constant can be changed too to keep results aligned
  double correction_factor = scale * logical_rect.height * 0.072;

```

then returned middle when reqested, but the results are far worse.

with strike throught:
image

with the standard ascent+descent / 2


Do you know how to better interpret that number?

Also i think i have to try on my old macbook on this new one i see that i have the text scaling bug even after recompiling.

asturur:
no as not said.
In the case in which the context is vertically scaled this thing breaks, Probably the striketrhoug position contains the scale factor of the context inside, and i have no idea how to revert it.

asturur:
The tests aren't passing anymore. While before they passed. What could it be? is something changed in tests? I reverted and push forced to the passing commit.

src/CanvasRenderingContext2d.cc
chearon: It looks like it is used for `alphabeticBaseline` below

chearon:
So this PR should be updated to use `pango_font_metrics_get_strikethrough_position` for middle baseline, right?

asturur:
i did update it, but the results with scaling where compleltely off, so i reverted back.

```
  PangoRectangle logical_rect;
  PangoLayout* measureLayout = pango_layout_copy(layout);
  PangoFontMetrics *metrics;
  pango_layout_set_text(measureLayout, "gjĮ測試ÅÊ", -1);
  pango_layout_line_get_extents(pango_layout_get_line(measureLayout, 0), NULL, &logical_rect);
  metrics = PANGO_LAYOUT_GET_METRICS(measureLayout);
  int strike_thickness = pango_font_metrics_get_strikethrough_thickness(metrics);
  int strike_position = pango_font_metrics_get_strikethrough_position(metrics);
  double scale = 1.0 / PANGO_SCALE;
  double ascent = scale * pango_layout_get_baseline(measureLayout);
  double descent = scale * logical_rect.height - ascent;
  double middle = ascent - (scale * strike_position) + (scale * strike_thickness / 2);
  // 0.072 is a constant that has been chosen comparing the canvas output
  // if some code change, this constant can be changed too to keep results aligned
  double correction_factor = scale * logical_rect.height * 0.072;
```

The result was OK for middle when used at scale 1, but bad when used with any other scaling.
So i removed it.

BlueGhostGH:
Any updates on this? This is an issue that I'd really like to see fixed.

asturur:
i did not have time to investigate why the test started failing, i can try again rebasing the pr on a fresh branch

asturur:
This was a minor fix, wasn't a final solution for 1:1 match with canvas.

chearon:
This is still an issue but since #2602 we no longer use Pango. Let's start over with the lower-level APIs in a new PR, it should be a lot easier.

asturur:
oh wow i want to upgrade fabric to use version 4. on it.

chearon:
I would love to get your feedback! Still working on the changelog and migration guide.

PR #1574, opened 2020-05-07
Returns an Object and not a String

Steven-Debande:
In this way the retrieval of information is much easier. Especially for information about the dimensions.

Thanks for contributing!

### DIFF
--- a/lib/image.js
+++ b/lib/image.js
@@ -76,11 +76,12 @@ Object.defineProperty(Image.prototype, 'src', {
 
 // TODO || is for Node.js pre-v6.6.0
 Image.prototype[util.inspect.custom || 'inspect'] = function(){
-  return '[Image'
-    + (this.complete ? ':' + this.width + 'x' + this.height : '')
-    + (this.src ? ' ' + this.src : '')
-    + (this.complete ? ' complete' : '')
-    + ']';
+  return {
+    width: this.complete ? this.width : '',
+    height: this.complete ? this.height : '',
+    src: this.src ? this.src : '',
+    complete: this.complete ? 'complete' : ''
+  };
 };
 
 function getSource(img){


zbjornson:
>  the retrieval of information is much easier

Can you explain more please? You can access an image's `width` and `height` properties without inspecting it (`x = new canvas.Image(), x.width`).

> Custom `[util.inspect.custom](depth, opts)` functions typically return a string but may return a value of any type that will be formatted accordingly by `util.inspect()`.
> -- https://nodejs.org/dist/latest-v12.x/docs/api/util.html#util_custom_inspection_functions_on_objects

lib/image.js
```diff
@@ -76,11 +76,12 @@ Object.defineProperty(Image.prototype, 'src', {
...
-  return '[Image'
-    + (this.complete ? ':' + this.width + 'x' + this.height : '')
-    + (this.src ? ' ' + this.src : '')
-    + (this.complete ? ' complete' : '')
-    + ']';
+  return {
+    width: this.complete ? this.width : '',
+    height: this.complete ? this.height : '',
+    src: this.src ? this.src : '',
+    complete: this.complete ? 'complete' : ''
```
NightYoshi370, line 83: Why not make this a boolean?

Steven-Debande:
Here's an example a little more speak.
```js 
const { loadImage } = require('canvas');
async function test(url) {
    const img = await loadImage(url);
    console.log(img) //it sends us back "[Image:512x512 complete]"
}
test("url");
```
And the no way to retrieve the image size information.

Whereas the way I'm proposing to send the thing back gives its
```js 
const { loadImage } = require('canvas');
async function test(url) {
    const img = await loadImage(url);
    console.log(img) //it sends us back "{ width: 512, height: 512, src: '', complete: 'complete' }"
}
test("url");
```
And here we're able to retrieve the information with (`img.width` and `img.height`).

PS: Then it's possible we can retrieve the information back, but I can't retrieve it properly.

LinusU:
> And here we're able to retrieve the information with img.width and img.height.

I'm not sure exactly how you are retrieving that? 🤔 it's not valid json and it's just a string being logged, right?

Steven-Debande:
> > Et ici, nous sommes en mesure de récupérer les informations avec img.width et img.height.
> 
> Je ne sais pas exactement comment vous récupérez cela? 🤔 ce n'est pas un json valide et c'est juste une chaîne en cours d'enregistrement, non?

That returns beautiful to me is well an Object validates the `"` around its to illustrate the object to recover.

zbjornson:
I'd need more convincing to change this. Logging `{ width: 512, height: 512, src: '', complete: 'complete' }` removes any indication that it's an Image object, and those properties are already easily visible in the current inspect string.

PR #1576, opened 2020-05-09
Fix .complete for errored images

strager:
According to the HTML specification [1], HTMLImageElement#complete indicates whether or not an image is loading. It does not indicate whether or not the image successfully loaded.

When an Image fails to load, we call the onerror callback. However, Image#complete is false, despite the current request's state being 'broken' (according to the spec). This is not spec-compliant.

Because our Image implementation loads images synchronously (as soon as the src property is set, which is not spec-compliant), then the current request's state is only ever 'completely available' or 'broken' [2] (if src is set). This means that, according to the spec, Image#complete must always be true. Fix Image#complete to return true unconditionally.

[1] https://html.spec.whatwg.org/multipage/embedded-content.html#dom-img-complete
[2] https://html.spec.whatwg.org/multipage/images.html#img-req-state

Thanks for contributing!

### DIFF
--- a/src/Image.cc
+++ b/src/Image.cc
@@ -100,8 +100,7 @@ NAN_METHOD(Image::New) {
  */
 
 NAN_GETTER(Image::GetComplete) {
-  Image *img = Nan::ObjectWrap::Unwrap<Image>(info.This());
-  info.GetReturnValue().Set(Nan::New<Boolean>(Image::COMPLETE == img->state));
+  info.GetReturnValue().Set(Nan::New<Boolean>(true));
 }
 
 /*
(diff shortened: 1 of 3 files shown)


src/Image.cc
```diff
@@ -100,8 +100,7 @@ NAN_METHOD(Image::New) {
  */
 
 NAN_GETTER(Image::GetComplete) {
-  Image *img = Nan::ObjectWrap::Unwrap<Image>(info.This());
-  info.GetReturnValue().Set(Nan::New<Boolean>(Image::COMPLETE == img->state));
+  info.GetReturnValue().Set(Nan::New<Boolean>(true));
```
zbjornson, line 103: This is more correct than it was, but from the spec and testing Firefox, it looks like we want to return `false` when `src` is set but the image is still loading.

Appreciate the tests, clear commit message and references ❤️

strager: > it looks like we want to return `false` when `src` is set but the image is still loading.

Because loading is synchronous in node-canvas, an Image is never in a loading state.

zbjornson:
Oops, I don't have permission to push to your branch. Can you rebase please?

PR #1577, opened 2020-05-09
Fix crash if exception thrown from onload/onerror

strager:
If a user-provided .onload or .onerror handler on an Image throws an exception, the Node.js process crashes:

    FATAL ERROR: v8::ToLocalChecked Empty MaybeLocal.
     1: 0x94d778 node::Abort() [node]
     2: 0x94edb1 node::OnFatalError(char const*, char const*) [node]
     3: 0xad5b0d v8::Utils::ReportApiFailure(char const*, char const*) [node]
     4: 0x7f98f47799ae Image::SetSource(Nan::FunctionCallbackInfo<v8::Value> const&) [/home/strager/tmp/Projects/node-canvas/build/Release/canvas.node]
     5: 0x7f98f4774fec  [/home/strager/tmp/Projects/node-canvas/build/Release/canvas.node]
     6: 0xb395b9 v8::internal::FunctionCallbackArguments::Call(v8::internal::CallHandlerInfo) [node]
     7: 0xb39970  [node]
     8: 0xb3a1ea  [node]
     9: 0xb3aa99 v8::internal::Builtin_HandleApiCall(int, unsigned long*, v8::internal::Isolate*) [node]
    10: 0x12e89b9  [node]

Fix this issue by not requiring a return value from the .onload and .onerror functions.

Thanks for contributing!

### DIFF
--- a/src/Image.cc
+++ b/src/Image.cc
@@ -272,14 +272,14 @@ NAN_METHOD(Image::SetSource){
         argv[0] = Nan::Error(Nan::New(cairo_status_to_string(status)).ToLocalChecked());
       }
       Local<Context> ctx = Nan::GetCurrentContext();
-      Nan::Call(onerrorFn.As<Function>(), ctx->Global(), 1, argv).ToLocalChecked();
+      Nan::Call(onerrorFn.As<Function>(), ctx->Global(), 1, argv);
     }
   } else {
     img->loaded();
     Local<Value> onloadFn = Nan::Get(info.This(), Nan::New("onload").ToLocalChecked()).ToLocalChecked();
     if (onloadFn->IsFunction()) {
       Local<Context> ctx = Nan::GetCurrentContext();
-      Nan::Call(onloadFn.As<Function>(), ctx->Global(), 0, NULL).ToLocalChecked();
+      Nan::Call(onloadFn.As<Function>(), ctx->Global(), 0, NULL);
     }
   }
 }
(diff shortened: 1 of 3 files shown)


src/Image.cc
```diff
@@ -272,14 +272,14 @@ NAN_METHOD(Image::SetSource){
...
-      Nan::Call(onerrorFn.As<Function>(), ctx->Global(), 1, argv).ToLocalChecked();
+      Nan::Call(onerrorFn.As<Function>(), ctx->Global(), 1, argv);
     }
   } else {
     img->loaded();
     Local<Value> onloadFn = Nan::Get(info.This(), Nan::New("onload").ToLocalChecked()).ToLocalChecked();
     if (onloadFn->IsFunction()) {
       Local<Context> ctx = Nan::GetCurrentContext();
-      Nan::Call(onloadFn.As<Function>(), ctx->Global(), 0, NULL).ToLocalChecked();
+      Nan::Call(onloadFn.As<Function>(), ctx->Global(), 0, NULL);
```
zbjornson, line 282: Nice find, thanks for the fix! Do you know for sure if this is the correct change, vs using [`Nan::TryCatch`](https://github.com/nodejs/nan/blob/master/doc/errors.md#api_nan_try_catch) and re-throwing?

strager: > Do you know for sure if this is the correct change, vs using [`Nan::TryCatch`](https://github.com/nodejs/nan/blob/master/doc/errors.md#api_nan_try_catch) and re-throwing? (If you don't know for sure, I'll try to investigate.)

I do not know. According to my testing, the solution I implemented does propagate exceptions in JavaScript land.

mingfang: I'm trying out this PR with resemblejs and is still crashing.

zbjornson: Looks like this is safe/correct.

zbjornson:
@strager this just needs a quick rebase to land.

zbjornson:
Landed as 12e671d. Thank you!

PR #1595, opened 2020-05-28
Binary build fixes

estadtherr:
Thanks for contributing!

----

This pull request fixes a couple errors in the `node-gyp` binary build:

1. There was JPEG-related code that wasn't surrounded by `#ifdef HAVE_JPEG` blocks.
2. `CanvasRenderingContext2d.cc` was including `cairo-pdf.h` with an incorrect path (the `pkg-config` include path provides the path directly to the Cairo includes, so having `cairo/` in the include directive is incorrect).

### DIFF
--- a/src/CanvasRenderingContext2d.cc
+++ b/src/CanvasRenderingContext2d.cc
@@ -4,7 +4,7 @@
 
 #include <algorithm>
 #include "backend/ImageBackend.h"
-#include <cairo/cairo-pdf.h>
+#include <cairo-pdf.h>
 #include "Canvas.h"
 #include "CanvasGradient.h"
 #include "CanvasPattern.h"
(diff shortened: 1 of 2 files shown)


src/CanvasRenderingContext2d.cc
```diff
@@ -4,7 +4,7 @@
 
 #include <algorithm>
 #include "backend/ImageBackend.h"
-#include <cairo/cairo-pdf.h>
+#include <cairo-pdf.h>
```
zbjornson, line 7: Thanks for the fixes!

#1594 just fixed the JPEG guards. Could you rebase on master please to isolate this change?

estadtherr:
I rebased to the later `master` branch as requested, so now it just has the Cairo compile fix.

PR #1606, opened 2020-06-13
Speed up putImageData for RGBA canvases

zbjornson:
Replaces the byte-wise version with both a faster portable version and a specific version for x86/64 (SSE2).

```
# ops/sec, higher is better
      before   portable      sse2
a=0   19,024     33,935    71,080
a=FF  16,546     13,847    30,995
a=__   7,537      9,175    22,998
```

@pcordes I know you had work on this from three years ago. Any comments would be welcome. I stuck with SSE2 for now to avoid dispatching.

Ref #909 


LinusU (approved):
Neat 👍 

Didn't understand all the code, but looks good 😅

zbjornson:
(I haven't merged this yet because I found an even faster way but haven't finished implementing it.)

PR #1611, opened 2020-06-24
Render small-caps font variant

samizdatco:
- pango has a PANGO_VARIANT_SMALL_CAPS variant that can be added to the font description, but setting it doesn't cause text to be rendered in small-caps, even if the font supports it
- adding the appropriate opentype features to the layout's attributes list will use the font's small-caps glyphs if it provides them

Thanks for contributing!


src/CanvasRenderingContext2d.cc
```diff
@@ -2547,6 +2549,17 @@ NAN_SETTER(Context2d::SetFont) {
   context->state->fontDescription = sys_desc;
   pango_layout_set_font_description(context->_layout, sys_desc);
 
+  PangoAttribute *features;
+  if (strlen(*variant) > 0 && strcmp("small-caps", *variant)==0) {
+    features = pango_attr_font_features_new("smcp 1, onum 1");
+  }else{
```
chearon, line 2555: nit: could you put spaces before/after the brackets, and spaces before/after the `==` above

chearon (commented):
Thanks for this. Looks good in general. `pango_attr_font_features_new` is not available in the (old) version of Pango for Windows that we link to in the wiki and use in CI, so we might have to `#ifdef PANGO_VERSION(...` the new code.

We should also probably change CI testing to use MSYS like the prebuilds do.

src/CanvasRenderingContext2d.cc
samizdatco: gladly

samizdatco:
I realized that my initial version didn't account for state and would get out of sync with the displayed font after a save() and a restore(). So I've added a field to the state struct called `textAttributes` that holds a pointer to the layout's current `PangoAttrList` (though maybe `fontAttributes` or `layoutAttributes` would be a better name?).

test/canvas.test.js
```diff
@@ -447,29 +447,34 @@ describe('Canvas', function () {
   });
 
   it('Context2d#font=small-caps', function () {
+    if (process.platform == 'win32'){
+      // the windows version of pango (< 1.37.1) is too old to support font varients
```
zbjornson, line 451: Doesn't have to be this PR (I can do it later unless you want to do it here) but it'd be better to expose the pango version (https://github.com/Automattic/node-canvas/blob/master/src/init.cc and https://github.com/Automattic/node-canvas/blob/master/index.js#L74) and test if it's older than 1.37.1. (Prebuilds use a newer version built from source.)

samizdatco: Ah, good to know. I was wondering how to detect that directly...

chearon (commented):
Looking good, I have one comment about state management.

jeffhaack:
It looks like no progress on this in awhile.  Is it a complicated fix to add small-caps support in?

chearon:
Since #2602 we no longer use Pango. Shouldn't be terribly difficult to do in the new `FontLayout.cc`, but this no longer applies.

Issue #1014, opened 2017-09-30
Kerning of fonts like canvas.style.letterSpacing = "-3px;"


bhouston:
## Issue or Feature
I would like to be able to adjust the kerning/letter-spacing of fonts when drawing text on node-canvas.  In a browser this is possible by setting canvas.style.letterSpacing to something like "-3px".  But canvas.style I believe does not exist on node-canvas.

(BTW we love node-canvas, it is an amazing piece of software.)

## Steps to Reproduce

Interactive example of it working with browser-based Canvas:

https://jsfiddle.net/hg4pbsne/1/

Example code:
```
  canvas.style.letterSpacing = '-5px';
  ctx = canvas.getContext('2d');
  ctx.textAlign = 'center';
  ctx.textBaseline = 'middle';
  ctx.font = '4em sans-serif';
  ctx.fillText('World', can.width/2, can.height*3/4);
```

This is not possible to reproduce with node-canvas as there is no canvas.style property exposed.

## Your Environment

All versions of node-canvas.

Hjlxc:
I found this possible solution that uses measure text to compute it manually:

https://gist.github.com/talltyler/5345894

chearon:
It could be done inside canvas with [`pango_attr_letter_spacing_new`](https://developer.gnome.org/pango/unstable/pango-Text-Attributes.html#pango-attr-letter-spacing-new) and [`pango_layout_set_attributes`](https://developer.gnome.org/pango/unstable/pango-Layout-Objects.html#pango-layout-set-attributes) but I'm not sure if adding more non-standard features is a good path to go down.

JamesCarlyleClarke:
Perhaps look at https://stackoverflow.com/questions/33955754/adding-letter-spacing-in-html-canvas (my post on Stack Overflow) that proposes a different method to the link above that resolves some of the issues mentioned in the link above. I'm sure it could be adapted.

asturur:
I do not think the canvas style should influence drawing on the bitmap data.
That would throw off jsdom + node-canvas combinations i believe.

asturur:
https://wpdev.uservoice.com/forums/257854-microsoft-edge-developer/suggestions/34427839-canvas-2d-font-kerning-support

is not greatly written but if someone wants to upvote this

whatyoubendoing:
Chrome now supports letter-spacing in `v56` https://github.com/CreateJS/EaselJS/issues/872. Could be time to revisit this.

[...]

PR #1615, opened 2020-07-02
Add support for letterspacing

samizdatco:
Building on the changes in #1611 (in particular the addition of a `PangoAttrList` to the context state), adding support for tracking/letterspacing required minimal additions.

The main change is a new context attribute which I've been calling `textTracking`, both as a way to avoid confusion with the css `letter-spacing` property and as a way to signal the units that it uses. In page layout apps, ‘tracking’ is typically a positive or negative integer that expresses the additional space to add or remove in terms of 1/1000 of an em. 

Using em-based units has the nice effect of keeping the spacing proportions constant even if the font size changes. Expressing the size in thousandths is helpful since typical amounts of letterspacing to add are usually in the range of `50`-`250` (which is more awkward to express in css units: `0.05em`–`0.25em`).

As with the other `ctx.text*` attributes, the `textTracking` value is part of the context state and behaves as expected across `save()` and `restore()` calls.


Readme.md
```diff
@@ -392,6 +393,14 @@ canvas.toDataURL('image/jpeg', {...opts}, (err, jpeg) => { }) // see Canvas#crea
 canvas.toDataURL('image/jpeg', quality, (err, jpeg) => { }) // spec-following; quality from 0 to 1
 ```
 
+### CanvasRenderingContext2D#textTracking
```
LinusU, line 396: Is this present in the spec for `CanvasRenderingContext2D`, or supported by any browsers?

samizdatco: No, this would be an extension since HTML Canvas makes no provisions for setting letterspacing. See @zbjornson's summary of current browser behaviors here: https://github.com/Automattic/node-canvas/issues/1014#issuecomment-488970083

MohamadKh75:
Any updates on this?

LinusU:
Sorry for dropping the ball on this one.

It seems like there now is a `letterSpacing` attribute in the Canvas specification.

ref: https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/letterSpacing

Would you be able to tweak this PR to match that?

I think that it would be okay to only support `em` values for now if that makes it easier. If we throw an appropriate error saying that something else is not implemented at the moment.

konser80:
still waiting for this feature

chearon:
Since #2602 we no longer use Pango, so this will need to be re-done

PR #1617, opened 2020-07-07
Speed up `fillStyle=` and `strokeStyle=`

zbjornson:
I was confused why I was seeing CanvasPattern show up in CPU profiles when we never use them. Turns out it's from these setters.

Roughly 2x performance improvement.
- Don't create extra `Local<FunctionTemplate>` handles
- Don't create any `Local<FunctionTemplate>` handles if the value is a string

### DIFF
--- a/test/canvas.test.js
+++ b/test/canvas.test.js
@@ -1756,7 +1756,7 @@ describe('Canvas', function () {
     var canvas = createCanvas(2, 2);
     var ctx = canvas.getContext('2d');
 
-    ctx.fillStyle = ['#808080'];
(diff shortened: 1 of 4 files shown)


test/canvas.test.js
```diff
@@ -1756,7 +1756,7 @@ describe('Canvas', function () {
     var canvas = createCanvas(2, 2);
     var ctx = canvas.getContext('2d');
 
-    ctx.fillStyle = ['#808080'];
+    ctx.fillStyle = '#808080';
```
zbjornson, line 1759: I have no idea what this test was supposed to be doing...

Issue #1411, opened 2019-04-24
 pattern.setTransform() is not a function


ritwickdey:
```js
const pattern = this.context.createPattern(image, 'repeat');
pattern.setTransform() // `setTransform` is undefined.
// etc.
```

I checked the type definition, `setTransform` function  is there. 

Few questions.
1. Is this issue of type def file ?
2. if yes, then how I can set `patternTransform` attribute into the pattern ?

zbjornson:
`pattern.setTransform` shows up in the type definitions because we borrow the definitions from native HTMLCanvas/pattern/etc. node-canvas doesn't support that method yet unfortunately. I'll update the title to reflect that.

asturur:
if we know how it should behave i can implement it.
I wonder if there is a clear explanation somewhere

aihornmac:
@asturur I believe the behavior is defined [here](https://html.spec.whatwg.org/multipage/canvas.html#dom-canvaspattern-settransform)

samizdatco:
I made an attempt at adding this in #1623 and to my eye it seems to be working pretty well at this point:

```js
function makeCheckerboard(w, h){
  let check = createCanvas(w, h),
      ctx = check.getContext('2d'),
      bmp = ctx.createImageData(w, h);
  for (let x=0; x<w; x++) for (let y=0; y<h; y++){
    let i = y*4*w + x*4,
        clr = (x<w/2 && y<h/2 || x>=w/2 && y>=h/2) ? 255 : 0
    bmp.data.set([clr,clr,clr,255], i)
  }
  ctx.putImageData(bmp, 0, 0)
  return check
}

let quad = 400,
    checker = makeCheckerboard(quad/4, quad/4),
    canvas = createCanvas(2*quad, 2*quad),
    ctx = canvas.getContext('2d'),
    pat = ctx.createPattern(checker, 'repeat'),
    mat = new DOMMatrix();

ctx.fillStyle = pat
ctx.fillRect(0, 0, quad, quad)

mat = mat.scale(.5).rotate(15)
pat.setTransform(mat)
ctx.fillRect(quad, 0, quad, quad)

mat = mat.scale(.5).rotate(15)
pat.setTransform(mat)
ctx.fillRect(0, quad, quad, quad)

mat = mat.scale(.5).rotate(15)
pat.setTransform(mat)
ctx.fillRect(quad, quad, quad, quad)
```

[...]

PR #1623, opened 2020-07-19
Add setTransform method to CanvasPattern

samizdatco:
As noted in #1411, CanvasPattern objects should accept a transformation matrix to be applied before the canvas's CTM. The one wrinkle is that cairo's conception of a pattern matrix is that it transforms from user- to pattern-space while canvas's is the reverse. As a result, the DOMMatrix passed to the method is inverted before being applied to the cairo pattern.


asturur:
I would suggest you add a small visual example to the list of examples available here:

https://github.com/Automattic/node-canvas/blob/master/test/public/tests.js#L433

they are not official tests, but helps to compare how things go compared to browsers.

src/CanvasRenderingContext2d.cc
```diff
@@ -389,6 +389,7 @@ Context2d::fill(bool preserve) {
       cairo_set_source(_context, new_pattern);
       cairo_pattern_destroy(new_pattern);
     } else {
+      cairo_pattern_set_filter(state->fillPattern, state->patternQuality);
```
asturur, line 392: patterQuality isn't really a canvas feature.
I think, but i may be wrong, it is used to simulate imageSmoothingEnabled property.

In canvas pattern quality is always best, there is no other option.

In case this change is ok, then it should be applied also to the stroke part.

samizdatco: Now applied to the stroke as well

samizdatco:
image

SLKnutson:
Is there any timeline on getting this merged? This would be very nice to have.

kmcclellan:
@asturur, @zbjornson What's preventing this from being merged? Anything we can help with?

LinusU (approved):
Looks good to me, sorry for the delay in getting this reviewed!

wxfred:
@LinusU Need upgrading on npm, pls.

PR #1647, opened 2020-08-12
translate "sans-serif" to "sans" for Pango

### LINKED ISSUE #1643: "(process:15840): Pango-WARNING **: 20:41:16.548: couldn't load font "sans serif Not-Rotated 10px", falling back to "Sans Not-Rotated 10px", expect ugly output."
Is there a way to prevent the warning `(process:15840): Pango-WARNING **: 20:41:16.548: couldn't load font "sans serif Not-Rotated 10px", falling back to "Sans Not-Rotated 10px", expect ugly output.`? My code doesn't issue a font instruction anywhere, so seeing it complain about not being able to fulfill "sans serif Not-Rotated 10px" is very odd indeed. I even tried setting an explicit `ctx.font = "sans not-rotated 10px";` but the warning still shows up.

chearon:
Might change the default font for users on macOS and Linux, but that might be okay if they're not specifying `ctx.font`.

Fixes #1643

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 ### Changed
 ### Added
 ### Fixed
+* Fix Pango logging "expect ugly output" on Windows (#1643)
 
 2.7.0
 ==================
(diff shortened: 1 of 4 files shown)


chearon:
Good call, updated

zbjornson:
>  Error: 500 status code downloading arm64 node.lib

nothing is ever easy. 😖 That's happening consistently, haven't looked at what changed to cause it.

zbjornson:
There we go!

PR #1678, opened 2020-10-14
Added stride to the Typescript typings

### LINKED ISSUE #1561: Typescript typedef missing "stride" Canvas property
The stride property of the Canvas object is missing from the typescript typedef.
please add
	readonly stride: number

willemmulder:
The stride property is non-standard, but we use it and even mention it in the readme. So it should also be exposed in the Typescript defs.

Not updated CHANGELOG.md because it is such a minor change.

Fixes #1561

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -60,8 +60,8 @@ export class Canvas {
 	/** _Non standard._ The type of the canvas. */
 	readonly type: 'image'|'pdf'|'svg'
 
-	// This is a getter, but it's non-standard and I don't know why we export it.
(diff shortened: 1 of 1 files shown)


zbjornson (approved):
Thanks!

> we use it

Curious to know: do you use a non-default pixel format? Wondering how important that is to preserve in a brand-new Canvas implementation.

willemmulder:
Hi @zbjornson what I meant with 'we use it' is that 'the node-canvas project uses it'. So no, I personally do not actually use it and would not mind if it were to be removed 👍

PR #1710, opened 2020-11-24
update setSource in nextTick

### LINKED ISSUE #2264: canvas.Image should call onload asynchronously when src is set
## Issue or Feature

The `Image` implementation lacks an `onload` call (at least based on https://github.com/Automattic/node-canvas/blob/master/lib/image.js), making it impossible to use it "the normal way" where the onload is our signal that image data is available and now we can do something with that data.

(`onload` on the HTML side = incredibly bad, do not use. `onload` on the JS side: unfortunately still extremely necessary for `Image`, `WebSocket`, etc. etc. it never went away in JS land and we kept reinforcing it by introducing new object types that kept the `onload` and `onerror` pattern firmly alive right up to today =S)

## Steps to Reproduce

```js
const IN_BROWSER = !!globalThis.document;

let Image = globalThis.Image;
let createCanvas = (w, h) => {
  const cvs = globalThis.document.createElement(`canvas`);
  cvs.width = w;
  cvs.heigh = h;
 return cvs;
};

if (!IN_BROWSER) {
  const { createCanvas as create, Image as ImageClass } = await import(`canvas`);
  createCanvas = create;
  Image = ImageClass;
}

const img = new Image();
img.src = "./some/path.png";
img.onload = () => {
  // actually do things here, because we like idempotent code that just runs,
  // irrespective of whether it's in the browser or Node.js
  const myCanvas = createCanvas(...);
};
```
The above code won't actually do anything for `node-canvas` (but will run perfectly fine in the browser) because the shimmed `Image` has no code to trigger the `onload` "callback".

## Suggested fix

Add an `onload` property with getter/setter:

```js
Object.defineProperty(Image.prototype, 'onload', {
  get: function() { return this._onload },
  set: function(fn) {
    this._onload = fn;
    if (this.src) fn();
  }
);
```
And update `setSource` to something like 

```js
function setSource (img, src, origSrc) {
  SetSource.call(img, src)
  img._originalSource = origSrc
  img._onload?.()
}
```

Gcaufy:
When set `img.src=some_buffer` the `onload` callback function may not be called.

Demo code:
```javascript
const img = new Image();
img.src = some_image_buffer;
img.onload = () => { alert('success'); }
```

In this case, `onload` will never be called.
After I check the code below:
https://github.com/Automattic/node-canvas/blob/master/src/Image.cc#L278-L281

I change the code to:

```
const img = new Image();
img.onload = () => { alert('success'); }
img.src = some_image_buffer;
```
And this code works well.

Fixes #2264

No, I will update it after the pr was approved.

### DIFF
--- a/lib/image.js
+++ b/lib/image.js
@@ -84,12 +84,14 @@ Image.prototype[util.inspect.custom || 'inspect'] = function(){
 };
 
 function getSource(img){
-  return img._originalSource || GetSource.call(img);
+  return img._originalSource || img._src || GetSource.call(img);
 }
 
 function setSource(img, src, origSrc){
-  process.nextTick(function(){
+  clearImmediate(img._setSourceTimer);
+  img._setSourceTimer = setImmediate(() => {
     SetSource.call(img, src);
   });
+  img._src = src;
   img._originalSource = origSrc;
 }
--- a/lib/image.js
+++ b/lib/image.js
@@ -88,6 +88,8 @@ function getSource(img){
 }
 
 function setSource(img, src, origSrc){
-  SetSource.call(img, src);
+  process.nextTick(function(){
+    SetSource.call(img, src);
+  });
   img._originalSource = origSrc;
 }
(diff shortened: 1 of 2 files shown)


jimmywarting:
if `createImageBitmap(blob)` is being introduced, then `new Image()` will be discouraged, depricated or removed

PR #1748, opened 2021-02-25
Use node pre gyp v1 (as non-scoped version is deprecated)

domoritz:
The non-scoped node-pre-gyp package is deprecated and only the @mapbox scoped package will receive updates in the future.

I have not updated the changelog.


zbjornson (approved):
Thanks! I'll update the changelog.

PR #1784, opened 2021-04-15
Export NodeCanvasRenderingContext2D

HeCorr:
Hello! I'd like to propose exporting the `NodeCanvasRenderingContext2DP` type so one can create helper functions for drawing specific shapes using the current context, for the sake of organisation.

Code example:

```typescript
function drawTriangle(ctx: NodeCanvasRenderingContext2D, pointX: number, pointY: number) {
    ctx.beginPath()
    ctx.moveTo(pointX, pointY)
    ctx.lineTo(pointX - 8, pointY - 11)
    ctx.lineTo(pointX + 8, pointY - 11)
    ctx.fill()
}
```

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -128,7 +128,7 @@ export class Canvas {
 	toDataURL(mimeType: 'image/jpeg', quality: number, cb: (err: Error|null, result: string) => void): void
 }
 
-export declare class NodeCanvasRenderingContext2D extends CanvasRenderingContext2D {
+export class NodeCanvasRenderingContext2D extends CanvasRenderingContext2D {
 	/**
 	 * _Non-standard_. Defaults to 'good'. Affects pattern (gradient, image,
 	 * etc.) rendering quality.
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -128,7 +128,7 @@ export class Canvas {
 	toDataURL(mimeType: 'image/jpeg', quality: number, cb: (err: Error|null, result: string) => void): void
 }
 
-declare class NodeCanvasRenderingContext2D extends CanvasRenderingContext2D {
+export declare class NodeCanvasRenderingContext2D extends CanvasRenderingContext2D {
 	/**
 	 * _Non-standard_. Defaults to 'good'. Affects pattern (gradient, image,
 	 * etc.) rendering quality.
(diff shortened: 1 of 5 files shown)


types/index.d.ts
```diff
@@ -128,7 +128,7 @@ export class Canvas {
 	toDataURL(mimeType: 'image/jpeg', quality: number, cb: (err: Error|null, result: string) => void): void
 }
 
-declare class NodeCanvasRenderingContext2D extends CanvasRenderingContext2D {
+export declare class NodeCanvasRenderingContext2D extends CanvasRenderingContext2D {
```
zbjornson, line 131: This line causes a linter error:
```
ERROR: 131:8  strict-export-declare-modifiers  'declare' keyword is redundant here. See: https://github.com/Microsoft/dtslint/blob/master/docs/strict-export-declare-modifiers.md
```

zbjornson (changes requested):
Thanks! One quick change. If you rebase on the master branch, the tests will pass now too.

types/index.d.ts
HeCorr: If I don't export it tho, it cannot be accessed from other packages. Should I remove `declare` and use `export class` instead?

HeCorr:
Rebased as requested.

types/index.d.ts
zbjornson: From the error message, I think it should just be `export class`, you don't need `declare` if you're exporting it.

HeCorr: alright, thank you!

HeCorr:
No idea why the tests failed..

zbjornson:
Thanks HeCorr! Landed as 2bf76b508cc1b3c891ce6f244d780b2f9a7282a9 so I could fix the merge conflicts.

HeCorr:
Thank you very much!

zbjornson:
@HeCorr I might have misunderstood this PR, I don't think we should be exporting this because it's already exported under the `CanvasRenderingContext2d` alias:
https://github.com/Automattic/node-canvas/blob/d7ebfcec12682bf7a57bca7aa98ed48398adf4bd/types/index.d.ts#L216

Are you able to just use `CanvasRenderingContext2d`? If yes, I'll revert this commit. If no, can you provide info why that doesn't work please?

HeCorr:
Yes, `CanvasRenderingContext2D` seems to work just fine. Please go ahead and revert it.  :)

Sorry for the inconvenience!

zbjornson:
Thanks! No problem.

PR #1807, opened 2021-05-19
Mention dropping prebuild support for Node <12

delucis:
This PR adds an item to the CHANGELOG for the most recent (2.8.0) release indicating that prebuilds are no longer provided for Node <12.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ project adheres to [Semantic Versioning](http://semver.org/).
 2.8.0
 ==================
 ### Changed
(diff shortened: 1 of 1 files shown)


zbjornson:
Thank you. I didn't realize this would affect that many people, so I just made the prebuilds for old and unsupported Node.js releases instead.

delucis:
Oh, nice! Thank you. I actually think it’s not unreasonable to stop providing prebuilds for outdated Node versions at some point, but it probably just needs pointing out as I guess it’s a potentially breaking change.

Issue #1362, opened 2019-01-22
Text rendering fails when specific fonts are preregistered


jhuckaby:
I was testing font rendering using [Google Fonts](https://github.com/google/fonts), specifically their [official downloadable TrueType collection - 300 MB](https://github.com/google/fonts/archive/master.zip), and found that quite a lot of them failed to display correctly.  node-canvas (well, probably Pango) seems to silently fall back to something that looks like Helvetica or Arial (e.g. generic sans-serif).  I am seeing this behavior on both macOS and Linux.

Here is a specific case that is easily reproducible macOS 10.12, 10.14, and Linux CentOS 7 using both pre-compiled and source compiled versions of node-canvas 2.3.1.  The font is [Advent Pro Regular](https://fonts.google.com/specimen/Advent+Pro) and is available for free download from Google (also included in my source link below).  All you need to reproduce the bug is to call `registerFont()` with all 7 Advent Pro styles, then try to render the regular style:

```js
const { registerFont, createCanvas } = require('canvas');

registerFont( 'AdventPro-Bold.ttf', { family: "AdventProBold" } );
registerFont( 'AdventPro-ExtraLight.ttf', { family: "AdventProExtraLight" } );
registerFont( 'AdventPro-Light.ttf', { family: "AdventProLight" } );
registerFont( 'AdventPro-Medium.ttf', { family: "AdventProMedium" } );
registerFont( 'AdventPro-Regular.ttf', { family: "AdventProRegular" } );
registerFont( 'AdventPro-SemiBold.ttf', { family: "AdventProSemiBold" } );
registerFont( 'AdventPro-Thin.ttf', { family: "AdventProThin" } );

var canvas = createCanvas( 600, 100 );
var ctx = canvas.getContext( '2d' );

ctx.fillStyle = '#ffffff';
ctx.fillRect( 0, 0, canvas.width, canvas.height );

ctx.font = "normal normal 30px AdventProRegular";
ctx.fillStyle = '#000000';
ctx.fillText( "Now is the time for all good men.", 10, 50 );

require('fs').writeFileSync( "out.png", canvas.toBuffer('image/png') );
```

The result image looks like this on all my systems:

![out-bad-30px](https://user-images.githubusercontent.com/700765/51566782-63114480-1e4a-11e9-9b61-9c9a197f88a8.png)

(Note: That is **not** what Advent Pro Regular is supposed to look like.)

[...]

PR #1811, opened 2021-05-23
Add `deregisterAllFonts` method.

### LINKED ISSUE #1762: Feature Request: deregisterAllFonts
Hello @chearon, and congratulations on the node-canvas 2.7.0 release!  I really appreciate your work on this important project.  I am still having issues loading certain fonts, even on the latest version (see #1362).

I don't want to re-open that old issue, but I am rather writing to ask about the `deregisterAllFonts()` method that we talked about back in 2019:

> > I'm really starting to think we should just implement that deregisterAllFonts function, and I just call it between every single use. What do you think?

> I can open a PR for that soon, it should be easy

Would it be possible to get this back on the TODO list?  I am just about dying for it now.  Currently I'm forking 32 parallel child processes, all loading separate isolated node-canvases, and then I am assigning children tasks based on the font path.  It's a mess, requires tons of memory, and it barely works (and I'm still hitting edge cases where it doesn't).  What I *really* need is a way to unload fonts.

Do you still think adding `deregisterAllFonts` would be relatively easy?

jhuckaby:
Thanks for contributing!

- [X] Fixes #1762 and #1362.

This PR adds a new API called `deregisterAllFonts()`, which will deregister (unload) all fonts that were previously registered with `registerFont()`.  This can be used to free up memory, and also resolve font conflicts.  Certain font variants clash with each other when Pango has both loaded.  This API allows a long-running process to render potentially hundreds of fonts without memory leaks or font conflicts.

See previous discussion in #1762 and #1362.

Example test script with accompanying fonts available here:
https://pixlcore.com/public/67dfda0807d530f3/node-canvas-deregister-example.zip

Tested on:

- [X] macOS 10.14.6
- [X] AWS Linux 2021
- [ ] Windows 10

### DIFF
--- a/src/Canvas.cc
+++ b/src/Canvas.cc
@@ -754,7 +754,7 @@ NAN_METHOD(Canvas::RegisterFont) {
       FontFace face;
       face.user_desc = user_desc;
       face.sys_desc = sys_desc;
-	  strcpy( (char *)face.file_path, (char *) *filePath );
+      strcpy( (char *)face.file_path, (char *) *filePath );
       font_face_list.push_back(face);
     } else {
       pango_font_description_free(user_desc);
@@ -771,11 +771,15 @@ NAN_METHOD(Canvas::RegisterFont) {
 }
 
 NAN_METHOD(Canvas::DeregisterAllFonts) {
-	// Unload all fonts from pango to free up memory
-	std::for_each(font_face_list.begin(), font_face_list.end(), [&](FontFace& f) {
-      deregister_font( (unsigned char *)f.file_path );
+  // Unload all fonts from pango to free up memory
+  bool success = true;
+  std::for_each(font_face_list.begin(), font_face_list.end(), [&](FontFace& f) {
+      if (!deregister_font( (unsigned char *)f.file_path )) success = false;
+      pango_font_description_free(f.user_desc);
+      pango_font_description_free(f.sys_desc);
     });
-	font_face_list.clear();
+  font_face_list.clear();
+  if (!success) Nan::ThrowError("Could not deregister one or more fonts");
 }
 
 /*
--- a/src/Canvas.cc
+++ b/src/Canvas.cc
@@ -79,6 +79,7 @@ Canvas::Initialize(Nan::ADDON_REGISTER_FUNCTION_ARGS_TYPE target) {
 
   // Class methods
   Nan::SetMethod(ctor, "_registerFont", RegisterFont);
+  Nan::SetMethod(ctor, "_deregisterAllFonts", DeregisterAllFonts);
 
   Local<Context> ctx = Nan::GetCurrentContext();
   Nan::Set(target,
@@ -753,6 +754,7 @@ NAN_METHOD(Canvas::RegisterFont) {
       FontFace face;
       face.user_desc = user_desc;
       face.sys_desc = sys_desc;
+	  face.file_path = (unsigned char *) *filePath;
       font_face_list.push_back(face);
     } else {
       pango_font_description_free(user_desc);
@@ -766,6 +768,14 @@ NAN_METHOD(Canvas::RegisterFont) {
   g_free(family);
   g_free(weight);
   g_free(style);
+}
+
+NAN_METHOD(Canvas::DeregisterAllFonts) {
+	// Unload all fonts from pango to free up memory
+	std::for_each(font_face_list.begin(), font_face_list.end(), [&](FontFace& f) {
+      deregister_font( f.file_path );
+    });
+	font_face_list.clear();
 }
 
 /*
--- a/src/Canvas.cc
+++ b/src/Canvas.cc
@@ -754,7 +754,7 @@ NAN_METHOD(Canvas::RegisterFont) {
       FontFace face;
       face.user_desc = user_desc;
       face.sys_desc = sys_desc;
-	  face.file_path = (unsigned char *) *filePath;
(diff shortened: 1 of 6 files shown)


src/Canvas.cc
```diff
@@ -768,6 +770,14 @@ NAN_METHOD(Canvas::RegisterFont) {
   g_free(style);
 }
 
+NAN_METHOD(Canvas::DeregisterAllFonts) {
+	// Unload all fonts from pango to free up memory
+	std::for_each(font_face_list.begin(), font_face_list.end(), [&](FontFace& f) {
+      deregister_font( (unsigned char *)f.file_path );
+    });
+	font_face_list.clear();
```
chearon, line 778: I think this is probably a memory leak. We need to free up the PangoFontDescriptions held by the FontFaces. It would be great if we did that in the `FontFace` destructor.

src/Canvas.cc
```diff
@@ -768,6 +770,14 @@ NAN_METHOD(Canvas::RegisterFont) {
   g_free(style);
 }
 
+NAN_METHOD(Canvas::DeregisterAllFonts) {
+	// Unload all fonts from pango to free up memory
```
chearon, line 774: Looks like there are some tabs throughout the PR that should be replaced with spaces.

chearon (changes requested):
I'm glad you were able to make a PR! I didn't have the time to code, but will definitely support this. I see a few things that need changes.

src/register_font.cc
```diff
@@ -260,3 +260,31 @@ register_font(unsigned char *filepath) {
...
+  success = RemoveFontResourceExA((LPCSTR)filepath, FR_PRIVATE, 0) != 0;
+  #else
+  FcConfigAppFontClear(FcConfigGetCurrent());
+  success = true;
+  #endif
+
+  if (!success) return false;
+
+  // Tell Pango to throw away the current FontMap and create a new one. This
+  // has the effect of registering the new font in Pango by re-looking up all
```
chearon, line 285: `s/registering the new font/unregistering the font`

src/register_font.cc
```diff
@@ -260,3 +260,31 @@ register_font(unsigned char *filepath) {
...
+  CFURLRef filepathUrl = CFURLCreateFromFileSystemRepresentation(NULL, filepath, strlen((char*)filepath), false);
+  success = CTFontManagerUnregisterFontsForURL(filepathUrl, kCTFontManagerScopeProcess, NULL);
+  #elif defined(_WIN32)
+  success = RemoveFontResourceExA((LPCSTR)filepath, FR_PRIVATE, 0) != 0;
+  #else
+  FcConfigAppFontClear(FcConfigGetCurrent());
+  success = true;
+  #endif
+
+  if (!success) return false;
```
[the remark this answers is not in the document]
jhuckaby: Ah, good catch.  I added basic error handling on the calling side, so this the return value is used now.

src/Canvas.cc
jhuckaby: Yup, you're right.  So, I added a `FontFace` destructor as you requested, and freed `user_desc` and `sys_desc` if they were non-null, but this resulted in a segmentation fault.  I tried to track it down but couldn't find the cause.  I ended up just freeing the memory in the `DeregisterAllFonts` function, which works fine on both my test platforms.  I hope this is acceptable.

src/Canvas.cc
jhuckaby: All fixed, thanks!

src/register_font.cc
jhuckaby: Haha yup, fixed!

jhuckaby:
Thanks for taking a look, @chearon!  I believe that I addressed all of your concerns, except for the `FontFace` destructor, which was segfaulting on me.  However, the memory leak is fixed, just in a slightly different place.

src/Canvas.cc
```diff
@@ -753,6 +754,7 @@ NAN_METHOD(Canvas::RegisterFont) {
       FontFace face;
       face.user_desc = user_desc;
       face.sys_desc = sys_desc;
+      strcpy( (char *)face.file_path, (char *) *filePath );
```
chearon, line 757: `strncpy((char *)face.file_path, (char *) *filePath, 1023)`

chearon (commented):
Thanks. Looks good, and I'm testing it out now. One final thing, then I'll merge!

src/Canvas.cc
jhuckaby: Ah, good catch.  I'll make that change tonight.  Thank you!

jhuckaby: Done!

chearon (approved):

chearon:
Hmm, the linter failed (`npm run lint`). Apparently that check runs after approval 🥴

jhuckaby:
Ah crap, I'll check it out.  I forgot about the linter. 

That should do it 😊

jhuckaby:
Awesome!  Thank you so much!  🎉

flohall:
Hello, I was very happy that this change was already released, because I also have issues with changing fonts. My project uses typescript, node.js 14 and ESM modules. The linter and my IDE (IntelliJ IDEA) cannot recognize this new method. There might be something broken around the typing. I solved this for me in this way, but it can't be intended:

```
import canvas from 'canvas';
const untypedCanvas: any = canvas as any;
untypedCanvas.deregisterAllFonts();
```

PR #1816, opened 2021-06-01
lib: check for parameter nullability

VoltrexMaster:
Thanks for contributing!

Value `null` is also a type of `object` so that could cause a type error here if passed, This PR fixes this issue.


zbjornson (approved):
Thanks!

PR #1818, opened 2021-06-01
lib: more strict number checking

VoltrexMaster:
Thanks for contributing!

In the spec, it is more recommended to use `Number.isInteger()` to check if something is a number rather than checking its primitive type as `NaN` is also a type of `number` which could lead to unexpected issues and errors.


VoltrexMaster:
Any ideas on why the checks are failing? :suspect:

PR #1820, opened 2021-06-02
lib: use strings for non-special cases

VoltrexMaster:
Thanks for contributing!

It is better to use string literals in place of regex if the provided regex does not contain any special regex characters for replace or split targets as regex construction is more costly.


zbjornson (approved):
Thanks!

PR #1824, opened 2021-06-06
tests(): add failing test for direction rtl and textAlign START/END

asturur:
Hi, this is a PR that will add failing tests to show that ctx.direction isn't handled properly when textAlign is either start or end.

I plan to open a PR where with some simple logic between direction and text align start/end we can achieve a basic support.

The test uses also some arab text ( translated from google for the word left,right,start and end) to show that there is additional complexity when handling rtl

I added on purpose the property `ctx.direction = 'ltr'` instead of the default inherit, since inherit requires the canvas to have an attribute to inspect that we don't have anyway.

https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/direction

This is how the test looks like now.
image


asturur:
@zbjornson this PR won't add anything more than tests and is done on my side unless you have some issue with those tests or you want the tests and the fix (partial probably)  in the same pr.

asturur:
Any update on those tests? do you think are useful or you prefer something different?

zbjornson (approved):
Good with me! 🍻

PR #1829, opened 2021-06-14
lib: fix formatting and remove semi-colons

VoltrexMaster:
Thanks for contributing!

Fixed the formatting, removed semi-colons and fixed some inconsistency of lib files.

### DIFF
--- a/lib/DOMMatrix.js
+++ b/lib/DOMMatrix.js
@@ -187,31 +187,31 @@ function setNumber3D(receiver, index, value) {
 }
 
 Object.defineProperties(DOMMatrix.prototype, {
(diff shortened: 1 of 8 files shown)


lib/DOMMatrix.js
```diff
@@ -187,31 +187,189 @@ function setNumber3D(receiver, index, value) {
...
+    set: function (v) {
+      return this.m42 = v
+    }
+  },
+
+  is2D: {
+    get: function () {
+      return this._is2D
+    }
+  }, // read-only
```
LinusU, line 372: Hmm, I think that this was much more readable before? 🤔 

I don't think that there is anything in `standard` that prevents us from still having this be one-liners?

zbjornson:
Thanks for the PR. I just merged another PR that made essentially the same changes on more files though, so will close this.

PR #1836, opened 2021-06-28
Add support for multi-byte font path on Windows

tignear:
Implemented a workaround to support paths containing multibyte strings on Windows. 
This should fix some of the `Could not parse font file` errors when running `registerFont`, but I'd like to change to a better implementation, so let me know. Please. 


LinusU:
@chearon would you be able to take a quick look at this ❤️

I'm afraid I'm not familiar enough with this part to be able to review it 😅

chearon (changes requested):
Thanks for the PR. Loading into memory is fine, however I think the number of changes here could be reduced a lot. Can we change it so that it doesn't mess with `get_pango_font_description` except in the way that it loads the face? I.e. some new `get_face_from_path` that returns what you have now as `face_temp`. That would save me some time reviewing!

I don't see the checks running, so that needs to be addressed too. Instead of copying all of the font files, can we just make a new `pfennigMultiByte🚀.ttf` in the same folder and load that file?

tignear:
The `buffer` (font data loaded in memory) can only be released (`UnmapViewOfFile`) after calling `FT_Done_Face`. Separating to multiple functions make resource management even more difficult.

tignear:
> Separating to multiple functions make resource management even more difficult.
resolved

kochizufan:
Are there any plan to adopt this PR to production?
I'm in trouble with this issue in my product...

zbjornson (changes requested):
This is more @chearon's wheelhouse, but found a few easy-to-fix C++ bugs.

If you don't want to have scattered `UnmapViewOfFile` calls throughout the code (risk of leaks), the other way is to use [`FT_Open_Face`](https://freetype.org/freetype2/docs/reference/ft2-base_interface.html#ft_open_face) with a [stream](https://freetype.org/freetype2/docs/reference/ft2-system_interface.html#ft_streamrec) specified in the [`FT_Open_Args`](https://freetype.org/freetype2/docs/reference/ft2-base_interface.html#ft_open_args) struct. The stream's `read` and `close` funcs would be normal read/close calls on the file.

zbjornson (commented):
Two tiny things, otherwise this looks good to me!

zbjornson (approved):
Looks great, thank you!

I can fix the conflicts in a bit if no one gets to it before me.

PR #1846, opened 2021-07-07
use classes/const/let etc

jimmywarting:
- [x] Have you updated CHANGELOG.md? yes

This conversion was done using lebab and `standard --fix`

```
lebab --replace <FILE> --transform arrow,let,includes,template,default-param,obj-shorthand,obj-method,arg-spread,arg-rest,for-each,class
```

wanted to include ESM but maybe to much breaking...

### DIFF
--- a/lib/DOMMatrix.js
+++ b/lib/DOMMatrix.js
@@ -4,21 +4,19 @@ const util = require('util')
 
 // DOMMatrix per https://drafts.fxtf.org/geometry/#DOMMatrix
 
-function DOMPoint(x, y, z, w) {
-  if (!(this instanceof DOMPoint)) {
-    throw new TypeError("Class constructors cannot be invoked without 'new'")
+class DOMPoint {
+  constructor (x, y, z, w) {
+    if (typeof x === 'object' && x !== null) {
+      w = x.w
+      z = x.z
+      y = x.y
+      x = x.x
+    }
+    this.x = typeof x === 'number' ? x : 0
+    this.y = typeof y === 'number' ? y : 0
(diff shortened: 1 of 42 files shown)


jimmywarting:
question: why do only some file lint with standardjs and not all?

lib/DOMMatrix.js
```diff
@@ -63,68 +61,441 @@ function parseTransform(tform) {
...
+  toFloat32Array() {
+    return Float32Array.from(this._values)
+  }
+
+  toFloat64Array() {
+    return this._values.slice(0)
   }
 }
 
-DOMMatrix.fromMatrix = function (init) {
```
zbjornson, line 119: Could you make these static members maybe? `static fromMatrix(init) { ... }`

zbjornson (commented):
This looks great, thank you. One suggestion to keep using modern syntax.

> why do only some file lint with standardjs and not all?

This was an attempt to gradually update the codebase to follow StandardJS formatting, but it's been so slow that it's created a lot of confusion. We should just convert everything finally.

lib/DOMMatrix.js
jimmywarting: will do

jimmywarting:
bdabed4 might be a bit have to look through... basically executed `standard --fix` on the hole project and fixed some error's manually

jimmywarting:
pump

jimmywarting:
bump

zbjornson (approved):
Thanks much! Sorry for taking forever to land this.

I just rebased to fix the merge conflicts and squashed the commits that were related.

Issue #1865, opened 2021-07-25
Context2d has no export


Tolfx:
## Issue
`Context2d` is not being exported instead `NodeCanvasRenderingContext2D` is..
This causes an issue when I want to add more methods to Context2d by using the `prototype`.

But once compiled an error happens on runtime saying `NodeCanvasRenderingContext2D` is undefined.

## Steps to Reproduce
```ts
import { NodeCanvasRenderingContext2D } from "canvas"
NodeCanvasRenderingContext2D.prototype.test= "test"; // Compiles and complains in runtime
```

But when using: 
```ts
import { Context2d } from "canvas"
Context2d.prototype.test = "test";
```
It runs without issues but compiler complains that `Module '"canvas"' has no exported member 'Context2d'`.

## Your Environment
* Typescript 4.3.5
* Node 12.14.1
* Windows 10
* canvas 2.8.0

chearon:
Not totally sure I understand this issue, but we made a lot of changes to `index.d.ts` since, and `Context2d` is gone. This worked:

```
import { CanvasRenderingContext2D } from "canvas"

declare module "canvas" {
	interface CanvasRenderingContext2D {
		test: string;
	}
}

CanvasRenderingContext2D.prototype.test = "test";
```

PR #1866, opened 2021-07-25
Context2d has no export #1865 fix

Tolfx:
Simply added class `Context2d` and extends it with `NodeCanvasRenderingContext2D` to solve issue #1865

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -128,6 +128,8 @@ export class Canvas {
 	toDataURL(mimeType: 'image/jpeg', quality: number, cb: (err: Error|null, result: string) => void): void
 }
 
+export class Context2d extends NodeCanvasRenderingContext2D {};	
+	
 export class NodeCanvasRenderingContext2D extends CanvasRenderingContext2D {
 	/**
 	 * _Non-standard_. Defaults to 'good'. Affects pattern (gradient, image,
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -80,7 +80,7 @@ export class Canvas {
 
 	constructor(width: number, height: number, type?: 'image'|'pdf'|'svg')
 
-	getContext(contextId: '2d', contextAttributes?: NodeCanvasRenderingContext2DSettings): NodeCanvasRenderingContext2D
+	getContext(contextId: '2d', contextAttributes?: NodeCanvasRenderingContext2DSettings): Context2d
 
 	/**
 	 * For image canvases, encodes the canvas as a PNG. For PDF canvases,
@@ -128,9 +128,7 @@ export class Canvas {
 	toDataURL(mimeType: 'image/jpeg', quality: number, cb: (err: Error|null, result: string) => void): void
 }
 
-export class Context2d extends NodeCanvasRenderingContext2D {};	
-	
-export class NodeCanvasRenderingContext2D extends CanvasRenderingContext2D {
+export class Context2d extends CanvasRenderingContext2D {
 	/**
 	 * _Non-standard_. Defaults to 'good'. Affects pattern (gradient, image,
 	 * etc.) rendering quality.
@@ -215,7 +213,7 @@ export class NodeCanvasRenderingContext2D extends CanvasRenderingContext2D {
 	 */
 	addPage(width?: number, height?: number): void
 }
-export { NodeCanvasRenderingContext2D as CanvasRenderingContext2D }
+export { Context2d as CanvasRenderingContext2D }
 
 declare class NodeCanvasCanvasGradient extends CanvasGradient {}
 export { NodeCanvasCanvasGradient as CanvasGradient }


LinusU:
This file references the DOM typings which should have `Context2D`, that's why ours is called `NodeCanvasRenderingContext2D ` in the first place:

```ts
/// <reference lib="dom" />
```

I'm not sure that this is the correct fix 🤔 

-----

Hmm, after reading the issue you opened, I think that the correct fix would be either to rename `NodeCanvasRenderingContext2D` to `Context2D`, or do something like:

```diff
-export class NodeCanvasRenderingContext2D extends CanvasRenderingContext2D {
+export interface NodeCanvasRenderingContext2D extends CanvasRenderingContext2D {
```

```diff
+export class Context2D implements NodeCanvasRenderingContext2D
```

Tolfx:
> This file references the DOM typings which should have `Context2D`, that's why ours is called `NodeCanvasRenderingContext2D ` in the first place:
> 
> ```ts
> /// <reference lib="dom" />
> ```
> 
> I'm not sure that this is the correct fix 🤔 
> 
> -----
> 
> Hmm, after reading the issue you opened, I think that the correct fix would be either to rename `NodeCanvasRenderingContext2D` to `Context2D`, or do something like:
> 
> ```diff
> -export class NodeCanvasRenderingContext2D extends CanvasRenderingContext2D {
> +export interface NodeCanvasRenderingContext2D extends CanvasRenderingContext2D {
> ```
> 
> ```diff
> +export class Context2D implements NodeCanvasRenderingContext2D
> ```
> 
> 

Due to not really understanding why it wasn't named `Context2D` I decided to do the way I did it. But your following method would work Indeed.

Tolfx:
I tested and honestly I think renaming it to `Context2d` would be the best solution.

Because if just implementing `Context2d` with `NodeCanvasRenderingContext2D` it would still be an issue with types.
Since if I want to add new methods to the class I need to declare it, and because `.getContext()` returns `NodeCanvasRenderingContext2D` it would not have been changed.

```ts
declare module "canvas" {
    export interface Context2d {
        hello: string;
    }
}

Context2d.prototype.hello = "hello";

let canvas = createCanvas(width, height)
let ctx = canvas.getContext('2d'); // Returns NodeCanvasRenderingContext2D and not Context2d
console.log(ctx.hello) // undefined according to typescript
```

But by just renaming it to `Context2d` this issue will be solved as well.

```diff
- export class NodeCanvasRenderingContext2D extends CanvasRenderingContext2D
+ export class Context2d extends CanvasRenderingContext2D
```
*needs to rename everything ofc*

zbjornson:
`Context2d

zbjornson:
`Context2d` is a legacy name from v0/v1 that should be removed sometime. 
https://github.com/Automattic/node-canvas/blob/d7ebfcec12682bf7a57bca7aa98ed48398adf4bd/index.js#L61

`CanvasRenderingContext2d` is the name from the HTML Canvas specification and is what's actually exported.

`NodeCanvasRenderingContext2d` is the name of our types interface. I think it was a mistake to export this (https://github.com/Automattic/node-canvas/commit/2bf76b508cc1b3c891ce6f244d780b2f9a7282a9 / #1784), it's already exported under an alias shown below. I'll see about undoing this.
https://github.com/Automattic/node-canvas/blob/d7ebfcec12682bf7a57bca7aa98ed48398adf4bd/types/index.d.ts#L216

I think what you want is this:
```ts
import {CanvasRenderingContext2D} from "canvas";
CanvasRenderingContext2D.prototype.foo = "bar";
```

Tolfx:
> I think what you want is this:
> 
> ```ts
> import {CanvasRenderingContext2D} from "canvas";
> CanvasRenderingContext2D.prototype.foo = "bar";
> ```

This could have been the cased but I think this will also resolve into a issue which I explained when it comes to declaring it. Because if I would be declaring `CanvasRenderingContext2D` again it would remove all of its methods and only leave what I specify myself to declare, which means I need to rewrite all of the methods again *(copy/paste it)*, which isn't very ethical if the package would have a major or update.

Perhaps there is a fix to it but can't really think of any at the moment.

zbjornson:
Can you provide a full example? Typescript [declaration merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html) should do what you want, if I understand what you mean.

Tolfx:
> Can you provide a full example? Typescript [declaration merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html) should do what you want, if I understand what you mean.

```ts
declare module "canvas" {
    export interface CanvasRenderingContext2D {
        hello: string;
    }
}

CanvasRenderingContext2D.prototype.hello = "hello"; // Works fine

...
ctx = canvas.getContext('2d');
ctx.beginPath() // doesn't exist anymore as for example
...
```

Also tried declaration mergin but didn't get it to work.

zbjornson:
`beginPath` is being removed because you're modifying `CanvasRenderingContext2D` instead of `NodeCanvasRenderingContext2D`. This works for me:

```ts
import {CanvasRenderingContext2D} from "canvas";

declare module "canvas" {
	interface NodeCanvasRenderingContext2D {
		hello: string;
	}
}

CanvasRenderingContext2D.prototype.hello = "hi";
CanvasRenderingContext2D.prototype.beginPath();
```

Tolfx:
But does it work while compiled. Why I renamed it to `Context2d` is because once compiled for js it doesn't return a error. Thus asking is `CanvasRenderingContext2D` a valid class that can be used once compiled. Will try once I got time.

Tolfx:
> `beginPath` is being removed because you're modifying `CanvasRenderingContext2D` instead of `NodeCanvasRenderingContext2D`. This works for me:
> 
> ```ts
> import {CanvasRenderingContext2D} from "canvas";
> 
> declare module "canvas" {
> 	interface NodeCanvasRenderingContext2D {
> 		hello: string;
> 	}
> }
> 
> CanvasRenderingContext2D.prototype.hello = "hi";
> CanvasRenderingContext2D.prototype.beginPath();
> ```

Finally tried the method and worked even compiled, so this `PR` can be closed. Thanks for the help. 👍

almeidx:
> `beginPath` is being removed because you're modifying `CanvasRenderingContext2D` instead of `NodeCanvasRenderingContext2D`. This works for me:
> 
> ```ts
> import {CanvasRenderingContext2D} from "canvas";
> 
> declare module "canvas" {
> 	interface NodeCanvasRenderingContext2D {
> 		hello: string;
> 	}
> }
> 
> CanvasRenderingContext2D.prototype.hello = "hi";
> CanvasRenderingContext2D.prototype.beginPath();
> ```

I cannot get this solution to work ever since the 2.9.0 release (which includes https://github.com/Automattic/node-canvas/commit/4a891ef22c5bdf56f3284279d490f53a8f60f396)

Issue #1782, opened 2021-04-10
Error: node-canvas was built without JPEG support


NikyPeng:
## Issue or Feature

> canvasdemo@1.0.0 dev
> node index.js

/Users/pengcheng/pc/canvas/node_modules/canvas/lib/image.js:91
  SetSource.call(img, src);
            ^

Error: node-canvas was built without JPEG support
    at setSource (/Users/pengcheng/pc/canvas/node_modules/canvas/lib/image.js:91:13)
    at Image.set (/Users/pengcheng/pc/canvas/node_modules/canvas/lib/image.js:62:9)
    at /Users/pengcheng/pc/canvas/node_modules/canvas/index.js:34:15
    at new Promise (<anonymous>)
    at loadImage (/Users/pengcheng/pc/canvas/node_modules/canvas/index.js:23:10)
    at Object.<anonymous> (/Users/pengcheng/pc/canvas/index.js:19:1)
    at Module._compile (node:internal/modules/cjs/loader:1092:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1121:10)
    at Module.load (node:internal/modules/cjs/loader:972:32)
    at Function.Module._load (node:internal/modules/cjs/loader:813:14)

## Steps to Reproduce

```javascript
const { createCanvas, loadImage } = require('canvas')
const canvas = createCanvas(200, 200)
const ctx = canvas.getContext('2d')

// Write "Awesome!"
ctx.font = '30px Impact'
ctx.rotate(0.1)
ctx.fillText('Awesome!', 50, 100)

// Draw line under text
var text = ctx.measureText('Awesome!')
ctx.strokeStyle = 'rgba(0,0,0,0.5)'
ctx.beginPath()
ctx.lineTo(50, 102)
ctx.lineTo(50 + text.width, 102)
ctx.stroke()

// Draw cat with lime helmet
loadImage('./2.jpg').then((image) => {
  ctx.drawImage(image, 50, 0, 70, 70)

  console.log('<img src="' + canvas.toDataURL() + '" />')
})
```

## Your Environment
* Version of node-canvas (output of `npm list canvas` or `yarn list canvas`): canvas@2.7.0
* Environment (e.g. node 4.2.0 on Mac OS X 10.8): noede@15.14.0  macOS Big Sur 11.1 MacBook Pro (13-inch, M1, 2020) Apple M1

tupig-7:
You should install a program named "libjpeg-turbo".

[...]

PR #1872, opened 2021-07-30
Update has_lib.js to support Apple M1 with homebrew

rwtombaugh:
Stated instructions for compiling on OS X include using homebrew: `brew install pkg-config cairo pango libpng jpeg giflib librsvg`

However on Apple M1 silicon, the compile via binding.gyp will fail to locate the above library packages. This is because on M1, homebrew places all installed libraries at `/opt/homebrew/lib`, unlike on Intel silicon, where it would place them under `/usr/local/lib`.

The simple remedy is  to add `/opt/homebrew/lib` to the SYSTEM_PATHS of `has_lib.js`

Fixes: #1782 


LinusU:
Awesome 👍

Issue #1845, opened 2021-07-07
Introduce blob's


jimmywarting:
NodeJS recently introduced Blobs into core

think this is grate cuz now we can have `canvas.toBlob(cb)`

we could also introduce a new fn that can create a ImageBitmap out of a more web/standarlized way with `createImageBitmap(blob)` instead of using the abnormal way with `loadImage(...)` or `new Image(...)` (that i think should be deprecated)

I also think `createCanvas` should go away for a web-worker related thing which is the OffscreenCanvas cunstructor

so that would also mean: we would use [OffscreenCanvas.convertToBlob](https://developer.mozilla.org/en-US/docs/Web/API/OffscreenCanvas/convertToBlob) instead of canvas.toBlob

This was what i did to feel more at home:
```js
const {
    // More correct naming convention
    createCanvas: OffscreenCanvas,
    loadImage: createImageBitmap
} = NodeCanvas;

export async function cropImage(...args) {
    /** @type {ImageBitmap} */
    const bitmap = await createImageBitmap(imagePath);
    const canvas = OffscreenCanvas(width, height);
```

fetch-blob is a good candidate for implementing support for it already

jimmywarting:
This would also help to quickly upload/download things using (node-)fetch and do other things with it

LinusU:
I think that it would be awesome to "transition" node-canvas to be an implementation of `OffscreenCanvas` instead of `Canvas`, which is more accurately what it is current anyways. Maybe even change `import { createCanvas } from 'canvas'` to `import { OffscreenCanvas } from 'canvas'` (but keep the old function as well for compatibility)?

Anything that makes us more compatible I'm 👍

jimmywarting:
>  (but keep the old function as well for compatibility)?

yea, sounds good!
but OffscreenCanvas is a constructor doe...

jimmywarting:
ctx.drawImage would have to accept a ImageBitmap class instead/also

jimmywarting:
fyi, fetch-blob is a ESM only package, so to load it from commonjs you would have to use the async `import()` instead cuz esm are async b/c of top level await

but that would be fine either way. createImageBitmap don't need to import something to know that it's a blob and can be read.
convertToBlob is async either way also. so can just lazy import it when needed instead

LinusU:
> > (but keep the old function as well for compatibility)?
> 
> yea, sounds good!
> but OffscreenCanvas is a constructor doe...

[...]

PR #1883, opened 2021-08-23
Add user agent to simple-get

GiorgioProductions:
Some sites (including Discord now) return a 403 error if no user agent is specified. Adding a user agent to the simple-get fixes the problem.

### DIFF
--- a/lib/image.js
+++ b/lib/image.js
@@ -49,7 +49,7 @@ Object.defineProperty(Image.prototype, 'src', {
 
         if (!get) get = require('simple-get');
 
-        get.concat(val, (err, res, data) => {
+        get.concat({url:val, headers: {'user-agent': 'Mozilla/5.0 (platform; rv:geckoversion) Gecko/geckotrail Firefox/firefoxversion'}}, (err, res, data) => {
           if (err) return onerror(err)
 
           if (res.statusCode < 200 || res.statusCode >= 300) {


almeidx:
From the Discord Developers server:
> At around 3:00 PDT, we became aware of an issue impacting bots' ability to gather images and other attachments from our CDN without specifying a User-Agent header. Impacted bots were experiencing 403 errors when attempting to gather files from our CDN without a User-Agent header. This change was not intentional, and has been resolved.

GiorgioProductions:
Would it still be a good idea to add a user agent?

zbjornson:
In another Node.js package I help maintain, we decided to not set a default user-agent because Node.js doesn't set a default either. We did expose the ability for users to set *any* HTTP option (including any header) though. I'm not sure how we would provide that functionality in node-canvas without significantly deviating from the HTML spec, so I'm open to adding a U-A header (but something more accurate like `node-canvas/<node-canvas-ver> node/<node-ver>`).

@LinusU @chearon ?

chearon:
> I'm not sure how we would provide that functionality in node-canvas without significantly deviating from the HTML spec

Providing a hook wouldn't be bad would it? Something like `Canvas.on("request", config => config.headers...`. That would be my first pick, followed by a hard-coded `node-canvas/<ver>` header.

jimmywarting:
I personally would like to see simple-get removed. along with Image ( #1845 )
simple-get isn't exposed in any meaningful way that it can be used somehow so, why not keep it that way?

a better proper way to load a image is via: [createImageBitmap](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/createImageBitmap)

nodejs have added support for blobs, so one way to load binary images could be to do:

```js
import { Blob } from 'node:buffer'
import { createImageBitmap }  from 'canvas/createImageBitmap.js'

const bitmap = await createImageBitmap(new Blob([ uint8 ]))
// Later paint bitmap to canvas
```

zbjornson:
> simple-get isn't exposed in any meaningful way that it can be used somehow

I'm sure you saw this already, but it's used when `img.src = "a URL"`. `createImageBitmap` doesn't currently accept a URL, and based on the number of issues opened here, quite a few people use `img.src=url`. More interesting info on the topic here: https://wiki.whatwg.org/wiki/ImageBitmap_Constructor_taking_URL

jimmywarting:
well, i know it's being used when `img.src` is set but you have no direct access to simple-get

more interesting would be if it could take a Request object instead of a url, not everything is simple get requests, others require credentials, cookie, other http methods... `createImageBitmap(new Request(...))`

or perhaps accept a ReadableStream from eg: res.body

PR #1943, opened 2021-12-08
Clean up isnan/isinf, use isfinite

zbjornson:
Cleans up funky C code and uses std::isfinite, which also compiles to fewer instructions (https://godbolt.org/z/qMT8bMTKn).


LinusU (approved):
So clean 🤩

Issue #1394, opened 2019-03-23
make module context-aware


mac-g:
## Issue or Feature
I've been experimenting with worker threads in node.js recently and noticed that the canvas module does not appear to be context-aware. In other words, the module will load/instantiate once, but as soon as another thread attempts to require canvas for it's own use, the following error is thrown:
```sh
Error: Module did not self-register.
    at Object.Module._extensions..node (internal/modules/cjs/loader.js:840:18)
    at Module.load (internal/modules/cjs/loader.js:666:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:606:12)
    at Function.Module._load (internal/modules/cjs/loader.js:598:3)
    at Module.require (internal/modules/cjs/loader.js:705:19)
    at require (internal/modules/cjs/helpers.js:14:16)
    at Object.<anonymous> (~/node_modules/canvas/lib/bindings.js:3:18)
    at Module._compile (internal/modules/cjs/loader.js:799:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:810:10)
    at Module.load (internal/modules/cjs/loader.js:666:32)
```
## Steps to Reproduce

In main thread...
```js
const {Worker} = require('worker_threads');
var worker1 = new Worker('./child-thread.js'); 
var worker2 = new Worker('./child-thread.js');
// etc.
```
In child-thread.js...
```js
var Canvas = require('canvas');
var canvas = Canvas.createCanvas(200, 200);
var ctx = canvas.getContext('2d');
// etc.
```
## References
**Node.js documentation:**
[https://nodejs.org/api/addons.html#addons_context_aware_addons](https://nodejs.org/api/addons.html#addons_context_aware_addons)
**Similar issue:**
[https://github.com/schroffl/node-lzo/pull/11](https://github.com/schroffl/node-lzo/pull/11)

## Environment
* node-canvas version **2.4.1**
* Mac OS **10.14.3**
* node version **11.12.1**

LinusU:
PRs welcome 👍

jgoralcz:
This would be an incredible addition to node-canvas. Not sure where to fully dive in and modify the code though 🤔

ruiwei:
I'm having the same issue. Any solution yet?

a-koka:
Nope. Need to fork processes and can't use worker threads.

siddarthareddyt:
[...]

PR #1981, opened 2022-02-02
Support worker env.[aka: context aware]

niyan-ly:
This `PR` removes all function template, instead, keep trace of js constructor for each individual v8::Context. To achieve this, it holds all necessary js constructor under global property `__node_canvas` in each context. It may not that graceful, but is relatively safer than type casting many void* pointer.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 ==================
 ### Changed
 ### Added
+* support nodejs worker
 ### Fixed
 * Stringify CanvasGradient, CanvasPattern and ImageData like browsers do. (#1639, #1646)
 * Add missing include for `toupper`.
--- a/binding.gyp
+++ b/binding.gyp
@@ -60,6 +60,7 @@
       'include_dirs': ["<!(node -e \"require('nan')\")"],
       'sources': [
         'src/AddonData.cc',
+        'src/lock.cc',
         'src/backend/Backend.cc',
         'src/backend/ImageBackend.cc',
         'src/backend/PdfBackend.cc',
--- a/binding.gyp
+++ b/binding.gyp
@@ -59,6 +59,7 @@
       'target_name': 'canvas',
       'include_dirs': ["<!(node -e \"require('nan')\")"],
       'sources': [
+        'src/AddonData.cc',
         'src/backend/Backend.cc',
         'src/backend/ImageBackend.cc',
         'src/backend/PdfBackend.cc',
--- a/package.json
+++ b/package.json
@@ -58,6 +58,7 @@
     "dtslint": "^4.0.7",
     "express": "^4.16.3",
     "mocha": "^5.2.0",
+    "piscina": "^3.2.0",
     "pixelmatch": "^4.0.2",
     "standard": "^12.0.1",
     "typescript": "^4.2.2"
--- a/src/AddonData.cc
+++ b/src/AddonData.cc
@@ -1,0 +1,25 @@
+#include "AddonData.h"
+
+AddonData::AddonData()
+{
+  node::AddEnvironmentCleanupHook(v8::Isolate::GetCurrent(), Dispose, this);
+}
+
+void AddonData::Dispose(void* ins)
+{
+  AddonData *data = static_cast<AddonData*>(ins);
+  data->canvas_ctor_tpl.Reset();
+  data->context2d_ctor_tpl.Reset();
+  data->context2d_dom_matrix.Reset();
+  data->context2d_parse_font.Reset();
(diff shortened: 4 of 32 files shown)


niyan-ly:
#1394

zbjornson (commented):
Thanks for starting on this! I don't think this is quite the right approach however. All static-duration data (such as `font_face_list` in Canvas.cc) has to be moved to instance data for the module, or has to not be JS objects *and* has to be thread-safe, which `std::vector` is not when there are multiple readers+writers. I think that is also the preferred approach for dealing with the `Persistent` handles, rather than storing them on a global object and retrieving them by calling into JS. https://nodejs.org/api/addons.html#context-aware-addons has the clearest info on this.

> is relatively safer than type casting many void* pointer.

Which casts are you looking at? The "receiver" (in V8 terms) should be checked in all cases. For example:
```
> x = canvas.createCanvas(50,50)
[Canvas 50x50]
> y = x.getContext('2d')
CanvasRenderingContext2D { canvas: [Canvas 50x50] }
> z = y.createLinearGradient()
CanvasGradient {}
> z.addColorStop.call({}, 4, "abc")
Uncaught TypeError: Illegal invocation
```

niyan-ly:
Yes, you are right. I've just focused on existing issues, but didn't take a complete check on the package level. I will continue this work to see if we can make it.

niyan-ly:
I've movd all persistent handle and non v8 data into a new context level store, `AddonData`. And I am wondering how should we do the test work. It seems like I should run every existing test case in multiple worker env to coverage as much code as possible.

niyan-ly:
There are some extra works to do. Font registration are not at thread level, so move `font_face_list` into `AddonData` is actually a bug, it's a sharing property.

niyan-ly:
> There are some extra works to do. Font registration are not at thread level, so move `font_face_list` into `AddonData` is actually a bug, it's a sharing property.

Fixed this issue by apply mutex lock, `font_face_list` are now shared in all threads(worker & main).

nateGarcia:
Hello! Not sure how useful this will be... But, for my use case of parallelization in some visual-regression tests using Resemble & Codeceptjs, I am still receiving the same error
```
Module did not self-register: 'frontend/tests/codecept_e2e/node_modules/.pnpm/github.com+Automattic+node-canvas@d987d4666a32c4b8_190300f15912bbbd76c9d4ab8a89e735/node_modules/canvas/build/Release/canvas.node'.
Error: Module did not self-register: 'frontend/tests/codecept_e2e/node_modules/.pnpm/github.com+Automattic+node-canvas@d987d4666a32c4b8_190300f15912bbbd76c9d4ab8a89e735/node_modules/canvas/build/Release/canvas.node'.
    at Object.Module._extensions..node (node:internal/modules/cjs/loader:1183:18)
    at Module.load (node:internal/modules/cjs/loader:975:32)
    at Function.Module._load (node:internal/modules/cjs/loader:822:12)
    at Module.require (node:internal/modules/cjs/loader:999:19)
    at require (node:internal/modules/cjs/helpers:102:18)
    at Object.<anonymous> (frontend/tests/codecept_e2e/node_modules/.pnpm/github.com+Automattic+node-canvas@d987d4666a32c4b8_190300f15912bbbd76c9d4ab8a89e735/node_modules/canvas/lib/bindings.js:3:18)
    at Module._compile (node:internal/modules/cjs/loader:1099:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1153:10)
    at Module.load (node:internal/modules/cjs/loader:975:32)
    at Function.Module._load (node:internal/modules/cjs/loader:822:12)
```

niyan-ly:
@nateGarcia Hi, did you still get this error on this PR? If so, could you give me a minimum replicable test case.

MeemeeLab:
Hello, first of all, thanks for this pull request! it will help my project.
but I found the problem. This pr made it possible to load canvas in the worker, but it cannot be loaded multiple times.
here is the test case:
```js
// index.js
import {Worker} from 'worker_threads';

function createWorker() {
    new Worker(`./worker.js`);
}

createWorker();
createWorker();
```
```js
// worker.js
import {Canvas} from "canvas";

console.log('thread', Canvas.toString());
```

By loading worker multiple times, it crash by showing `Module did not self-register`.

It happens too when loading canvas from main thread, and worker thread:
```js
// index.js
import {Worker} from 'worker_threads';
import {Canvas} from "canvas";

console.log('main', Canvas.toString());

new Worker(`./worker.js`);
```
```js
// worker.js
import {Canvas} from "canvas";

console.log('thread', Canvas.toString());
```

Windows x64 21H2 22000.795

chearon:
Context-aware now works for `npm install canvas@next`

PR #1982, opened 2022-02-03
Replace binary for rebuild cases

Apollon77:
When doing an "npm rebuild" (e.g. on nodejs major upgrades) then nothing happens right now because node-pre-gyp simply assumes that binary is already there, but do not "rebuild" it ... This fix makes sure that existing binary gets replaced on install or rebuilds

Thanks for contributing!


Apollon77:
@zbjornson DO you see any way to merge and release this?

TA2k:
@zbjornson This issue is really urgent and is affecting all people when they update their node version

zbjornson:
`npm rebuild` is specific to rebuilding dependencies *from source*. Instead you should use `npm install`, which re-runs the install script, which detects the new `NODE_MODULE_VERSION` and installs the correct binary. I just verified that this is working correctly by installing canvas with Node.js v16, downgrading to v14 and rerunning `npm install`. Forcing a new binary download each time would unnecessarily slow it down. Can you try that please?

Apollon77:
The problem is that the execution of "npm install" the the module directory has other effects because the module dependencies will then be duplicated in the library package directory and this then duplicates npm packages in the npm tree.

In my eyes it is good that also the "npm rebuild" command is supported by your library and has effects, because this is the usual way: After a nodejs upgrade the user executes "npm rebuild" in the "root package directory" and all dependent libraries in the tree rebuild themself if required.
Currently if the user do that e.g. serialport rebuilds itself correctly but canvas does not, so the application might not run. An "npm install" innthe root package will also not rebuild canvas because it will detect canvas as unchanged version and so not "npm install" in the canvas module ... This means that the user need to manualy find the package dir out our a cryptic crash message of the application and manually fix it himself ... This is not very convenient ...

Or do I miss anything?

zbjornson:
> the module dependencies will then be duplicated in the library package directory and this then duplicates npm packages in the npm tree.

I'm talking about running `npm install` from your package root directory, not in your_package/node_modules/canvas. There should be no duplication as a result.

TA2k:
@zbjornson Is this your final decision to not support rebuild from sub node_modules in favor for a faster installation time?

zbjornson:
Yes, but I'm talking about running `npm install` from your package root directory, not in your_package/node_modules/canvas (https://github.com/Automattic/node-canvas/pull/1982#issuecomment-1056052647).

Apollon77:
Yes but "npm install" from the root directory will not fix a "canvas sub depenedency" because it detects that the version is already correct and soo will not also execute a "Npm install" for canvas ... this is exactly the issue ... A "npm rebuild" on root level is indeed then having effects on the camvas sub dep, but because of the originalscript then having no rebuild effect because "binaries are there"

Apollon77:
You areright when it comes to updating the library ... but rebuild is usually "same version" case ... Or do I get you wrong?

chearon:
> And as said: most packages do exactly support that ... I only found two packages in our ecosystem that don't ... canvas and node-bluetooth-hci-socket :-)

Which other packages support this? It would be helpful to see.

My instinct was against this too, but I read up on [`npm rebuild`](https://docs.npmjs.com/cli/v7/commands/npm-rebuild) (which I have never used) and it does sound like given what it's supposed to do and given the [life cycle scripts](https://docs.npmjs.com/cli/v7/using-npm/scripts#life-cycle-operation-order) available, it makes more sense.

Apollon77:
> Which other packages support this? It would be helpful to see.

Honestly ... would need to see because it only seems a relevant issue im node-gyp-pre is used and then depending on the parameters :-)
Lets phrase it that way: We work on the smart home system ioBrober with over 460 plugins (aka dynamically added own npm projects with their own sub-deps) and we only found canvas and node-bluetooth-hci-socket to be problematic when the users do nodejs updates ... in node-bluetooth-hci-socket we proposed the same change and was accepted there.

Commonly used in the pluings is unix-dgram, authenticate-pam (because used in core system, but not using node-gyp-pre) and also serialport (older versions work fine, new is migrated to NAPI :-) ) ... so hard to "find" them  :-)

TA2k:
@chearon @zbjornson 
Have you come to a decision in your team?

Apollon77:
@chearon @zbjornson Hey, do you have any update on that?

chearon:
I still [can't find any repos](https://github.com/search?l=&p=4&q=node-pre-gyp+install+filename%3Apackage.json&ref=advsearch&type=Code) that do what you're suggesting and it's [not what node-pre-gyp](https://github.com/mapbox/node-pre-gyp#configuring) recommends. It would be helpful if you could show us an example of a popular repo that is doing this. The big drawback to merging this is that there could be a lot of unnecessary downloads.

Apollon77:
Ok, then lets's "simply" thy this  https://github.com/mapbox/node-pre-gyp/issues/646 ... maybe we get insights from the makers of node-gyp.

zbjornson:
Sorry for sort of blocking this and then taking forever to follow up. I just played more with the behavior of npm and yarn and see how the behavior is unintuitive and unique to modules with prebuilds. I'm good with landing this.

zbjornson (approved):

mateomarconi:
Hi,

I switched to Node.js version 18, but the 'canvas.node' module is compiled for version 16. I attempted to rebuild it without success; the error persists.

Error: The module '/app/node_modules/canvas/build/Release/canvas.node' was compiled against a different Node.js version using NODE_MODULE_VERSION 108. This version of Node.js requires NODE_MODULE_VERSION 93. Please try re-compiling or re-installing the module (for instance, using npm rebuild or npm install).

I am using a dependency that relies on the 'canvas' module.

Regarding this thread, how can I solve the issue?

Apollon77:
I would say simply delete /app/node_modules/canvas/build/* and do "npm rebuild" un the directory /app/node_modules/canvas/

PR #1985, opened 2022-02-09
Remove the last glib calls, replace with libicu

chearon:
Removes some GLib call, replaces others with libicu. We might eventually not need Pango, at which point we could remove GLib. If we remove Pango, libicu would also be needed for grapheme iteration for font fallbacks. All operating systems come with libicu, so I thought this would be easy, but it was a pain. macOS doesn't expose headers, so we have to use the homebrew version. Windows uses different naming, and the blocker: it compiles but crashes on startup.

I'm stumped on Windows but I'll try more later, for now I wanted to get this opened.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -64,7 +64,7 @@ jobs:
       - name: Install Dependencies
         run: |
           brew update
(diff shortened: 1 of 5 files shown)


chearon:
I ended up doing the important parts of this in #1987, we can figure the libicu nightmare when the time comes.

PR #1986, opened 2022-02-10
CI fix: make types compatible with typescript 4.6

chearon:
Typescript 4.6 seems to have changed `globalCompositeOperation` from `string` to an enum of strings, which means instead of narrowing the type we are now trying to widen it, which is impossible.


LinusU (approved):
Neat! 👍

PR #2013, opened 2022-03-29
add Path2D

tachibana-shin:
Thanks for contributing!

---> no

### DIFF
--- a/index.js
+++ b/index.js
@@ -10,6 +10,7 @@ const PNGStream = require('./lib/pngstream')
 const PDFStream = require('./lib/pdfstream')
 const JPEGStream = require('./lib/jpegstream')
 const { DOMPoint, DOMMatrix } = require('./lib/DOMMatrix')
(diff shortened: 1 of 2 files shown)


zbjornson (changes requested):
Nice start, thanks for working on this! It looks like there's some more work left to make this fully implement the Path2D spec. I pointed a few missing parts out below. Do you want to/have time to work on it?

We have the node-gfx organization where I think the actual `Path2D` class could live btw. (I'd like to move the `DOMMatrix` class there also.) I'm happy to work on the C++ changes so that we don't have to shim the context prototype methods like this also.

Surprisingly the [web-platform-tests](https://github.com/web-platform-tests/wpt) don't seem to really cover Path2D aside from `isPointInPath`. I was hoping those could save us time writing tests.

CreativeTechGuy:
For the record @zbjornson, the majority of code in this PR looks like it's directly from: https://github.com/nilzona/path2d-polyfill

PR #1987, opened 2022-02-10
Near-perfect font selection on Linux

chearon:
Finally done. Removes some GLib because it ended up making that code easier. Also updates the prebuild image so the next prebuilds will have improved matching too.

Fixes #1572

PR #2062, opened 2022-06-24
improve multi-family output in font desc resolver

### LINKED ISSUE #2041: 2.9.1 fails to load the correct weights in the same font family
## Issue or Feature
When loading multiple font weights in the same font family, node-canvas 2.9.1 fails to render the correct font weight. This issue did not occur in 2.9.0.

I'm loading four styles of the free "Open Sans" font family: Light (300), Regular (400), Bold (700), and Extra Bold (800). When trying to render text in Regular and Extra Bold as follows, 2.9.1 seems to instead render in Light and Regular, respectively:

```
const defaultFontFamily = '"Open Sans"';
ctx.fillStyle = "#fff";

ctx.font = `400 40px ${defaultFontFamily}`;
ctx.fillText("REGULAR", 40, canvas.height / 2 - 50);

ctx.font = `800 40px ${defaultFontFamily}`;
ctx.fillText("EXTRABOLD", 40, canvas.height / 2);
```

## Steps to Reproduce

I've created a barebones reproducible project, adapted from my actual project.

Running `node index.js` will generate an image named "canvas.png" in the same directory.

Project attached here: [repro.zip](https://github.com/Automattic/node-canvas/files/8734043/repro.zip)

### The following is how 2.9.1 incorrectly renders the example: 

canvas

### And how 2.9.0 correctly renders the example:

expected-2 9 0

## Your Environment
* Version of node-canvas (output of `npm list canvas` or `yarn list canvas`): 2.9.1
* Environment (e.g. node 4.2.0 on Mac OS X 10.8): node 14.18.2 on Debian 11 (via WSL2)

chearon:
This bug can be understood without considering the changes from #1987. When the user adds multiple TTFs and assigns them the same family name, we send the system families from the TTFs in a comma separated list, but we weren't correctly ordering that system-family list by the best matches on the user-description side. #1987 exposed it because it almost guarantees each file will have a unique system family.

<hr>

If the user registers fonts like this:

```is
registerFont("OpenSans-Bold.ttf", {family: "Open Sans", weight: 300});
registerFont("Arimo-Regular.ttf", {family: "Open Sans", weight: 600});
```

The `font_face_list` will look like this (approximated system descriptions):

| OpenSans-Bold.ttf | sys desc | user desc |
| -- | -- | -- |
| family | Open Sans | Open Sans |
| weight | 600 | 300 |

| Arimo-Regular.ttf | sys desc | user desc |
| -- | -- | -- |
| family | Arimo | Open Sans |
| weight | 400 | 600 |

When the user does this:

```is
ctx.font = "600 Open Sans";
```

We should send this to the OS:

| Arimo-Regular.ttf | sys desc |
| -- | -- |
| family | Arimo, Open Sans |
| weight | 400 |

But we were sending this:

| Arimo-Regular.ttf | user desc |
| -- | -- |
| family | Open Sans, Arimo |
| weight | 400 |

fixes #2041

### DIFF
--- a/src/Canvas.cc
+++ b/src/Canvas.cc
@@ -879,18 +879,21 @@ Canvas::ResolveFontDescription(const PangoFontDescription *desc) {
       if (streq_casein(family, pangofamily)) {
         const char* sys_desc_family_name = pango_font_description_get_family(ff.sys_desc);
         bool unseen = seen_families.find(sys_desc_family_name) == seen_families.end();
+        bool better = best.user_desc == nullptr || pango_font_description_better_match(desc, best.user_desc, ff.user_desc);
 
         // Avoid sending duplicate SFNT font names due to a bug in Pango for macOS:
         // https://bugzilla.gnome.org/show_bug.cgi?id=762873
         if (unseen) {
           seen_families.insert(sys_desc_family_name);
-          if (renamed_families.size()) renamed_families += ',';
-          renamed_families += sys_desc_family_name;
+
+          if (better) {
+            renamed_families = string(sys_desc_family_name) + (renamed_families.size() ? "," : "") + renamed_families;
+          } else {
+            renamed_families = renamed_families + (renamed_families.size() ? "," : "") + sys_desc_family_name;
+          }
         }
 
-        if (first && (best.user_desc == nullptr || pango_font_description_better_match(desc, best.user_desc, ff.user_desc))) {
-          best = ff;
-        }
+        if (first && better) best = ff;
       }
     }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 ### Changed
 ### Added
 ### Fixed
+* Wrong fonts used when calling `registerFont` multiple times with the same family name (#2041)
 
 2.9.1
 ==================


src/Canvas.cc
```diff
@@ -879,18 +879,21 @@ Canvas::ResolveFontDescription(const PangoFontDescription *desc) {
...
-        if (first && (best.user_desc == nullptr || pango_font_description_better_match(desc, best.user_desc, ff.user_desc))) {
-          best = ff;
+          if (better) {
+            renamed_families = string(sys_desc_family_name) + (renamed_families.size() ? "," : "") + renamed_families;
+          } else {
+            renamed_families = renamed_families + (renamed_families.size() ? "," : "") + sys_desc_family_name;
+          }
         }
+
+        if (first && better) best = ff;
```
zbjornson, line 896: Not tested, but could you get rid of `first` and replace this line with `if (best.user_desc == nullptr)`?

chearon: That would be a behavior change since we weren't setting the `best` if the first family wasn't found in the registered fonts. If you specify `"Arimo", "Open Sans"` but you never registered `"Arimo"`, then `"Arimo", sys_desc("Open Sans")` goes to the OS with the rest of the user description untouched.

`renamed_families` is also not sorted correctly, it only guarantees that the top match is first rather than ranking them all, but it's better than before and I don't think many people even use the font cascade that way. This probably needs an overhaul. It's all so complicated when we don't have entirely internal font selection 😞

zbjornson (approved):

Issue #528, opened 2015-02-22
node-gyp rebuild fails during npm install canvas


hayesmaker:
I get this when trying to install via $npm install canvas

npm ERR! canvas@1.2.1 install: `node-gyp rebuild`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the canvas@1.2.1 install script.
npm ERR! This is most likely a problem with the canvas package,
npm ERR! not with npm itself.
npm ERR! Tell the author that this fails on your system:
npm ERR!     node-gyp rebuild

kanzelm3:
+1

I'm having this issue too, seems that latest build is missing a required file for FontFace (fontconfig.h), stack trace below:

> canvas@1.2.1 install /Users/jkanzelmeyer/node_modules/canvas
> node-gyp rebuild

  SOLINK_MODULE(target) Release/canvas-postbuild.node
  SOLINK_MODULE(target) Release/canvas-postbuild.node: Finished
  CXX(target) Release/obj.target/canvas/src/Canvas.o
  CXX(target) Release/obj.target/canvas/src/CanvasGradient.o
  CXX(target) Release/obj.target/canvas/src/CanvasPattern.o
In file included from ../src/CanvasPattern.cc:10:
../src/CanvasPattern.h:23:9: warning: private field '_width' is not used [-Wunused-private-field]
    int _width, _height;
        ^
../src/CanvasPattern.h:23:17: warning: private field '_height' is not used [-Wunused-private-field]
    int _width, _height;
                ^
2 warnings generated.
  CXX(target) Release/obj.target/canvas/src/CanvasRenderingContext2d.o
../src/CanvasRenderingContext2d.cc:374:22: warning: unused variable 'surface' [-Wunused-variable]
    cairo_surface_t *surface = cairo_get_group_target(_context);
                     ^
1 warning generated.
  CXX(target) Release/obj.target/canvas/src/color.o
  CXX(target) Release/obj.target/canvas/src/Image.o
  CXX(target) Release/obj.target/canvas/src/ImageData.o
  CXX(target) Release/obj.target/canvas/src/init.o
  CXX(target) Release/obj.target/canvas/src/PixelArray.o
  CXX(target) Release/obj.target/canvas/src/FontFace.o
../src/FontFace.cc:9:10: fatal error: 'fontconfig/fontconfig.h' file not found
#include <fontconfig/fontconfig.h>
         ^
1 error generated.
[...]

PR #2073, opened 2022-07-07
Add M1 compiling note

infodusha:
I have not you updated CHANGELOG.md, because of that is just doc change.

The change was got from the issue #528

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -23,7 +23,7 @@ For detailed installation information, see the [wiki](https://github.com/Automat
 
 OS | Command
 ----- | -----
-OS X | Using [Homebrew](https://brew.sh/):<br/>`brew install pkg-config cairo pango libpng jpeg giflib librsvg`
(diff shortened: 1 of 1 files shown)


LinusU:
Hmm, I'm using an M1 and I don't need to run `arch -arm64` in front of brew 🤔 

Are you sure that your Homebrew is setup correctly? Is there any more background on why one would want to have that in front?

LinusU:
I'm not sure, what happens if you run `brew` without `arch -arm64`?

infodusha:
It installs all the deps fine but the error that present in the issue i mentioned still presents

infodusha:
That could be, i don't actually know how homebrew works with Rosetta

infodusha:
Can we pause that for a while, i am going to dig into details and versify my thoughts

MetaverseMan:
with respect to all you bros ,I'm using an M1 ,Monterey 12.4, and I am already prepared mentally to use the second way "arch -arm64 brew ..."to  figure out ,but when i just  try it with no "arch -arm64" in front of brew, it works, so i think we can just try another when the first method we choose to use didn't work.

PR #2090, opened 2022-08-11
Fix: missing type definition

arshmalikrs:
Add missing type definition for ```deregisterAllFonts()```

Thanks for contributing!


zbjornson:
Thanks for this! I just landed the same change in #2096, which had a test added.

PR #2104, opened 2022-08-29
fix: add user agent to remote images request

### LINKED ISSUE #2099: Setting request headers for remote image requests
## Issue or Feature
Some websites block image requests made from canvas, for example Wikipedia. From Googling around it may be due to missing or invalid headers, e.g. User-Agent. Is there a way to set them currently? If not, please consider setting sane default headers or allowing the user to specify them for individual requests.

## Steps to Reproduce
```javascript
const Image = require('canvas');

async function getImage(src) {
    return new Promise(resolve => {
        const img = new Image();
        img.onload = () => resolve(img);
        img.onerror = err => { console.log(err); resolve(null); };
        img.src = src;
    });
}

// will be null, and 403 error printed out to the console
var img = await getImage("https://upload.wikimedia.org/wikipedia/commons/4/48/Daala_Rainbow_Vomit_Tiger.png");
```

## Your Environment
* Version of node-canvas (output of `npm list canvas` or `yarn list canvas`): canvas@2.9.3
* Environment (e.g. node 4.2.0 on Mac OS X 10.8): node v14.17.3 on Windows 10

apollisa:
Fixes #2099

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 * Crashes and hangs when using non-finite values in `context.arc()`. ([#2055](https://github.com/Automattic/node-canvas/issues/2055))
 * Incorrect `context.arc()` geometry logic for full ellipses. ([#1808](https://github.com/Automattic/node-canvas/issues/1808), ([#1736](https://github.com/Automattic/node-canvas/issues/1736)))
 * Added missing `deregisterAllFonts` to the Typescript declaration file ([#2096](https://github.com/Automattic/node-canvas/pull/2096))
+* Add `User-Agent` header when requesting remote images ([#2099](https://github.com/Automattic/node-canvas/issues/2099))
 
 2.9.3
 ==================
--- a/lib/image.js
+++ b/lib/image.js
@@ -49,7 +49,10 @@ Object.defineProperty(Image.prototype, 'src', {
 
         if (!get) get = require('simple-get')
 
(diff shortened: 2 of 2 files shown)


LinusU:
I'm not sure if it's great to send the version of canvas being used, or even that canvas is at all.

What do we think about mimicking what Chrome is doing here:
https://www.chromium.org/updates/ua-reduction/#TOC-Proposed-Rollout-Plan

I'm not sure if we even want to expose which platform we are running though, maybe just pick something and stick with it?

e.g.

```text
Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/102.0.0.0 Safari/537.36
```

(102 seems to be an LTS version, we could bump this every ~6 months, [ref](https://support.google.com/chrome/a/answer/11333726))

This way we won't inadvertently leak something that the user isn't expecting to be exposed, and we also should be somewhat indistinguishable from one of the most common browsers. So we should get stuck in even less filters...

apollisa:
I see no problem with this, I just based myself upon what [curl is doing](https://everything.curl.dev/http/requests/user-agent), but your solution looks good too. I’ll change this right away. :smiley:

LinusU (approved):
LGTM 👍

apollisa:
Just for info @LinusU, how soon can we expect the fix to be 1) merged 2) released?

zbjornson:
I can make a release later today.

zbjornson:
@apollisa v2.10.0 is published, sorry for the delay.

PR #2110, opened 2022-09-07
Parse rgba(r,g,b,0) correctly

H3r3zy:
In https://github.com/Automattic/node-canvas/commit/52551952c3d78ff12110880a2101ab980dd7bf6c rgba with alpha to 0 is not parsed correctly.

It must be 0.00 instead of 1

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,5 @@ package-lock.json
 *.swp
 *.un~
 npm-debug.log
+
+.idea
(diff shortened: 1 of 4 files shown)


zbjornson (approved):
Oops, thanks!

Prebuilds are running now: https://github.com/Automattic/node-canvas/actions/runs/3009310715

PR #2115, opened 2022-09-10
remove save() limit, improve save/restore perf, fix some props not saved/restored

### LINKED ISSUE #1936: Context save/restore does not handle textBaseline
## Issue

The context.textBaseline property doesn't seem to be part of the drawing state and is unaffected by save/restore calls.
https://html.spec.whatwg.org/multipage/canvas.html#the-canvas-state
https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/save#drawing_state

## Steps to Reproduce

```js
const {createCanvas} = require('canvas');

const canvas = createCanvas(100, 100);

context.textBaseline = 'top';
console.log('top', context.textBaseline);
context.save();
context.textBaseline = 'middle';
console.log('middle', context.textBaseline);
context.restore();
console.log('top', context.textBaseline);
```

*Expected output*
top top
middle middle
top top

*Actual output*
top top
middle middle
top middle

## Your Environment
* 2.8.0
* Node 12.16.2, Windows 10 (21H2)

Also tested with canvas 2.6.1 and node 8.17.0, issue there as well.

zbjornson:
1. One WPT test fails if there are not at least 512 save/restore slots. This removes that limit entirely.

2. Gets rid of clunky C code and uses `std::stack` with a proper C++ class. End result is >1.6x faster with MSVC.

3. Reorders fields and types some enums so the state struct shrinks from 192 bytes to 168 bytes (-24 bytes; i.e. 24 bytes saved per state).

4. Fixes several properties that were not saved/restored: `textBaseline`, `textAlign`. `quality` is not saved/restored, but it's not wired up to anything and needs to be removed.

Fixes #1936

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,11 @@ project adheres to [Semantic Versioning](http://semver.org/).
 (Unreleased)
 ==================
 ### Changed
+* Improve performance and memory usage of `save()`/`restore()`.
+* `save()`/`restore()` no longer have a maximum depth (previously 64 states).
 ### Added
 ### Fixed
+* `textBaseline` and `textAlign` were not saved/restored by `save()`/`restore()`. ([#1936](https://github.com/Automattic/node-canvas/issues/2029))
 
 2.10.1
 ==================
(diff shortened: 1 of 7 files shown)


LinusU (approved):
Awesome! 🎉

PR #2127, opened 2022-10-05
feat: implement `OffscreenCanvas.convertToBlob`

### LINKED ISSUE #1845: Introduce blob's

jimmywarting:

This new feature would require ppl to use NodeJS v18 but it's not required. 
ppl could also polyfill Blob themself it they wanted to use it.

closes #1845

### DIFF
--- a/lib/canvas.js
+++ b/lib/canvas.js
@@ -7,13 +7,13 @@
  */
 
 const bindings = require('./bindings')
-const Canvas = module.exports = bindings.Canvas
 const Context2d = require('./context2d')
 const PNGStream = require('./pngstream')
 const PDFStream = require('./pdfstream')
 const JPEGStream = require('./jpegstream')
 const FORMATS = ['image/png', 'image/jpeg']
 const util = require('util')
+const Canvas = bindings.Canvas
 
 // TODO || is for Node.js pre-v6.6.0
 Canvas.prototype[util.inspect.custom || 'inspect'] = function () {
@@ -42,6 +42,31 @@ Canvas.prototype.createPDFStream = function (options) {
 Canvas.prototype.jpegStream =
 Canvas.prototype.createJPEGStream = function (options) {
   return new JPEGStream(this, options)
+}
+
+/**
+ * The OffscreenCanvas.convertToBlob() method creates a Blob object representing
+ * the image contained in the canvas.
+ *
+ * @see https://developer.mozilla.org/en-US/docs/Web/API/OffscreenCanvas/convertToBlob
+ * @since NodeJS v18.0.0
+ */
+Canvas.prototype.convertToBlob = async function (options = {}) {
+  // If the user agent does not support the requested type, then it must create the file using the PNG format.
+  // ref: https://html.spec.whatwg.org/multipage/canvas.html#a-serialisation-of-the-bitmap-as-a-file
+  const type = options.type && FORMATS.includes(options.type)
+    ? options.type
+    : 'image/png'
+
+  const quality = options.quality != null
+    ? { quality: options.quality }
+    : undefined
+
+  return new Promise((resolve, reject) => {
+    this.toBuffer((err, buf) => {
+      err ? reject(err) : resolve(new Blob([buf], { type }))
+    }, type, quality)
+  })
 }
 
 Canvas.prototype.toDataURL = function (a1, a2, a3) {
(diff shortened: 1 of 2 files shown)


lib/canvas.js
```diff
@@ -44,6 +44,25 @@ Canvas.prototype.createJPEGStream = function (options) {
...
 
+/**
+ * The OffscreenCanvas.convertToBlob() method creates a Blob object representing
+ * the image contained in the canvas.
+ *
+ * @see https://developer.mozilla.org/en-US/docs/Web/API/OffscreenCanvas/convertToBlob
+ * @since NodeJS v18.0.0
+ */
+Canvas.prototype.convertToBlob = async function (options = {}) {
+  const type = options.type && FORMATS.includes(options.type)
```
LinusU, line 55: This seemed quite stupid, so I think it's best to include a link that this is done according to spec 😅 

```suggestion
  // If the user agent does not support the requested type, then it must create the file using the PNG format.
  // ref: https://html.spec.whatwg.org/multipage/canvas.html#a-serialisation-of-the-bitmap-as-a-file
  const type = options.type && FORMATS.includes(options.type)
```

lib/canvas.js
```diff
@@ -111,3 +130,5 @@ Canvas.prototype.toDataURL = function (a1, a2, a3) {
     return `data:${type};base64,${this.toBuffer(type, opts).toString('base64')}`
   }
 }
+
+module.exports = Canvas
```
LinusU, line 134: Can you add a final newline here?

```suggestion
module.exports = Canvas
```

lib/canvas.js
```diff
@@ -44,6 +44,25 @@ Canvas.prototype.createJPEGStream = function (options) {
...
+ */
+Canvas.prototype.convertToBlob = async function (options = {}) {
+  const type = options.type && FORMATS.includes(options.type)
+    ? options.type
+    : 'image/png'
+
+  return new Promise((resolve, reject) => {
+    this.toBuffer((err, buf) => {
+      err ? reject(err) : resolve(new Blob([buf], { type }))
+    }, type, options)
```
LinusU, line 62: Instead of passing all `options` thru, what do you think about just looking at the supported `quality`. So that people don't start depending on passing other options thru here? Since as far as I can tell from the specification, only `type` and `quality` is allowed? 🤔 

```js
  const quality = options.quality != null
    ? { quality: options.quality }
    : undefined
```

```suggestion
    }, type, quality)
```

LinusU (commented):
Nice!! Left som small comments to address

Could you add a line in the changelog file as well?

lib/canvas.js
jimmywarting: sure 👍

jimmywarting:
bump

LinusU:
Sorry for the delay on this, could you add a line to the changelog please? 🙏

jimmywarting:
done

LinusU (approved):
Neat 👍

LinusU:
(Windows tests are failing, testing rebasing on lastest `master`)

jimmywarting:
The failed test is unrelated to this PR. maybe this can be merged anyway?

```
  1) Canvas
       registerFont:
     Error: ENOENT: no such file or directory, lstat 'D:\a\node-canvas\node-canvas\examples\pfennigFont\pfennigMultiByte🚀.ttf'
```

PR #2174, opened 2022-12-14
use tailored types instead of extending DOM

### LINKED ISSUE #1656: TypeScript typings pollute global types by including lib.dom.d.ts
## Issue or Feature
[types/index.d.ts](https://github.com/Automattic/node-canvas/blob/595d5590a0c4199c483d4244b59d28341086edc9/types/index.d.ts#L2) contains the following line:
```
/// <reference lib="dom" />
```
It references lib.dom.d.ts, which usually isn't included in Node.js projects. This makes TypeScript think `document`, `requestAnimationFrame` and other browsers-only APIs are available in the global namespace.

## Steps to Reproduce

```sh
mkdir canvas-bug
cd canvas-bug
npm init --yes
npm install --save-dev typescript @types/node
echo "document.write('Hello')" > index.ts
npx tsc --lib es2020 index.ts
```
As expected, the code will not compile because `document` doesn't exist in Node.js.
```sh
npm install canvas
echo "import { Canvas } from 'canvas'; document.write('Hello');" > index.ts
npx tsc --lib es2020 index.ts
```
The code will now compile despite being invalid.

## Your Environment
* Version of node-canvas (output of `npm list canvas` or `yarn list canvas`): 2.6.1
* Environment (e.g. node 4.2.0 on Mac OS X 10.8): node 14.8.0 on Windows 10

chearon:
The properties are ordered the same as their implementations, and I looked at each to make sure the signatures match. There are a lot of arguments in the DOM types that weren't supported by node-canvas, so this feels like the right way.

I tried to be careful, but this was a lot. I checked that every example file passes `// @ts-check`. Even if there are no mistakes, since these new types are smaller, someone could have been passing arguments that node-canvas wouldn't recognize and it would break their typescript build.

Fixes #1656

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -1,5 +1,4 @@
 // TypeScript Version: 3.0
-/// <reference lib="dom" />
 
 import { Readable } from 'stream'
 
@@ -80,7 +79,7 @@ export class Canvas {
 
 	constructor(width: number, height: number, type?: 'image'|'pdf'|'svg')
 
-	getContext(contextId: '2d', contextAttributes?: NodeCanvasRenderingContext2DSettings): NodeCanvasRenderingContext2D
+	getContext(contextId: '2d', contextAttributes?: NodeCanvasRenderingContext2DSettings): CanvasRenderingContext2D
 
 	/**
 	 * For image canvases, encodes the canvas as a PNG. For PDF canvases,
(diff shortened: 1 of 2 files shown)


LinusU (approved):
This seems like a good way forward 👍

types/index.d.ts
```diff
@@ -80,7 +79,7 @@ export class Canvas {
 
 	constructor(width: number, height: number, type?: 'image'|'pdf'|'svg')
 
-	getContext(contextId: '2d', contextAttributes?: NodeCanvasRenderingContext2DSettings): NodeCanvasRenderingContext2D
+	getContext(contextId: '2d', contextAttributes?: NodeCanvasRenderingContext2DSettings): CanvasRenderingContext2D
```
zbjornson, line 82: This probably means a semver-major release, right?

types/index.d.ts
```diff
@@ -128,19 +127,138 @@ export class Canvas {
...
+	ellipse(x: number, y: number, radiusX: number, radiusY: number, rotation: number, startAngle: number, endAngle: number, counterclockwise?: boolean): void;
+	setLineDash(segments: number[]): void;
+	getLineDash(): number[];
+	createPattern(image: Canvas|Image, repetition: 'repeat' | 'repeat-x' | 'repeat-y' | 'no-repeat' | '' | null): CanvasPattern
+	/**
+	 * **Do not use this overload. Use the other three overload.** This is a
+	 * catch-all definition required for compatibility with the DOM
+	 * `CanvasRenderingContext2D` interface.
+	 */
+	createPattern(...args: any[]): CanvasPattern
```
zbjornson, line 240: Might be able to delete this?

zbjornson (commented):
👍

types/index.d.ts
LinusU: Hmm, might be worth keeping the old name then? 🤔

chearon: We could also export aliases to the old names. I thought the "NodeCanvas" prefix was to avoid naming conflicts from `lib.dom.ts`.

Even then, it still could break someone's build if they were making function calls with arguments we ignore. Not sure if that's as big of a deal though?

zbjornson: I'm not at all opposed to a semver-major release. This would be an easy one for users to handle. Just want to be on the same page.

Other question is usability - users could alias the type themselves (`import type {CanvasRenderingContext2D as NodeCanvasRenderingContext2D}`), but is it more convenient for the provided name to be NodeCanvasRenderingContext2D?

chearon: I prefer the shorter name. We call it Canvas everywhere else, not NodeCanvas. Our main users are using node, so it shouldn't cause a naming conflict should it? If they do have DOM types, the imported type would take priority over the ambient types, which is what you'd expect. I'm not 100% certain it's the right call, but pretty sure it would be fine...

types/index.d.ts
chearon: Removed both of them

types/index.d.ts
chearon: Plus we already kind did this. You can `import {Image}` which would conflict with the browser-defined `Image`.

zbjornson: Fair point, same for the the gradient and pattern classes.

@LinusU  good with you? Shorter name and semver-major?

LinusU: Hmm, it seems like the old code exported without the `Node` prefix, so I'm not sure this is a breaking change?

https://github.com/Automattic/node-canvas/blob/672104c1a4bd202e56d8837ef83ebf7aee2dfce2/types/index.d.ts#L210

chearon: That's true. I think it would still let you `import {NodeCanvasRenderingContext2D} from 'canvas'` because it's a .d.ts. Could still export both. IMO the chances of this breaking someone's build are small as-is and we can only make it smaller but not get to 0.

zbjornson: Good point! Makes more sense as semver-minor then.

PR #2180, opened 2022-12-24
Fix missing canvas property in CanvasRenderingContext2D

Suyooo:
Yesterday's v2.11 update with the new type definitions is great to have - one thing I noticed missing though is the [`CanvasRenderingContext2D.canvas` property](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/canvas), which is set in `lib/canvas.js` instead of the `Initialize` function of `Context2d`. This is just a small one-line PR to declare it on the type.


zbjornson (approved):
Thank you!

PR #2184, opened 2023-01-02
fix incorrect text width with newer (1.43?) Pango

chearon:
Text renders wider or narrower, as if letter-spacing is set, than it should with newer versions of Pango.

| `master` (node-canvas - browser) | `ch/text-width` (node-canvas - browser) |
| -- | -- |
|  Screen Shot 2023-01-02 at 5 54 29 PM | Screen Shot 2023-01-02 at 5 53 10 PM |

My best understanding of this problem is that around version 1.43 Pango dropped support for font hinting because it switched to Harbuzz for glyph postions instead of Freetype. For some reason it still rounds the glyph positions by default.

There's no need for node-canvas to support font hinting. The maintainers of the Linux font stack (Behdad and Matthias) have stated that they wont, and font hinting is subjective, and browsers have moved to subpixel positioning too.

Reading (warning: lots of drama to wade through):
- https://gitlab.gnome.org/GNOME/pango/-/issues/404
- https://gitlab.gnome.org/GNOME/pango/-/issues/463
- https://github.com/harfbuzz/harfbuzz/issues/1892
- https://github.com/harfbuzz/harfbuzz/issues/2394

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 ### Fixed
 * Add missing property `canvas` to the `CanvasRenderingContext2D` type
+* Fixed glyph positions getting rounded, resulting text having a slight `letter-spacing` effect
 
 2.11.0
 ==================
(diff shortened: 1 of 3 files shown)


LinusU (approved):
👍

PR #2187, opened 2023-01-07
move ctx.font string to the state struct

### LINKED ISSUE #1946: CanvasRenderingContext2d.restore() does not restore font weight
## Issue
CanvasRenderingContext2d.restore() does not restore font weight.

## Steps to Reproduce

```js

ctx.font = 'normal 16px sans-serif';
ctx.save():
ctx.font = 'bold 16px sans-serif';
ctx.restore();
ctx.font === 'bold 16px sans-serif' // should be 'normal 16px sans-serif' now
```

## Your Environment
* canvas@2.8.0
* node@16.4.2 on Windows 10

chearon:
I don't totally understand why we were using a `Nan::Persistent<v8::Value>` instead of storing a standard string + turn it into a v8 value on the fly, so hopefully this is right. 

fixes #1946

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 * Add missing property `canvas` to the `CanvasRenderingContext2D` type
 * Fixed glyph positions getting rounded, resulting text having a slight `letter-spacing` effect
+* Fixed `ctx.font` not being restored correctly after `ctx.restore()` (#1946)
 
 2.11.0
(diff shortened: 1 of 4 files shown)


chearon:
Ah, @zbjornson you fixed it the same way for other properties in 0e6504a1f6ad28eba5f40835fc233275a4170d46

zbjornson (approved):
My only guess is that someone wanted to avoid the overhead of going between a V8 string and a std::string, but this branch improves performance of `get font` by about 10% and has no impact on `set font`.

PR #2214, opened 2023-03-20
add string tags for browser polyglot classes

### LINKED ISSUE #1646: `ImageData` not stringifying consistently on Node 14.7
## Issue or Feature

In [`typeson-registry`](https://github.com/dfahlander/typeson-registry/), we need to detect object type in order to properly serialize objects for JSON.

~For some reason, on Node 14 in Linux (Xenial), this is failing, but it is not failing in Node 10 or 12 in Linux, nor in Node 14 on the Mac.~ *This is failing on Node 14.7, but not in Node 10 or 12.*

In performing some logging, it turns out that calling `Object.prototype.toString.call()` on `ImageData` objects will give `[object Object]` in Node 14.7, but in other environments, as expected, it gives `[object ImageData]`.

Perhaps this is due to a proper fix in Node 14.7 since objects should only give `[object Object]` unless [`Symbol.toStringTag`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/toStringTag) is used, but in any case, if it is not doing so already, I'd hope your build process could add the equivalent of the following to `ImageData` to ensure it behaves as in the browser so that one can detect the type of such objects (this approach will, unlike `instanceof`, also ensure detection continues to work cross-frame when used in polyglot browser code).

```js
                get [Symbol.toStringTag] () {
                    return 'ImageData';
                }
```

## Steps to Reproduce
```js
const {ImageData} = require('canvas');
const imageData = new ImageData(1, 3);
console.log(Object.prototype.toString.call(imageData));
// Expected `[object ImageData]` but get `[object Object]` in Node 14 Linux
```

## Your Environment
* Version of node-canvas (output of `npm list canvas` or `yarn list canvas`): 2.6.1
* Environment (e.g. node 4.2.0 on Mac OS X 10.8): Node 14.7 on Linux xenial or MacOSX; see [the Travis job](https://travis-ci.com/github/brettz9/typeson-registry/jobs/370534567)

Thanks!

brettz9:
Thanks for contributing!

Fixes #1646

### DIFF
--- a/lib/bindings.js
+++ b/lib/bindings.js
@@ -5,11 +5,13 @@ const bindings = require('../build/Release/canvas.node')
 module.exports = bindings
 
 Object.defineProperty(bindings.Canvas.prototype, Symbol.toStringTag, {
-  value: 'HTMLCanvasElement'
+  value: 'HTMLCanvasElement',
+  configurable: true
 })
 
 Object.defineProperty(bindings.Image.prototype, Symbol.toStringTag, {
-  value: 'HTMLImageElement'
+  value: 'HTMLImageElement',
+  configurable: true
 })
 
 bindings.ImageData.prototype.toString = function () {
@@ -17,7 +19,8 @@ bindings.ImageData.prototype.toString = function () {
 }
 
 Object.defineProperty(bindings.ImageData.prototype, Symbol.toStringTag, {
-  value: 'ImageData'
+  value: 'ImageData',
+  configurable: true
 })
 
 bindings.CanvasGradient.prototype.toString = function () {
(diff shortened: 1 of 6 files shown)


brettz9:
Anything else I can do for this?

lib/bindings.js
```diff
@@ -4,10 +4,34 @@ const bindings = require('../build/Release/canvas.node')
 
 module.exports = bindings
 
+Object.defineProperty(bindings.Canvas.prototype, Symbol.toStringTag, {
+  value: 'HTMLCanvasElement'
+})
+
+Object.defineProperty(bindings.Image.prototype, Symbol.toStringTag, {
+  value: 'HTMLImageElement'
+})
```
zbjornson, line 13: Sorry for the delay. I can't find this in the spec, but both Firefox and Chrome have these `configurable: true`.

```
> Object.getOwnPropertyDescriptor(Image.prototype, Symbol.toStringTag)
Object { value: "HTMLImageElement", writable: false, enumerable: false, configurable: true }
```

```
> Object.getOwnPropertyDescriptor(Object.getPrototypeOf(document.createElement("canvas")), Symbol.toStringTag)
Object { value: "HTMLCanvasElement", writable: false, enumerable: false, configurable: true }
```

brettz9: Good catch, thanks. I confirmed and applied for the others as well.

zbjornson (approved):
Great, thank you and sorry for the delay.

PR #2218, opened 2023-03-27
Fix build error C2664 - cannot convert argument 6 from 'nullptr' to 'DWORD

acalcutt:
This is a precursor to https://github.com/Automattic/node-canvas/pull/2218, but it looks like it also fixes an issue in the current prebuild for windows.

In the "prebuilds" workflow there is currently an error

`Fixes "error C2664: 'HANDLE CreateFileW(LPCWSTR,DWORD,DWORD,LPSECURITY_ATTRIBUTES,DWORD,DWORD,HANDLE)': cannot convert argument 6 from 'nullptr' to 'DWORD' [D:\a\node-canvas\node-canvas\build\canvas.vcxproj]"`

image

This change fixes that error.

You can see the build succeeds in my test here https://github.com/acalcutt/node-canvas/actions/runs/4527936648/jobs/7974248529
(note, it still fails to upload since I didn't set that up for my fork)


chearon:
Thanks but this is a duplicate of #2216. I'd rather pass one of the documented arguments than zero.

PR #2223, opened 2023-04-03
Update simple-get dependency ^3.0.3 to ^4.0.1

YuMuuu:
I've updated the versionof  simple-get module due to security issue.

versions bellow 4.0.1 have a vulnerability that leaks cookie headers to third-party sites. 

url: https://huntr.dev/bounties/42c79c23-6646-46c4-871d-219c0d4b4e31/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 ### Changed
 ### Added
 ### Fixed
-- Upgrade single-get dependency ^3.0.3 to ^4.0.1
+- Upgrade sinmle-get dependency ^3.0.3 to ^4.0.1
 
 2.11.2
 ==================
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 ### Changed
 ### Added
 ### Fixed
+- Upgrade single-get dependency ^3.0.3 to ^4.0.1
 
 2.11.2
 ==================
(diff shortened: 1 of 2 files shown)


piranna:
I think it should be replaced by built-in `fetch` API.

chearon:
> I think it should be replaced by built-in fetch API.

That's only in node 17+

This PR bumps up the node version too, but we currently support node 8+, so I don't think we can merge yet.

LinusU:
> This PR bumps up the node version too, but we currently support node 8+, so I don't think we can merge yet.

Yes, this would be a breaking change.

-----

> https://huntr.dev/bounties/42c79c23-6646-46c4-871d-219c0d4b4e31/

The way that this package is using `simple-get` is not vulnerable to this issue since there isn't a way to set the `cookie` or `authorization` header in the first place.

-----

Long term it would be nice to use the built-in `fetch`.

Node.js 16.x goes end-of-life on the 11th of september:

https://nodejs.org/en/blog/announcements/nodejs16-eol

If version 3.0.0 of Canvas is released on or after that date we should be able to support only 18+ and thus use the built-in fetch...

YuMuuu:
umm...
I understanded impossible "versin up simple-get" and "unuse simple-get".
But I don't think you should rely on vulnerable libraries even if they have no direct repercussions.

Do you have any other good ideas?

chearon:
> But I don't think you should rely on vulnerable libraries even if they have no direct repercussions.

It's not really a vulnerability if there's no way to exploit it

YuMuuu:
I got it.  This PR will be closed. Thanks for your support. 
I'll issue a PR to fix it to use "fetch-api" soon.

PR #2231, opened 2023-04-18
tools: update GitHub Actions workflows

VoltrexKeyva:
Thanks for contributing!

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -10,7 +10,7 @@ on:
 jobs:
   Linux:
     name: Test on Linux
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     strategy:
       matrix:
         node: [10, 12, 14, 16, 18, 20]
@@ -18,6 +18,7 @@ jobs:
       - uses: actions/setup-node@v3
         with:
           node-version: ${{ matrix.node }}
+          check-latest: true
       - uses: actions/checkout@v3
       - name: Install Dependencies
         run: |
@@ -30,7 +31,7 @@ jobs:
 
   Windows:
     name: Test on Windows
-    runs-on: windows-2019
+    runs-on: windows-2022
     strategy:
       matrix:
         node: [10, 12, 14, 16, 18, 20]
@@ -38,12 +39,14 @@ jobs:
       - uses: actions/setup-node@v3
         with:
           node-version: ${{ matrix.node }}
+          check-latest: true
       - uses: actions/checkout@v3
       - name: Install Dependencies
         run: |
+          [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
           Invoke-WebRequest "https://ftp-osl.osuosl.org/pub/gnome/binaries/win64/gtk+/2.22/gtk+-bundle_2.22.1-20101229_win64.zip" -OutFile "gtk.zip"
           Expand-Archive gtk.zip -DestinationPath "C:\GTK"
-          Invoke-WebRequest "https://downloads.sourceforge.net/project/libjpeg-turbo/2.0.4/libjpeg-turbo-2.0.4-vc64.exe" -OutFile "libjpeg.exe" -UserAgent NativeHost
+          Invoke-WebRequest "https://downloads.sourceforge.net/project/libjpeg-turbo/2.1.5.1/libjpeg-turbo-2.1.5.1-vc64.exe" -OutFile "libjpeg.exe" -UserAgent NativeHost
           .\libjpeg.exe /S
           npm install -g node-gyp@8
           npm prefix -g | % {npm config set node_gyp "$_\node_modules\node-gyp\bin\node-gyp.js"}
@@ -54,7 +57,7 @@ jobs:
 
   macOS:
     name: Test on macOS
-    runs-on: macos-latest
+    runs-on: macos-12
     strategy:
       matrix:
         node: [10, 12, 14, 16, 18, 20]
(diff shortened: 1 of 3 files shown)


VoltrexKeyva:
This also fixes a typo in the changelog.

.github/workflows/ci.yaml
```diff
@@ -7,17 +7,18 @@ on:
...
 
-jobs: 
+jobs:
   Linux:
     name: Test on Linux
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node: [10, 12, 14, 16, 18]
+        node: [14, 16, 18]
```
LinusU, line 16: We are still supporting these versions so I don't think we should remove them until the next major version when we can bump our minimum Node.js version...

.github/workflows/ci.yaml
```diff
@@ -30,22 +31,24 @@ jobs:
 
   Windows:
     name: Test on Windows
-    runs-on: windows-2019
+    runs-on: windows-latest
```
LinusU, line 34: A nice thing with pinning is that the CI won't randomly start breaking when the next runner is released, but would rather require a PR that changes this value to the next one and then we would see if that breaks there.

Having things run on `latest` can be frustrating if e.g. someone want's to submit a small PR that fixes a bug, but then that small PR is blocked because CI is failing because of change brought by the newer runner...

.github/workflows/ci.yaml
VoltrexKeyva: I see, should I open a separate PR to drop support for them? Is there anything that should land before we can bump the minimum Node.js version?

.github/workflows/ci.yaml
VoltrexKeyva: You're right though all the other platforms specified in these two GitHub Actions workflows files use the `latest` versions except for this particular step, should I update all the platform versions to the latest version without the `latest` tag? E.g. `ubuntu-22.04`.

zbjornson: I indeed had to pin to windows-2019 recently because -latest broke something that I didn't have time to look into.

.github/workflows/prebuild.yaml
zbjornson, line 1: This file isn't used in this branch.

.github/workflows/ci.yaml
```diff
@@ -30,22 +31,24 @@ jobs:
...
       - name: Install Dependencies
         run: |
+          [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
           Invoke-WebRequest "https://ftp-osl.osuosl.org/pub/gnome/binaries/win64/gtk+/2.22/gtk+-bundle_2.22.1-20101229_win64.zip" -OutFile "gtk.zip"
           Expand-Archive gtk.zip -DestinationPath "C:\GTK"
-          Invoke-WebRequest "https://downloads.sourceforge.net/project/libjpeg-turbo/2.0.4/libjpeg-turbo-2.0.4-vc64.exe" -OutFile "libjpeg.exe" -UserAgent NativeHost
+          Invoke-WebRequest "https://downloads.sourceforge.net/project/libjpeg-turbo/2.1.5.1/libjpeg-turbo-2.1.5.1-vc64.exe" -OutFile "libjpeg.exe" -UserAgent NativeHost
           .\libjpeg.exe /S
-          npm install -g node-gyp@8
+          npm install -g node-gyp
```
zbjornson, line 51: This breaks with earlier versions of node.

zbjornson (commented):
Do you have a specific motivation for wanting to drop older versions of Node.js? Right now it's not adding to our maintenance burden, and folks have complained when we stopped providing prebuilds for EOL versions, so we've continued to test them in CI.

VoltrexKeyva:
> Do you have a specific motivation for wanting to drop older versions of Node.js? Right now it's not adding to our maintenance burden, and folks have complained when we stopped providing prebuilds for EOL versions, so we've continued to test them in CI.

My main motivation for that is to be able to update dependencies without having to worry about it breaking support for older Node.js versions, as well as being able to use newer features not available in older versions that are more convenient to use and offer better performance (such as the nullish coalescing operator (`??`) and others), since most users use the newer versions rather than the older ones nowadays, but I would understand if you folks still want to support older versions.

LinusU:
I personally think that we shouldn't rush to release a new major version since it creates extra churn for the consumers of this package. We only have three dependencies right now, and only one of them has a newer version available, `simple-get`, and there is nothing in the new version that would benefit us at this time...

VoltrexKeyva:
Fair enough, I'll revert the changes that removes the old Node.js versions from the testing matrix.

.github/workflows/prebuild.yaml
VoltrexKeyva: Then is there a reason it exists in this branch? Should we remove it?

.github/workflows/ci.yaml
```diff
@@ -75,11 +79,12 @@ jobs:
...
   Lint:
     name: Lint
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     steps:
       - uses: actions/setup-node@v3
         with:
-          node-version: 14
+          node-version: 18
+          chexk-latest: true
```
LinusU, line 87: typo?

```suggestion
          check-latest: true
```

LinusU (approved):
LGTM 👍 

The only thing I'm not 100% sure about is the `check-latest: true`, not sure that it's worth busting the cache more often just to always be on the latest version. In fact, I'm not sure that it's nice to automatically be on the latest version because of the same reasons discussed before, even if breakage from newer Node.js versions should be very rare...

VoltrexKeyva:
@LinusU the main reason I think getting the latest version is better is because if we use the cached one, a scenario can occur where it lands on a Node.js version with regressions and/or critical issues and keeps using that, however always getting the latest version should avoid this problem as such issues in the runtime generally get fixed pretty quickly.

I also encountered this problem myself which was very annoying to deal with.

chearon:
The workflows are changing differently on the `next` branch

Issue #2232, opened 2023-04-19
🚢 3.0.0 planning issue


LinusU:
Things to consider for the 3.0.0 major version:

- [x] Drop support for old Node.js versions (14.x EOL at 2023-04-30, 16.x EOL at [2023-09-11](https://nodejs.org/en/blog/announcements/nodejs16-eol/))
- [ ] Change the `Canvas` class to [`OffscreenCanvas`](https://developer.mozilla.org/en-US/docs/Web/API/OffscreenCanvas)
    - [ ] Make our `OffscreenCanvas` extend [Node.js `EventTarget`](https://nodejs.org/api/events.html#eventtarget-and-event-api)
- [ ] Make `Image` and `ImageData` behave like browsers (using [node-gfx/image](https://github.com/node-gfx/image) & [node-gfx/image-data](https://github.com/node-gfx/image-data))
- [ ] Replace `simple-get` with [Node.js builtin `fetch`](https://nodejs.org/dist/latest-v20.x/docs/api/globals.html#fetch)
- [ ] Require [Pango 2](https://blogs.gnome.org/mclasen/2022/06/22/pango-1-90/) (if it's released in time)
    - [ ] New [`FontFace`](https://developer.mozilla.org/en-US/docs/Web/API/FontFace/FontFace) and [`document.fonts`](https://developer.mozilla.org/en-US/docs/Web/API/FontFaceSet) based API instead of `registerFont`
    - [ ] Non-standard `fillGlyphs(face, glyphIds)` API for fast rendering if shaping has already been done (e.g. harfbuzzjs)

Please note that this issue is for discussing **breaking** changes and not a place for general feature requests!

chearon:
Added Pango 2, which will solve a ton of font issues.

Should N-API be considered a breaking change? I plan on testing a lot, but it will introduce at least a few regressions.

LinusU:
> Added Pango 2, which will solve a ton of font issues.

Nice! 🙌 

> Should N-API be considered a breaking change? I plan on testing a lot, but it will introduce at least a few regressions.

Might be a good idea 👍

piranna:
[...]

PR #2229, opened 2023-04-13
src: refactor and apply fixes

VoltrexKeyva:
Thanks for contributing!

- [x] Have you updated CHANGELOG.md?

Issue #1954, opened 2021-12-22
Node-canvas dependency packages are too large to build project on Netlify


scott-schibli:
## Steps to Reproduce
Follow the installation guide.
I am running on OS Big Sur 11.5.2: and I had trouble initially using, so I ran the OS command for compiling : 

brew install pkg-config cairo pango libpng jpeg giflib librsvg.

I then import 'canvas' into a Nextjs Api Route to watermark images in a node environment before sending to the server (as I cannot load the non-watermarked images into the browser in any scenario. 

This works completely fine on my local machine - BUT when I try to push my project to netflify - which pushes to AWS, it is trying to install the package in a different environment. I assume this is totally normal, however, the dependency package for node-canvas named librsvg-2.so.2 is very large. It causes my project size to go over the maximum limit. 

I do not even need librsvg-2.so.2 (as in I do not need support for svg's). So I then prevent downloading this node dependency on build, which works to build the project and push it live. However, now when I call my api route the package crashes as its missing a dependency--Which I thought this was not a necessary dependency to node-canvas? Only an optional one? 

Any thoughts on this? Has anyone run into a similar situation trying to use node-canvas while hosting on netlify? 

Many thanks in advance!

spence:
Running into the same issue on Vercel. `ibrsvg-2.2.dylib` is 22MB and we're not using this. Removing this breaks the library.

benjamindell:
I have the same issue too. Did either of you two find a fix?

LinusU:
It seems like you are using the prebuilt version of node-canvas? If you build your own version you can make it only include the libraries you want. `jpeg`, `gif`, and `svg` are optional.

The prebuilt version includes support for all image formats as to be the most useful to most people.

iamdagy:
Hoy can I build my own later version ?  Thanks

iamdagy:
@LinusU I realized that I don't need the svg library is actually the biggest size with over 100 mb. or gif. but the question, is how do I remove them or install canvas without them. Im trying to use node-canvas in Netlify functions. and I dont have access to apt-get.. thanks a lot

benjamindell:
Does anyone have an example of how to build your own version of node-canvas? I've been googling for ages and cant seem to find anything concrete. I basically just need to build a version without SVG support but cant seem to find any examples :(

tourbillonlabs:
@scott-schibli Did you find a solution for this?

sshakyaUR:
[...]

Issue #1891, opened 2021-08-30
node-pre-gyp still needed?


jimmywarting:
if binaries are precompiled and downloaded, do we still need node-pre-gyp?
It seems to take up much of the total dependency: https://npmgraph.js.org/?q=canvas@2.8.0

maybe it can be optional somehow to install it?

LinusU:
I was looking at this earlier for another package as well, it's crazy how big node-pre-gyp is when all it does for the end user is to download some binaries 😅 

It also tries to require in `aws-sdk` and some other packages that aren't declared as dependencies which gives an error if you try to use webpack to bundle your code...

Would love some small package just for fetching the binaries! Maybe we should inquire about this upstream?

fuchunhui:
you have to try many times, otherwise node-pre-gyp is always waiting for you.

Brooooooklyn:
Here is an alternative package without any postinstall scripts and dependencies: https://github.com/Brooooooklyn/canvas

zbjornson:
v3.0.0-rc switches to prebuild-install, which drops the prebuild-related dependency size from 2 MB to 948 KB.

Right now the binaries are too large to include in the NPM tarball, but with quite a bit of work we might be able to get them small enough. (librsvg might make that difficult though.)

chearon:
The scope of SVG support is gigantic and impossible to do correctly, and I think people would be less sad if they actually avoided rendering SVGs at all. If we must have librsvg, we can support linking to librsvg in `npm rebuild`, but not ship it since it's so big. Or we could provide examples of how to use [canvg](https://github.com/canvg/canvg) in user-space. We'll need some SVG APIs for `Path2D`, so maybe we could use/contribute to [resvg](https://github.com/RazrFalcon/resvg).

Yesterday I started work on redoing the entire font stack to fix long-standing font issues, which will reduce dependencies even more. With all that, I think we would be able to then ship in the NPM tarball.

I would like us to move to a static build to simplify CI/prebuild hell. [Meson subprojects](https://mesonbuild.com/Subprojects.html) make doing that really easy, but sadly most of our dependencies haven't migrated to it yet. Maybe we'll have to add a `vendor/` directory like Firefox and Chrome do.

chearon:
The upcoming version of node-canvas ships binaries via NPM, so this is fixed 🎉 

```
npm install canvas@next
```

Edit: fixed via #2602

PR #2235, opened 2023-04-23
Port to node-addon-api (and N-API) and remove NAN, libuv, and v8

### LINKED ISSUE #923: N-API support for node-canvas
The recently announced Node 8 has a new experimental feature called [N-API](https://nodejs.org/dist/latest-v8.x/docs/api/n-api.html) which is aimed at reducing maintenance cost for node native addons. Checkout this [blog](https://medium.com/@nodejs/n-api-next-generation-node-js-apis-for-native-modules-169af5235b06) for more details on its benefits.

`node-canvas` is one of the top native addons by download dependency count, and in order to help the Node.js community make the important transition to N-API, we are hoping you will be able to work with us in order to migrate `node-canvas` to N-API. Your support and feedback is important in making this effort a success.

We (N-API team) have built [a preliminary port of `node-canvas` on top of N-API](https://github.com/jasongin/node-canvas/tree/napi) as part of our effort to understand the API surface used by native modules. We used this port and the port of some other modules to determine which areas to focus on converting to N-API. This port has all of the NAN and V8-specific code removed and replaced by N-API equivalents. While it passes all the canvas unit tests and visual tests, it might not meet the quality and performance level to go in as-is. It would certainly be a springboard if someone on `node-canvas` was serious about doing the conversion to N-API. Please take a look if you're interested. I will be happy to answer questions about the port, point to documentation/code, etc.

If you'd like, you may join our [WG meeting](https://plus.google.com/u/0/events/c0eevtrlajniu7h8cjrdk0f56c8?authkey=COH04YCalJS8Ug) which happens every Thursday at 1:30 Eastern / 10:30 Pacific US time, to discuss any issues.

We'd assume you don't want to immediately take a dependency on a feature that still has "experimental" status. The best case scenario from our point of view might be if the node-canvas team could eventually maintain a branch that uses N-API, that is kept mostly up to date with development occurring in node-canvas/master. But I understand you might not be ready for that just yet. Of course we'd appreciate feedback about anything we can do to improve the development or migration experience.

### LINKED ISSUE #1394: make module context-aware

chearon:
I've finished the conversion to N-API after a month of working on this. This should help a lot with prebuilds (my main motivation tbh) enable compatibility with Bun and Electron, etc.

Because this uses APIs from N-API v6, I've bumped the minimum node version to v10.20.0.

#### Changes

* Because `node-addon-api` uses C++ strings, I replaced some C strings with C++
* `CanvasRenderingContext2dInit` and `CanvasPatternInit` (sends JS values to C++) are now simply `bindings.setParseFont` and `bindings.setDOMMatrix`

#### Drive-bys

* I replaced `maybe.Check()` calls, which crash the whole process if it is empty, with `UnwrapTo` and `UnwrapOr`, so that in these rare cases node-canvas doesn't crash. In some instances, like when we're reading our own `ImageData`, I left the force-unwraps.

#### Performance 

* `node-addon-api` (C++ wrapper) and `napi` (ABI stable C library) incur two new layers of overhead when going between JS and C++. I've measured this to be an extra 0.2 µs per call function on my (2012) laptop. Users that call functions millions of times, for example, lots of `lineTo`s, will see performance degradation after this PR. In the `lineTo` case, we may be able to make a faster `Path2D` API if we pack the paths into a byte array.
* I included two perf commits to improve benchmarks a little bit

#### Future

* `node-addon-api` has the ability to throw C++ exceptions when a JS exception happens. It is far easier to correctly handle errors with this enabled (for example, we wouldn't need the `CHECK_RECEIVER` code) and we should think about turning that on. For this PR, I considered that change out of scope, so the green code still uses Maybes.
* We're closer to being able to close #1394. JS is all thread-safe now, but I don't know if all of the various C libraries we use are (and `font_face_list` is still global).

Fixes #923


gabrielschulhof:
@chearon I was thinking about this a little. Might it make sense to create an ArrayBuffer on the JS side so that the parameters to all the moveto and lineto and other API calls do not result in any C++ calls at all, but get recorded in the ArrayBuffer instead? Then, when the ArrayBuffer gets full or the user issues a call to render, one call is made to C++ which iterates over the ArrayBuffer and executes all the moveto/lineto/etc. API calls stored therein.

The size of the ArrayBuffer can be tuned by the user as part of initialization. We can pick a sensible default if the user chooses not to provide a value.

Most importantly, this would not be visible to the users at all, because we'd start the execution of the batched API calls whenever the user issued a render.

It might affect debugging though, in which case we'd document that, for debugging, the ArrayBuffer size must be zero. or one, which would result in today's behaviour.

src/JPEGStream.h
```diff
@@ -23,18 +23,15 @@ init_closure_destination(j_compress_ptr cinfo){
...
 empty_closure_output_buffer(j_compress_ptr cinfo){
-  Nan::HandleScope scope;
-  Nan::AsyncResource async("canvas:empty_closure_output_buffer");
   closure_destination_mgr *dest = (closure_destination_mgr *) cinfo->dest;
+  Napi::Env env = dest->closure->canvas->Env();
+  Napi::HandleScope scope(env);
+  Napi::AsyncContext async(env, "canvas:empty_closure_output_buffer");
 
-  v8::Local<v8::Object> buf = Nan::NewBuffer((char *)dest->buffer, dest->bufsize).ToLocalChecked();
+  Napi::Object buf = Napi::Buffer<char>::New(env, (char *)dest->buffer, dest->bufsize);
```
Julusian, line 31: Wrapping an existing pointer in a nodejs Buffer are not compatible with recent versions of electron.
See https://github.com/nodejs/node-addon-api/blob/main/doc/external_buffer.md
It looks like you might have some other occurrences of this too

src/CanvasRenderingContext2d.cc
```diff
@@ -76,100 +89,139 @@ inline static bool checkArgs(const Nan::FunctionCallbackInfo<Value> &info, doubl
...
+      Napi::Value pixelFormat;
+
+      if (ctxAttributes.Get("pixelFormat").UnwrapTo(&pixelFormat) && pixelFormat.IsString()) {
+        std::string utf8PixelFormat = pixelFormat.As<Napi::String>();
+        if (utf8PixelFormat == "RGBA32") format = CAIRO_FORMAT_ARGB32;
+        else if (utf8PixelFormat == "RGB24") format = CAIRO_FORMAT_RGB24;
+        else if (utf8PixelFormat == "A8") format = CAIRO_FORMAT_A8;
+        else if (utf8PixelFormat == "RGB16_565") format = CAIRO_FORMAT_RGB16_565;
+        else if (utf8PixelFormat == "A1") format = CAIRO_FORMAT_A1;
+#ifdef CAIRO_FORM
```
zbjornson, line 208:
```suggestion
#ifdef CAIRO_FORMAT_RGB30
```

zbjornson:
> Might it make sense to create an ArrayBuffer on the JS side so that the parameters to all the moveto and lineto and other API calls do not result in any C++ calls at all, but get recorded in the ArrayBuffer instead? Then, when the ArrayBuffer gets full or the user issues a call to render, one call is made to C++ which iterates over the ArrayBuffer and executes all the moveto/lineto/etc. API calls stored therein.

I happened to have tried that a while ago (super sloppily, just for benchmarking) in https://github.com/Automattic/node-canvas/commit/f2e073cdd322f2fee448438d0177b12fceef858c. That gave a 3.25x speedup in the NAN version and should give a larger speedup in the NAPI version because of the higher overhead. I'm not sure how it would actually work without explicit calls to render, since you can mix `moveTo`/`lineTo`, and as you said debugging would be weird.

However, even if it's not batched, copying the values to a TypedArray from JS helps: 2.3 Mops/sec -> 3.0 Mops/sec for `lineTo()`.

<details><summary>crude diff</summary>

```diff
diff --git a/lib/context2d.js b/lib/context2d.js
index 103ec63..082df97 100644
--- a/lib/context2d.js
+++ b/lib/context2d.js
@@ -9,3 +9,12 @@
 const bindings = require('./bindings')
 
 module.exports = bindings.CanvasRenderingContext2d
+
+bindings.CanvasRenderingContext2d.prototype.lineTo = function(x, y) {
+	if (isFinite(x) && isFinite(y)) {
+		const ab = this._argsBuffer;
+		ab[0] = x;
+		ab[1] = y;
+		this._lineTo();
+	}
+}
diff --git a/src/CanvasRenderingContext2d.cc b/src/CanvasRenderingContext2d.cc
index e1fb4b9..64c0c03 100644
--- a/src/CanvasRenderingContext2d.cc
+++ b/src/CanvasRenderingContext2d.cc
@@ -126,7 +126,7 @@ Context2d::Initialize(Napi::Env& env, Napi::Object& exports) {
     InstanceMethod<&Context2d::RoundRect>("roundRect"),
     InstanceMethod<&Context2d::MeasureText>("measureText"),
     InstanceMethod<&Context2d::MoveTo>("moveTo"),
-    InstanceMethod<&Context2d::LineTo>("lineTo"),
+    InstanceMethod<&Context2d::LineTo>("_lineTo"),
     InstanceMethod<&Context2d::BezierCurveTo>("bezierCurveTo"),
     InstanceMethod<&Context2d::QuadraticCurveTo>("quadraticCurveTo"),
     InstanceMethod<&Context2d::BeginPath>("beginPath"),
@@ -233,6 +233,10 @@ Context2d::Context2d(const Napi::CallbackInfo& info) : Napi::ObjectWrap<Context2
   states.emplace();
   state = &states.top();
   pango_layout_set_font_description(_layout, state->fontDescription);
+
+  _argsArray = Napi::Float64Array::New(env, 9);
+  info.This().As<Napi::Object>().Set("_argsBuffer", _argsArray);
+  args = _argsArray.Data();
 }
 
 /*
@@ -2572,10 +2576,6 @@ Context2d::setTextPath(double x, double y) {
 
 void
 Context2d::LineTo(const Napi::CallbackInfo& info) {
-  double args[2];
-  if(!checkArgs(info, args, 2))
-    return;
-
   cairo_line_to(context(), args[0], args[1]);
 }
 
diff --git a/src/CanvasRenderingContext2d.h b/src/CanvasRenderingContext2d.h
index 745106e..385e06d 100644
--- a/src/CanvasRenderingContext2d.h
+++ b/src/CanvasRenderingContext2d.h
@@ -220,4 +220,6 @@ class Context2d : public Napi::ObjectWrap<Context2d> {
     cairo_t *_context = nullptr;
     cairo_path_t *_path;
     PangoLayout *_layout = nullptr;
+    Napi::Float64Array _argsArray;
+    double* args = nullptr;
 };
```
</details>

Tedious to do for everything, and we'd want to hide stuff better.

gabrielschulhof:
@zbjornson sounds like it might be worth the effort, if it improves perf beyond just cancelling out the overhead incurred from switching to Node-API.

src/JPEGStream.h
chearon: Thanks for pointing that out. So I guess we can't claim Electron 21+ support yet. I'm going to consider that out of scope for this PR since we used external memory before these changes, but we should support Electron too without impacting Node performance. I wonder if Node will ever make the same change that Electron did too...

I'm so curious to see the use case for node-canvas in Electron. I thought you would just use `<canvas>`...

chearon:
Buffering all of the path calls is super interesting but I do agree that the flushing of the path is the downside. It has to happen in at least `clip()`, `fill()`, `stroke()`, `beginPath()` and `endPath()`. Path2D would be much more elegant to implement since you make all the Cairo calls right in `ctx.fill(path)`, but also looks harder, and still leaves `ctx.lineTo` slow.

> However, even if it's not batched, copying the values to a TypedArray from JS helps: 2.3 Mops/sec -> 3.0 Mops/sec

I am kind of bummed about this being less pretty but that's a good speedup and it's easy.

I wish `node-addon-api` could be optimized more (that's got to be `Napi::CallbackInfo` and `Napi::Number` construction) but I don't see much opportunity. Unless we could modify `InstanceMethod<&Context2d::LineTo>` to take a `LineTo` function that's just a low-level `napi_callback`. That would be nice...

src/JPEGStream.h
Julusian: > but we should support Electron too without impacting Node performance.

There is an alternate function `Napi::Buffer<char>::NewOrCopy(` designed for this case. Depending on whether you use `dest` again after this call, it may not require any more changes than that.

> I'm so curious to see the use case for node-canvas in Electron. I thought you would just use <canvas>...

That would depend on whether you want to use it on the backend or frontend of the application, but I agree it does sound like an usual case.

I have a couple of applications which use electron for packaging and to get a tray icon, nothing more. They are also possible to run in plain nodejs. It might technically be possible to use a canvas element, but it wouldnt be nice.
I found [a library for a usb device](https://github.com/foxxyz/loupedeck) which uses canvas for generating the images to display. I think that one is quite likely to be in electron instead of nodejs

GitMurf: Actually there is a HUGE use case in electron. It is why myself and many have been praying for node-canvas adoption of N-API ;) as mentioned above a big use case is because in electron you split everything into renderer and main threads. In main, where you do all your node stuff, you cannot use <canvas> and hence the need for node-canvas. Passing stuff between main and renderer is a pain (via IPC) so it is very useful to have node-canvas in main thread.

Lastly, and most important for me, Mozilla pdf-js relies on node-canvas and I use it for many things in my electron app. Without node-canvas napi compatibility I am not able to use pdf-js in newer versions of electron.

Does that make sense? Any questions?

Sakari369: Yeah we are also relying on node-canvas to handle rendering the contents of the <canvas> on the renderer side as a PDF or SVG in the main process. Node-canvas support for rendering to PDF/SVG is a great plus in this case, and would not be possible to do in the main thread.

Hoping that this would work with Electron versions 21+, but honestly if this already works with Electron version 20, that's a huge improvement already, so thank you for your work!

gabrielschulhof:
> I am kind of bummed about this being less pretty but that's a good speedup and it's easy.

I agree wholeheartedly. It may not be pretty, but node-canvas is definitely a performance-sensitive add-on.

> I wish `node-addon-api` could be optimized more (that's got to be `Napi::CallbackInfo` and `Napi::Number` construction) but I don't see much opportunity. Unless we could modify `InstanceMethod<&Context2d::LineTo>` to take a `LineTo` function that's just a low-level `napi_callback`. That would be nice...

I'll take a look at node-addon-api to see if we can provide variants which accept a napi_callback, thus avoiding the extrs construction and additional level of indirection. We did, after all, design node-addon-api to mix and match with plain Node-API.

src/Canvas.cc
```diff
@@ -227,132 +223,110 @@ Canvas::ToPngBufferAsync(uv_work_t *req) {
...
+    if (obj.Get("filters").UnwrapTo(&filters) && filters.IsNumber()) {
+      pngargs.filters = filters.As<Napi::Number>().Uint32Value();
+    }
 
-    Local<Value> palette = Nan::Get(obj, Nan::New("palette").ToLocalChecked()).ToLocalChecked();
-    if (palette->IsUint8ClampedArray()) {
-      Local<Uint8ClampedArray> palette_ta = palette.As<Uint8ClampedArray>();
-      pngargs.nPaletteColors = palette_ta->Length();
+    Napi::Value palette;
+    if (obj.Get("palette").UnwrapTo(&palette) && palette.IsTypedArray()) {
```
zbjornson, line 256: Need to check the type here, `palette.As<TypedArray>.TypedArrayType() == napi_uint8_clamped_array`.

src/Canvas.cc
```diff
@@ -227,132 +223,110 @@ Canvas::ToPngBufferAsync(uv_work_t *req) {
...
+      Napi::TypedArray palette_ta = palette.As<Napi::TypedArray>();
```
zbjornson, line 257:
```suggestion
      Napi::Uint8ClampedArray palette_ta = palette.As<Napi::Uint8ClampedArray>();
```

src/Canvas.cc
```diff
@@ -227,132 +223,110 @@ Canvas::ToPngBufferAsync(uv_work_t *req) {
...
+      pngargs.nPaletteColors = palette_ta.ElementLength();
       if (pngargs.nPaletteColors % 4 != 0) {
         throw "Palette length must be a multiple of 4.";
       }
+      pngargs.palette = static_cast<uint8_t*>(palette_ta.ArrayBuffer().Data());
```
zbjornson, line 262:
```suggestion
      pngargs.palette = palette_ta.Data();
```

(the typedarray's offset might not be 0)

src/Canvas.cc
```diff
@@ -557,69 +519,51 @@ streamPNG(void *c, const uint8_t *data, unsigned len) {
...
   // TODO this is technically wrong, we're returning a pointer to the data in a
   // vector in a class with automatic storage duration. If the canvas goes out
   // of scope while we're in the handler, a use-after-free could happen.
-  Local<Object> buf = Nan::NewBuffer(const_cast<char *>(reinterpret_cast<const char *>(data)), len, stream_pdf_free, 0).ToLocalChecked();
-  Local<Value> argv[3] = {
-      Nan::Null()
-    , buf
-    , Nan::New<Number>(len) };
-  async.runInAsyncScope(Nan::GetCurrentContext()->Global(), streaminfo->fn, sizeof argv / sizeof *argv, argv);
+  Napi::Object buf = Napi::Buffer<uint8_t>::New(env, (uint8_t *)(data), len);
```
zbjornson, line 565: Here and everywhere there's `Napi::Object xx = Napi::Buffer...`, including the return type of `Canvas::ToBuffer`, should be
```suggestion
  auto buf = Napi::Buffer<uint8_t>::New(env, (uint8_t *)(data), len);
```
or `Napi::Buffer<uint8_t>` to avoid casting to the superclass.

(Avoids an MSVC warning https://learn.microsoft.com/en-us/cpp/code-quality/C26437.)

src/Canvas.cc
```diff
@@ -958,9 +887,7 @@ Canvas::createCairoContext() {
...
 
-Local<Value>
-Canvas::Error(cairo_status_t status) {
-  return Exception::Error(Nan::New<String>(cairo_status_to_string(status)).ToLocalChecked());
+Napi::Error
+Canvas::CairoError(cairo_status_t status) {
+  return Napi::Error::New(env, cairo_status_to_string(status));
 }
-
-#undef CHECK_RECEIVER
```
zbjornson, line 966: Probably keep this line? The macro's name no longer reflects what it does either.

src/CanvasPattern.cc
```diff
@@ -4,122 +4,115 @@
...
+void
+Pattern::setTransform(const Napi::CallbackInfo& info) {
+  if (!info[0].IsObject()) {
+    Napi::TypeError::New(env, "Expected DOMMatrix").ThrowAsJavaScriptException();
+    return;
+  }
+
+  Napi::Object mat = info[0].As<Napi::Object>();
 
 #if NODE_MAJOR_VERSION >= 8
```
zbjornson, line 93: Don't need the guards now that the min version is 10.something.

src/CanvasPattern.cc
```diff
@@ -4,122 +4,115 @@
...
 }
 
 /*
  * Initialize a new CanvasPattern.
  */
 
-NAN_METHOD(Pattern::New) {
-  if (!info.IsConstructCall()) {
-    return Nan::ThrowTypeError("Class constructors cannot be invoked without 'new'");
+Pattern::Pattern(const Napi::CallbackInfo& info) : ObjectWrap<Pattern>(info), env(info.Env()) {
```
zbjornson, line 36: A crash happens with `new canvas.CanvasPattern()` or `ctx.createPattern()`.

src/CanvasPattern.cc
```diff
@@ -4,122 +4,115 @@
...
+    return;
+  }
+
+  Napi::Object mat = info[0].As<Napi::Object>();
 
 #if NODE_MAJOR_VERSION >= 8
-  if (!mat->InstanceOf(ctx, _DOMMatrix.Get(Isolate::GetCurrent())).ToChecked()) {
-    return Nan::ThrowTypeError("Expected DOMMatrix");
+  InstanceData* data = env.GetInstanceData<InstanceData>();
+  if (!mat.InstnaceOf(data->DOMMatrix.Instance)) {
```
zbjornson, line 95:
```suggestion
  if (!mat.InstanceOf(data->DOMMatrixCtor.Value()).UnwrapOr(false)) {
```

zbjornson (changes requested):
Got through half of the files so far. Looks great, NAPI is so much nicer to work with.

src/Canvas.cc
chearon: The name still reflects the purpose, not so much the way it's done anymore.

I think I'm going to wait until another node-addon-api release so we can just delete the checks.

gabrielschulhof:
@chearon
> I'll take a look at node-addon-api to see if we can provide variants which accept a napi_callback, thus avoiding the extra construction and additional level of indirection. We cannot avoid the unwrap though, because we need to retrieve the pointer to the native instance. We did, after all, design node-addon-api to mix and match with plain Node-API.
I actually don't think we'd gain too much from being able to provide a `napi_callback` to an `InstanceMethod`, because `napi_get_cb_info()` needs to get called no matter what, because we need to retrieve the instance whose method is getting called. The call to `napi_get_cb_info()` is the heavy part. Now, FWIW, if any of the methods take more than 6 arguments, then there's likely need for optimization, because `Napi::CallbackInfo` has room on the stack for 6 arguments, but if the function takes more than that, it calls `napi_get_cb_info()` twice, after it allocates the required room for args:

https://github.com/nodejs/node-addon-api/blob/358b2d3b4f4cf0b3bdfc9897be888494ee36ae1d/napi-inl.h#L3745-L3766

If needed, we may want to expose the number of statically allocated arguments as a preprocessor directive:

https://github.com/nodejs/node-addon-api/blob/358b2d3b4f4cf0b3bdfc9897be888494ee36ae1d/napi.h#L1875

gabrielschulhof:
@chearon actually, it looks like napi_unwrap takes the cake: http://gabrielschulhof.github.io/misc/templated.instance.method.perf.breakdown.svg. That is not something easily changed, unfortunately.

src/Canvas.cc
chearon: Does that make a difference? We're accessing the underlying buffer and the `ElementLength()`, which should be the same if it's a `TypedArray` or a `Uint8ClampedArray`.

src/Canvas.cc
chearon: `Napi::Value` should work too if I understand correctly.

I don't know our stance on `auto` but I prefer seeing the types most of the time. Maybe I should set up type checking in my editor.

src/Canvas.cc
chearon: > I think I'm going to wait until another node-addon-api release so we can just delete the checks.

`node-addon-api` 7.0 was released with my fix, but they also bumped the minimum node version to 16. We can remove like 100 lines and this PR is marked as semver-major anyways, so I'm going to go forward with it...

src/CanvasPattern.cc
chearon: Wow, this typo indicates that the macro above (`NODE_MAJOR_VERSION`) was not working

src/CanvasPattern.cc
chearon: I'm not able to reproduce any crash. I get `Uncaught TypeError: Image or Canvas expected` for both of those, which is correct. I also tried passing a real image, and verified that is painting correctly.

chearon:
@gabrielschulhof thank you for looking into it. I was suggesting a direct call to an `napi_function` because it looked like the invocation of `Napi::details::TemplatedInstanceVoidCallback` and family was causing significant extra time. It doesn't look that way in your flame chart. `ellipse` takes more than 6 arguments and would be the only one I'm concerned about.

I think we should merge this and do Path2D after we're sure performance is a problem. 

@zbjornson huge thanks for the detailed look. I think I got everything.

The `CHECK_RECEIVER` calls are all gone thanks to `node-addon-api` v7. That version claims to bump the minimum node version to 16, but I think that is wrong, at least for our purposes. There are no commits that look like they are adding new ABI calls that would break it. All of the tests pass with node 10.

zbjornson:
Cool, I'll try to get through the remaining review this weekend.

Sakari369:
Hey, just wondering, any update on this getting merged ? :) 
Or can I already use this branch, should work without the final changes ?

GitMurf:
Any update on plans for when this will be "officially" released? Thanks much for the hard work!

Sakari369:
> I'm satisfied with it as-is, but would rather @zbjornson or @LinusU give an approval before merging. And anyone else that wants to review/test would be helping a lot!
> 
> @Sakari369 you can use this branch temporarily, it passes all of our tests.

Thank you for the info, will test it out! 🙏

src/CanvasPattern.cc
zbjornson:
```
> canvas = createCanvas(1,1)
[Canvas 1x1]
> canvas.getContext("2d").createPattern()
# Exception thrown at 0x0000000068DF09A6 (libcairo-2.dll) in node.exe: 0xC0000005: Access violation reading location 0xFFFFFFFFFFFFFFFF
```

Turns out a crash happens with the same code in master also, but with a different cause. Can deal with later.

src/Canvas.cc
```diff
@@ -392,146 +361,137 @@ static void setPdfMetadata(Canvas* canvas, Local<Object> opts) {
...
-      closure = static_cast<SvgBackend*>(canvas->backend())->closure();
+      closure = static_cast<SvgBackend*>(backend())->closure();
     }
 
-    cairo_surface_finish(canvas->surface());
-    Local<Object> buf = Nan::CopyBuffer((char*)&closure->vec[0], closure->vec.size()).ToLocalChecked();
-    info.GetReturnValue().Set(buf);
-    return;
+    cairo_surface_finish(surface());
+    return Napi::Buffer<uint8_t>::New(env, &closure->vec[0], closure->vec.size());
```
zbjornson, line 386: I think you need to continue to copy here, `Napi::Buffer<uint8_t>::Copy(...)`.

zbjornson (changes requested):
Great, just a few minor comments.

src/Canvas.cc
zbjornson: I think it lets you avoid the `.As<Uint8Array>()` on line 256R. Using the specific type also means code using `ByteLength()` would just work in the future.

src/Canvas.cc
zbjornson: `Napi::Value` is still an ancestor class. Only `auto` or `Napi::Bufer<uint8_t>` avoid the implicit cast.

zbjornson:
Also - this fixes a few WPT tests (`yarn test-wpt`) but breaks several others. The ones broken in this branch are:
```
  WPT: path-objects
    42) 2d.path.arc.shape.3
    43) 2d.path.arc.selfintersect.1
    45) 2d.path.arc.zeroradius
    46) 2d.path.rect.selfintersect
    47) 2d.path.roundrect.zero.6
    48) 2d.path.roundrect.selfintersect

  WPT: shadows
    78) 2d.shadow.attributes.shadowBlur.invalid # this one looks like an int/uint issue
```

src/Canvas.cc
chearon: The link above is about [slicing](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-slice), which is slightly different than an implicit cast, but I see what you're saying. But wouldn't there be a slice happening every time you return a more derived type in a node-addon-api callback, which must have a return type of `Napi::Value`? It is designed to be used this way, so I'm not sure if it's really an issue. I don't see that MSVC warning on CI either.

src/Canvas.cc
```diff
@@ -763,27 +702,32 @@ NAN_METHOD(Canvas::RegisterFont) {
...
       pango_font_description_free(user_desc);
-      Nan::ThrowError("Could not load font to the system's font host");
+      Napi::Error::New(env, "Could not load font to the system's font host").ThrowAsJavaScriptException();
+
     }
   } else {
     pango_font_description_free(user_desc);
-    Nan::ThrowError(GENERIC_FACE_ERROR);
+    if (!env.IsExceptionPending()) {
+      Napi::Error::New(env, GENERIC_FACE_ERROR).ThrowAsJavaScriptException();
```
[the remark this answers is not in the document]
chearon: `NODE_API_SWALLOW_UNTHROWABLE_EXCEPTIONS` is not enabled currently, so these are needed. I'll look into adding it though, I didn't know about that.

chearon:
I do get a different number of failures (123 on `master`, 121 here), but the ones you listed are all failing for me on `master` too. Are you getting a difference in kinds of errors? I don't know much about those tests, but I'll fix anything they uncover.

src/Canvas.cc
zbjornson: You're right, maybe I was looking at 281L still. I guess you can't call `TypedArrayType()` without converting to the `TypedArray` super first.

src/Canvas.cc
zbjornson: Implicit cast *to a super class*, right. It's not actually an issue because the only members in the derived class are retrieved from V8 when needed. Fine to ignore.

src/Canvas.cc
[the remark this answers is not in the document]
zbjornson: Good eyes. Fine by me if you want to keep the checks since the fatal exception is meant to reveal bugs.

src/CanvasRenderingContext2d.cc
```diff
@@ -1479,330 +1492,325 @@ NAN_GETTER(Context2d::GetGlobalCompositeOperation) {
...
-  int n = Nan::To<double>(value).FromMaybe(0);
-  if (n >= 0) {
-    Context2d *context = Nan::ObjectWrap::Unwrap<Context2d>(info.This());
-    context->state->shadowBlur = n;
-  }
+void
+Context2d::SetShadowBlur(const Napi::CallbackInfo& info, const Napi::Value& value) {
+  Napi::Number n;
+  if (value.ToNumber().UnwrapTo(&n) && n.DoubleValue() >= 0) state->shadowBlur = n.DoubleValue();
 }
```
zbjornson, line 1660:
```
  78) WPT: shadows
       2d.shadow.attributes.shadowBlur.invalid:

      AssertionError [ERR_ASSERTION]: ctx.shadowBlur
      + expected - actual

      --2147483648
      +1
      
      at Context.<anonymous> (test\wpt\generated\shadows.js:130:10)
```

```
ctx.shadowBlur = 1;
ctx.shadowBlur = Infinity;
assert.strictEqual(ctx.shadowBlur, 1, "ctx.shadowBlur", "1")
```

Converting any floating type value to an integer, where the value can't be represented by the integer, is undefined behavior. That might be why you're seeing it fail on master and I'm not.

The canvas spec says:

> context.[shadowBlur](https://html.spec.whatwg.org/multipage/canvas.html#dom-context-2d-shadowblur) [ = value ]
>
>    Can be set, to change the blur level. **Values that are not finite numbers greater than or equal to zero are ignored.**

This would work:

```cpp
  Napi::Number n;
  if (value.ToNumber().UnwrapTo(&n)) {
    double v = n.DoubleValue();
    if (v >= 0 && v <= std::numeric_limits<decltype(state->shadowBlur)>::max()) {
      state->shadowBlur = v;
    }
  }
```

(Using `Int32Value()` does *not* work; that returns 0.)

src/CanvasRenderingContext2d.cc
```diff
@@ -53,12 +44,30 @@ constexpr double twoPi = M_PI * 2.;
...
+    napi_valuetype type;
+    double val = 0;
+
+    napi_typeof(env, argv[i], &type);
+    if (type == napi_number) {
+      // fast path
+      napi_get_value_double(env, argv[i], &val);
+    } else {
+      napi_value num;
+      napi_create_double(env, 0, &num);
```
zbjornson, line 66: I think this line can be deleted.

zbjornson (commented):
Added a comment with the fix for the one new failing WPT test.

The others went away after I pulled and rebuilt. The value of `Math.PI/2` passed from JS was somehow very slightly different in the NAN and N-API versions. I went back to the earlier commits in this PR and couldn't reproduce it. 😞 

```
  42) WPT: path-objects
       2d.path.arc.shape.3:

      AssertionError [ERR_ASSERTION]: Red channel of the pixel at (1, 48)
      + expected - actual

      -242
      +0
      
      at _assertPixel (test\wpt\generated\path-objects.js:57:9)
      at Context.<anonymous> (test\wpt\generated\path-objects.js:1439:3)
      at processImmediate (node:internal/timers:466:21)
```

This branch before pull/rebuild:
image
master:
image

```
-1.5707963267948966  master
... something I didn't write down that ended in 7
```

---

With the fix below, 3 new WPT tests are passing in this branch. Nice work.

zbjornson (approved):
Awesome job!

LinusU:
Absolutely amazing work @chearon! 🤩

matthieu-beteille:
Hi 👋 ! Thanks for the PR! When can we expect this to be released?

chearon:
It's marked as semver-major, but there aren't breaking changes here. IMO we should do a minor release for this soon since it now works with Bun and seems to solve a lot of other issues.

LinusU:
> I've bumped the minimum node version to v10.20.0.

I think this is a breaking change?

chearon:
Oh right, I guess we were supporting v6 before this, but we didn't have CI tests for it. Just don't think this should be held up by everything in #2232, but maybe we could get some of the easier stuff done soon.

zbjornson:
We could do this as 3.0 and the other things in 2232 as 4.0+. I think the nonzero perf impact is worth keeping this semver-major.

I can kick off the prebuilds if y'all agree on 3.0?

matthieu-beteille:
3.0 sounds great to me!

zbjornson:
I forgot that we need to adjust prebuilds/node-pre-gyp to work with N-API (https://github.com/mapbox/node-pre-gyp#node-api-considerations). I don't have time to figure that out today. If anyone else has time, please go ahead.

chearon:
Oh, it should be assignable since it's assignable in the browser. If you/someone wants to make a PR I can review it, or I can get around to it sometime soon.

abrenoch:
Gotcha @chearon, good to hear! I would love to help but I have just about no experience with with C++ - unfortunately the best I can do at the moment is complain 😄 

Thanks again!

GitMurf:
@abrenoch good catch! Given my main use of node-canvas is for usage with pdfjs, this is key for me as well :) thanks!

GitMurf:
@chearon as an FYI, I do not know if this is expected or not, but the maintainer of pdf-to-img mentioned here https://github.com/k-yle/pdf-to-img/issues/228#issuecomment-1758790659 that they were not able to compile v3 pre-release and provided the error they received. Probably worth taking a look just to make sure it isn't something needed to be fixed before v3 release.

zbjornson:
Update on prebuilds:
- node-pre-gyp requires `{napi_build_version}` to be in the `module_path`, which means that canvas.node needs to live in something like build/Release/napi-7/canvas.node. That's annoying: gyp doesn't let you specify the output directory, so you need to use a post-install hook to copy the file there, and a bunch of lines will need updating when we change NAPI versions.

- A DIY-approach is doable, but there are a handful of details to get right (e.g. proxy support). It would be nice to get rid of node-pre-gyp since we only use like 10% of its functionality (#1891).

    <details><summary>rough script</summary>

    ```js
    const build_from_source = process.argv.includes("--build-from-source")
        && !process.argv.includes("--build-from-source=false");

    function getBinStr() {
        const module_name = require("./package.json").name;
        const version = require("./package.json").version;
        const napi_build_version = "7";
        const platform = process.platform; // TODO support --target_platform
        const arch = process.arch; // TODO support --target_arch
        const libc = require("detect-libc").family;
        return `${module_name}-v${version}-napi-v${napi_build_version}-${platform}-${libc}-${arch}.tar.gz`;
    }

    if (build_from_source) {
        throw new Error("--build-from-source specified, skipping prebuild download.");
    } else {
        // TODO Need to support proxies
        // TODO Need to support --canvas_binary_host_mirror
        const get = require("simple-get");
        const base_url = "https://github.com/Automattic/node-canvas/releases/download";
        const bin_str = getBinStr();
        console.log(`Downloading prebuild for ${bin_str}`);
        get.concat(`${base_url}/${bin_str}`, (err, res, data) => {
            if (err) {
                throw err;
            }
            if (res.statusCode !== 200) {
                console.error(`Download failed: ${res.statusCode}`);
                throw err;
            }
            // TODO untar and save data to build/Release/canvas.node
            return;
        });
    }
    ```
    </details>

- Other modules seem to be moving away from node-pre-gyp in favor of prebuild and prebuild-install (e.g. [sqlite3](https://github.com/TryGhost/node-sqlite3/commit/605c7f981abbcf980cb44f54abba285de9a69a23)). It's a lot smaller than node-pre-gyp, if nothing else. prebuild-install has a fixed package name string format that looks like `canvas-v{version}-napi-v{napi_version}-{platform}{libc}-{arch}.tar.gz`, where `{libc}` is omitted if it's `glibc`.

- The size of all three platforms' bundles is 73.5 MB, and will grow if/when we support ARM and musl. That's probably(?) too big to put in the npm tarball. librsvg alone is too big for some PaaSes like Netlify (#1954).

The tip commits on the master and prebuilds branches switch to prebuild-install, but the build failed for the rsvg reason above: https://github.com/Automattic/node-canvas/actions/runs/7352791664/job/20017985691 -- so it remains untested until that's fixed. If we don't want to switch to prebuilds, feel free to force-push over my commits.

I might have time to keep working on this tomorrow. Input welcome on approaches.

chearon:
Supporting librsvg alone is as much of a pain as building to every version of node was before N-API. It has so many `Dockerfile` dependencies. Would removing it from prebuilds allow us to publish the binaries in the NPM tarball? I'm sure there will be pitchforks, but people could still manually compile, or we could remove it in favor of a JS/WASM dependency, implement `Path2D`, etc.

zbjornson:
> Would removing it from prebuilds allow us to publish the binaries in the NPM tarball?

That shrinks it from 73.5 MB to 40-50 MB. The AWS Lambda ZIP upload size limit is 50 MB; Netlify I think is the same because it uploads to Lambda; Netlify Edge is 20 MB; Good Cloud Functions 2nd gen has no limit. So, maybe/it depends.

chearon:
If you look at the "Make Bundle" section of the prebuild action, it looks like we are distributing `.o` files that total ~7MiB per platform as well. When I delete that and `librsvg-2.2.dylib` (not including any unreachable dependencies after that) the gzipped bundle on macOS is 4.9 MiB (down from 10.4 MiB).

If Linux and Windows get down to a number like that, I think it would be worth distributing in the NPM tarball and either remove SVG from prebuilds or start shipping it as a JS or WASM dependency.

(I would rather remove it from prebuilds than distribute it any way, which is kind of insane: that's an entire layout engine with a CSS engine, XML parser, etc. I think there are good technical reasons not to do so, but I don't expect users to care when they see SVG support "removed").

kotasudhakar:
Please when can we expect this to be released?
This is a real problem as most of the libraries today uses the old Nan api and have node canvas as a dependency, which makes any electron app having canvas as dependency not work at all!

Sakari369:
> Please when can we expect this to be released? This is a real problem as most of the libraries today uses the old Nan api(pdfjs, react-pdf, etc.,) and have node canvas as a dependency, which makes any electron app having canvas as dependency not work at all!

I somehow got the newer node-canvas building on newer electron, both on Windows and macOS, both for arm64 and intel x64.

But it's not trivial, and requires quite the fiddly setup to work, the most trouble I've had is been tweaking these builds to work. If you are interested I can maybe share what worked for me, but to be honest at some point it just worked, and I dont even entirely sure know why.

zbjornson:
Had some time to work on this finally, but eesh are prebuilds miserable to maintain. Lots of new build breaks:
- One of the macOS casks changed dependencies and created a conflict with another cask.
- Maybe related, `pip3 install --user macpack` stopped working
- A recent node-canvas PR added use of a C++20 feature (fixed)
- Something changed with libseccomp-dev on Linux. pax-utils is failing to build and I don't know why. We're using the same git SHA of pax-utils that we've used before, and we're in a Docker container that I thought was supposed to provide a (more) fixed environment.

I only got one of those fixed, which at least made the Windows prebuild work. Any help on the other issues would be welcome.

kotasudhakar:
Can we make a release until this, for now... And whenever we get to fix the issue for linux and Mac, you can make one more release

zbjornson:
@kotasudhakar no, sorry. I haven't gotten back to the actual prebuild distribution part yet (the switch away from node-pre-gyp).

chearon:
I hope to pick up where @zbjornson left off on prebuilds, but it's the last thing I want to do in my free time. We could release under a beta tag without prebuilds for electron/other users in the meantime.

GitMurf:
> I hope to pick up where @zbjornson left off on prebuilds, but it's the last thing I want to do in my free time. We could release under a beta tag without prebuilds for electron/other users in the meantime.

@chearon I hear ya! I wish I could help but I just have no experience in this domain.

Doesn't electron require the prebuilds though? Or is that essentially what electron-rebuild does where it runs that instead of using a prebuild? Sorry for my lack of knowledge in this area.

zbjornson:
Progress today:
1. Linux: Avoided the pax-utils issue by installing the built package instead of building it from source.
2. Linux: Update the librsvg version from 2.50.2 to 2.52.8. #2229 switched to an API that was introduced in 2.52.

Linux and Windows now build, but we still have the installation part to do. I might work on that later today. Also want to look into making prebuilds from the 2.x NAN branch for newer Node.js versions.

I am completely new to homebrew, so hoping someone familiar with it can work on the Mac build. I've been fumbling around with the tip commits in https://github.com/Automattic/node-canvas/commits/prebuilds/.

zbjornson:
Alright, v3.0.0-rc2 is available for testing, with prebuilds available on Linux x64 and Windows x64.

https://github.com/Automattic/node-canvas/releases/tag/v3.0.0-rc2

zbjornson:
macOS x64 prebuilds are up and ready for testing. I didn't realize that `macos-latest` now means M1.

`npm install canvas@next`

PR #2245, opened 2023-05-16
Add M1 prebuilds

davidanthoff:
I have no idea whether this will work, but maybe worth a try? I don't have a M1, so I can't even test this :)

But it _does_ build something when I run the workflow on my fork, and the compiler log has output that suggests it is actually building an arm64 version. The upload step fails in my fork, and I am not familiar enough with that to fix this. Maybe a maintainer here could test things and see whether this might actually be producing the correct binaries?

### DIFF
--- a/.github/workflows/prebuild.yaml
+++ b/.github/workflows/prebuild.yaml
@@ -175,6 +175,7 @@ jobs:
           . prebuild/macOS/bundle.sh
 
       - name: Test binary
+        if: matrix.arch == 'x64'
         run: |
           brew uninstall --force --ignore-dependencies cairo pango librsvg giflib harfbuzz
           npm test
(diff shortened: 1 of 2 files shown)


chearon:
I don’t think it could be this easy because all of the dependencies have to be in the right architecture too.

davidanthoff:
Ah, true...

Is the most realistic option here to get a Mac mini, and add that as a self-hosted runner? At least until GitHub offers M-chip hosted runners? I have a downstream project that is very much stuck on the lack of prebuild M1 binaries here, and it might be the easiest for me to just get a Mac mini and (if you are ok with that) connect it to this repo here?

antonio-iodice:
Any clue on how this could be picked up again? Is this something anyone can do?

jamesbvaughan:
This should be possible now that GitHub offers Apple silicon runners, right?

https://github.blog/2023-10-02-introducing-the-new-apple-silicon-powered-m1-macos-larger-runner-for-github-actions/

chearon:
Merged @jamesbvaughan's PR, so this is obsolete. Thanks for the effort anyways!

PR #2309, opened 2023-11-27
Replace fetch api

YuMuuu:
Use fetch api and remove simple-get depedency (context: https://github.com/Automattic/node-canvas/pull/2223#issuecomment-1496177541).

---

### DIFF
--- a/lib/image.js
+++ b/lib/image.js
@@ -14,9 +14,6 @@ const bindings = require('./bindings')
 const Image = module.exports = bindings.Image
 const util = require('util')
 
-// Lazily loaded simple-get
-let get
-
 const { GetSource, SetSource } = bindings
 
 Object.defineProperty(Image.prototype, 'src', {
@@ -47,20 +44,20 @@ Object.defineProperty(Image.prototype, 'src', {
           }
         }
 
-        if (!get) get = require('simple-get')
-
-        get.concat({
-          url: val,
+        fetch(val, {
+          method: 'GET',
           headers: { 'User-Agent': 'Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/102.0.0.0 Safari/537.36' }
-        }, (err, res, data) => {
-          if (err) return onerror(err)
-
-          if (res.statusCode < 200 || res.statusCode >= 300) {
-            return onerror(new Error(`Server responded with ${res.statusCode}`))
-          }
-
-          setSource(this, data)
         })
+          .then(res => {
+            if (!res.ok) {
+              throw new Error(`Server responded with ${res.statusCode}`)
+            }
+            return res.arrayBuffer()
+          })
+          .then(data => {
+            setSource(this, Buffer.from(data))
+          })
+          .catch(onerror)
       } else { // local file path assumed
         setSource(this, val)
       }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 (Unreleased)
 ==================
 ### Changed
+* Replaced `simple-get ` with ` Node.js builtin` `fetch` (#2309)
 ### Added
 ### Fixed
 
--- a/package.json
+++ b/package.json
@@ -46,8 +46,7 @@
   ],
   "dependencies": {
     "node-addon-api": "^7.0.0",
-    "prebuild-install": "^7.1.1",
-    "simple-get": "^3.0.3"
+    "prebuild-install": "^7.1.1"
   },
   "devDependencies": {
     "@types/node": "^10.12.18",


YuMuuu:
I will make sure that the relevant PR is made 👀 
I rebase if necessary

YuMuuu:
> Thanks, that's a nice change. https://github.com/Automattic/node-canvas/pull/2310 will need to land first to drop support for older versions of Node.js.
Ok. I can rebaes after merged #2310

YuMuuu:
https://github.com/Automattic/node-canvas/pull/2309/commits/c0725b896e47641e4c02615bbf57f4c6672bfba7

This commit is important, Should I separate PR?

lib/image.js
```diff
@@ -46,20 +43,22 @@ Object.defineProperty(Image.prototype, 'src', {
             throw err
           }
         }
-
-        if (!get) get = require('simple-get')
-
-        get.concat({
-          url: val,
+        const request =  fetch(url = val, data = {
```
LinusU, line 46: This creates two new global variables, `url` and `data`. I'm guessing it was a misstake

```suggestion
        const request =  fetch(val, {
```

lib/image.js
```diff
@@ -46,20 +43,22 @@ Object.defineProperty(Image.prototype, 'src', {
...
+          method: "POST",
```
LinusU, line 47: Can you use single quotes to align with the rest of the file?

```suggestion
          method: 'POST',
```

lib/image.js
```diff
@@ -46,20 +43,22 @@ Object.defineProperty(Image.prototype, 'src', {
...
           headers: { 'User-Agent': 'Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/102.0.0.0 Safari/537.36' }
-        }, (err, res, data) => {
-          if (err) return onerror(err)
+        })
 
+        request.then((res) => {
```
LinusU, line 51: Since `request` is only used once, you could chain here and remove the variable:

```suggestion
        })
        .then((res) => {
```

lib/image.js
```diff
@@ -46,20 +43,22 @@ Object.defineProperty(Image.prototype, 'src', {
...
           if (res.statusCode < 200 || res.statusCode >= 300) {
             return onerror(new Error(`Server responded with ${res.statusCode}`))
           }
-
-          setSource(this, data)
+          return res.arrayBuffer()
+        })
+        .then((data) => {
+          return setSource(this, data)
```
LinusU, line 58: `setSource` doesn't return anything?

```suggestion
          setSource(this, data)
```

lib/image.js
```diff
@@ -46,20 +43,22 @@ Object.defineProperty(Image.prototype, 'src', {
...
+        })
+        .catch((err) => {
+          return onerror(err)
```
LinusU, line 61: Same here.

```suggestion
          onerror(err)
```

Also, the behaviour of `onerror` is that it throws if a handler hasn't been set. Before this change, that thrown was delegated to the [global error handler](https://nodejs.org/docs/latest-v20.x/api/process.html#event-uncaughtexception).

Now it will instead be eaten by this promise, which I don't think is desirable.

I think that `util.callbackify` or `process.nextTick` would be best used to restore the old behaviour 🤔

LinusU:
@YuMuuu could you rebase on latest `master`? I've merged #2310 and #2313 which are alternative ways of dealing with the problems you had with `dtslint` and minimum version of Node...

YuMuuu:
- rebase master and drop unnecessary commir

lib/image.js
YuMuuu: => https://github.com/Automattic/node-canvas/pull/2309/commits/93cfd2f012cf40bbf0d7a7f88f60b9c3659106a2

lib/image.js
YuMuuu: => https://github.com/Automattic/node-canvas/pull/2309/commits/93cfd2f012cf40bbf0d7a7f88f60b9c3659106a2

lib/image.js
YuMuuu: => https://github.com/Automattic/node-canvas/pull/2309/commits/93cfd2f012cf40bbf0d7a7f88f60b9c3659106a2

lib/image.js
YuMuuu: => https://github.com/Automattic/node-canvas/pull/2309/commits/93cfd2f012cf40bbf0d7a7f88f60b9c3659106a2

lib/image.js
YuMuuu: => https://github.com/Automattic/node-canvas/pull/2309/commits/93cfd2f012cf40bbf0d7a7f88f60b9c3659106a2

I'm not sure if the changes I made are correct...

lib/image.js
```diff
@@ -46,21 +43,30 @@ Object.defineProperty(Image.prototype, 'src', {
...
-        get.concat({
-          url: val,
-          headers: { 'User-Agent': 'Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/102.0.0.0 Safari/537.36' }
-        }, (err, res, data) => {
-          if (err) return onerror(err)
-
+        util
+        .callbackify(
+          fetch(val, {
+            method: 'POST',
```
YuMuuu: I'm not sure if the changes I made are correct.

simple-get.concat method type is GET.

YuMuuu: => https://github.com/Automattic/node-canvas/pull/2309/commits/1eb26cf66bf2b8c3a35c7e02a2da0dd230248e24

YuMuuu:
@LinusU This is just a reminder if you missed my commit. Could you please take a look at my commits I’ve done last week and merge it if it looks ok.

YuMuuu:
rebase master

YuMuuu:
👀

YuMuuu:
@LinusU  Could you please review this PR? ;;

YuMuuu:
@zbjornson Could you review, as it seems that @LinusU is busy?

kunipon:
@LinusU Could you please let me know when the pull request is expected to be merged?

LinusU:
Sorry for being a bit slow on the feedback here...

Looking at the code, I don't think that it works? Have you tested this?

YuMuuu:
We have not been able to check all the test cases that are not on CI. What kind of test cases do you think are not working?

chearon:
I squashed the commits and removed `require('node-fetch')`, which I couldn't justify. Also spent a while thinking about @LinusU's [comment](https://github.com/Automattic/node-canvas/pull/2309#discussion_r1407771904) but I'm pretty sure it doesn't apply to the current diff. Throwing in a `catch` handler still goes to that nodejs event. Though it does get scheduled differently than before, I don't think that should be considered a breaking change.

chearon (approved):

YuMuuu:
Thank you very much for reviewing and merging and releasing it!
I'm grateful for the node-cavas and contributors.

PR #2310, opened 2023-11-27
Drop support for older versions of Node.js

LinusU:
Part of breaking changes discussed in #2232, this changes to align with LTS-supported Node.js versions

This changes the minimum supported versions of Node.js:
- 16 and earlier = unsupported
- 17 = unsupported
- 18 = supported from 18.12.0 (LTS)
- 19 = unsupported
- 20 = supported from 20.9.0 (LTS)
- 21 and later = not tested, but should work

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -14,4 +14,4 @@ var ctx = canvas.getContext('2d');
 
 ## Your Environment
 * Version of node-canvas (output of `npm list canvas` or `yarn list canvas`):
-* Environment (e.g. node 4.2.0 on Mac OS X 10.8):
+* Environment (e.g. node 20.9.0 on macOS 14.1.1):
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -68,6 +68,7 @@ jobs:
           brew update
           brew install python3 || : # python doesn't need to be linked
           brew install pkg-config cairo pango libpng jpeg giflib librsvg
+          pip install setuptools
       - name: Install
         run: npm install --build-from-source
       - name: Test
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node: [10, 12, 14, 16, 18, 20]
+        node: [18.12.0, 20.9.0]
     steps:
       - uses: actions/setup-node@v3
         with:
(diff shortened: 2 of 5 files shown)


chearon (approved):

zbjornson:
```
 Traceback (most recent call last):
  File "/Users/runner/hostedtoolcache/node/20.9.0/x64/lib/node_modules/npm/node_modules/node-gyp/gyp/gyp_main.py", line 42, in <module>
    import gyp  # noqa: E402
    ^^^^^^^^^^
  File "/Users/runner/hostedtoolcache/node/20.9.0/x64/lib/node_modules/npm/node_modules/node-gyp/gyp/pylib/gyp/__init__.py", line 9, in <module>
    import gyp.input
  File "/Users/runner/hostedtoolcache/node/20.9.0/x64/lib/node_modules/npm/node_modules/node-gyp/gyp/pylib/gyp/input.py", line 19, in <module>
    from distutils.version import StrictVersion
ModuleNotFoundError: No module named 'distutils'
```

I had to fix this error in another package recently by adding this before `npm install`:
```
- run: pip install setuptools
```

LinusU:
macOS test failures are fixed, but getting a strange error on Windows with Node.js 20:

>  1) Canvas
>       registerFont:
>     Error: ENOENT: no such file or directory, lstat 'D:\a\node-canvas\node-canvas\examples\pfennigFont\pfennigMultiByte🚀.ttf'
>      at Object.realpathSync (node:fs:2631:7)
>      at registerFont (index.js:51:34)
>      at Context.<anonymous> (test\canvas.test.js:116:5)
>      at process.processImmediate (node:internal/timers:478:21)


I don't think it's related to the changes here though, just that we are now testing on Node.js 20.

We could allow failures for Windows/Node.js 20 for now? Or if you have any ideas on how to fix it @chearon? I don't have access to a windows machine myself 🙈 so it's going to be hard for me to debug...

LinusU:
I think that this is the issue:

https://github.com/nodejs/node/issues/48673

It seems like it might be fixed with this PR:

https://github.com/nodejs/node/pull/50650

Which hopefully lands in the next Node.js 20.x release.

Removing the 20.x test for Windows in the meantime.

-----

Also, I think that we should make sure that this project is included in the [node citgm](https://github.com/nodejs/citgm) so that new Node.js releases gets tested against Canvas. In that case this regression would have been caught...

Generally, I think that this approach (that is, test on lowest supported versions in our CI, and be included in CITGM to test latest Node.js version) is the best approach for larger projects like this, since I think that it's more likely that we accidentally introduce code that only works on newer versions of Node.js than our lowest supported, rather than Node.js introducing a breaking change...

PR #2313, opened 2023-11-28
Replace dtslint with tsd

LinusU:
Thanks for contributing!


LinusU:
Since `dtslint` was deprecated, and their replacement `@definitelytyped/dtslint` was on a 0.0.x version, and since they also recommended `tsd`, I thought it was better to use `tsd` over `@definitelytyped/dtslint`.

This also simplifies the entire setup a bit, with the potential drawback that the types files isn't linted for stylistic changes (but this should probably be handled by Standard anyways)

package.json
```diff
@@ -43,12 +44,13 @@
   },
   "files": [
     "binding.gyp",
+    "browser.js",
+    "index.d.ts",
+    "index.js",
```
LinusU, line 49: Potentially, this aren't needed, but [the `files` documentation](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#files) only mentions `main`, and I think it's better to just include them all...

zbjornson (approved):
👍 cool

PR #2322, opened 2023-12-06
fix the wrong type of setTransform

huankong233:
Thanks for contributing!

fix the wrong type of setTransform

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 ### Changed
 ### Added
 ### Fixed
+* fix the wrong type of setTransform
 
 3.0.0
 ==================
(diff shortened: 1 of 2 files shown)


chearon:
Our implementation doesn't take the raw coordinates yet:

https://github.com/Automattic/node-canvas/blob/ad793dab1fd2fd64bd8e9c55381679423b840236/src/CanvasRenderingContext2d.cc#L2321

huankong233:
> Our implementation doesn't take the raw coordinates yet:
> 
> https://github.com/Automattic/node-canvas/blob/ad793dab1fd2fd64bd8e9c55381679423b840236/src/CanvasRenderingContext2d.cc#L2321

but my code works well with this type 😕
code ↓
~~~ typescript
    const canvas = createCanvas(800, 250)
    ctx.setTransform(1, 0, -0.4, 1, 0, 0)
    ctx.font = '80px sans-serif'
    ctx.fillStyle = 'red'
    ctx.fillText('testText', 250, 200)
    ctx.resetTransform()
    return canvas.toBuffer('image/png')
~~~

image

there is a test file,I thought someone is forget this ?
https://github.com/Automattic/node-canvas/blob/ad793dab1fd2fd64bd8e9c55381679423b840236/test/canvas.test.js#L1046

chearon (approved):
Aw, you're right. I misread the `else` path.

PR #2328, opened 2023-12-27
added deregisterAllFonts to readme

stepancar:
Added description for deregisterAllFonts to readme.md

It does not change an implementation and does not require to publish this change to npm

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -78,6 +78,8 @@ This project is an implementation of the Web Canvas API and implements that API 
 * [createImageData()](#createimagedata)
 * [loadImage()](#loadimage)
 * [registerFont()](#registerfont)
+* [deregisterAllFonts()](#deregisterAllFonts)
+
 
 ### Non-standard APIs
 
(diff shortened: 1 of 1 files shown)


Readme.md
```diff
@@ -170,6 +172,35 @@ ctx.fillText('Everyone hates this font :(', 250, 10)
 
 The second argument is an object with properties that resemble the CSS properties that are specified in `@font-face` rules. You must specify at least `family`. `weight`, and `style` are optional and default to `'normal'`.
 
+### deregisterAllFonts()
+
+> ```ts
+> deregisterAllFonts() => void
+> ```
+
+Use deregisterAllFonts If you registered font and need to remove it, for example if you use canvas in tests.
```
michaeljherrmann, line 181:
```suggestion
Use deregisterAllFonts if you registered font and need to remove it, for example if you use canvas in tests.
```

michaeljherrmann: hmm perhaps this description makes it sound like you can deregister a specific font? I believe the api will deregister all previously added fonts

stepancar: @michaeljherrmann do you think it's better?

'Use `deregisterAllFonts` to unregister all fonts that have been previously registered. This method is useful when you want to remove all registered fonts, such as when using the canvas in tests'

stepancar, line 181:
```suggestion
Use `deregisterAllFonts` to unregister all fonts that have been previously registered. This method is useful when you want to remove all registered fonts, such as when using the canvas in tests
```

stepancar: updated

zbjornson (approved):
Thanks!

PR #2340, opened 2024-01-24
Update simple-get to 4.0.1

nikolas:
Previous versions of simple-get are vulnerable to CVE-2022-0355.

https://nvd.nist.gov/vuln/detail/CVE-2022-0355

Thanks for contributing!


LinusU:
Our usage of simple-get is not vulnerable. We cannot upgrade without a breaking change, so this will be fixed in the next major.

More info: https://github.com/Automattic/node-canvas/pull/2223#issuecomment-1496177541

Issue #2343, opened 2024-02-02
textBaseline top middle and hanging do not align with javascript canvas api


semohr:
## Issue or Feature
The textBaseline property does not produce the expected results. I noticed a misalignment between renders of the JavaScript Canvas api and the node-canvas implementation. 

It seems like only the `alphabetic` and `bottom` option align with the JavaScript canvas api whereas `top`, `middle` and `hanging` do not. 

Is that an expected behavior? If yes how can I align text from both implementations?

## Steps to Reproduce
I overlayed the JavaScript and node-canvas results of the following scripts. Red text is the expected value using the canvas api and blue text is the results from the node-canvas package.
![canvas](https://github.com/Automattic/node-canvas/assets/39738318/afdaa91e-56ab-4f8e-869f-19299f354a8d)

### Node script
```js
import { createCanvas } from "canvas";
import { writeFileSync } from "fs";

function main() {
    const canvas = createCanvas(480, 500);
    const ctx = canvas.getContext("2d");

    ctx.font = "50px Arial";
    ctx.globalAlpha = 0.5;
    ctx.fillStyle = "blue";

    ["top", "hanging", "middle", "alphabetic", "bottom"].forEach((baseline, index) => {
        ctx.textBaseline = baseline;
        ctx.fillText(baseline, 10, index * 90 + 10);

        ctx.beginPath();
        ctx.moveTo(0, index * 90 + 10);
        ctx.lineTo(480, index * 90 + 10);
        ctx.stroke();
    });

    const buf = canvas.toBuffer("image/png", { compressionLevel: 3 });
    writeFileSync("./canvas_node.png", buf);
}

main();
```

### HTML
```html
<!DOCTYPE html>

<body>
    <canvas id="canvas" width="480" height="500"></canvas>

    <script>
        var canvas = document.getElementById("canvas");
        var ctx = canvas.getContext("2d");

        ctx.font = "50px Arial";
        ctx.globalAlpha = 0.5;
        ctx.fillStyle = "red";

        ["top", "hanging", "middle", "alphabetic", "bottom"].forEach((baseline, index) => {
            ctx.textBaseline = baseline;
            ctx.fillText(baseline, 10, index * 90 + 10);

            ctx.beginPath();
            ctx.moveTo(0, index * 90 + 10);
            ctx.lineTo(480, index * 90 + 10);
            ctx.stroke();
        });

        // Get buffer from canvas and convert to png
        var buffer = canvas.toDataURL("image/png");
[...]

PR #2347, opened 2024-02-13
Added HANGING, IDEOGRAPHIC and TOP textBaseline to adjustments

semohr:
see #2343

### DIFF
--- a/src/CanvasRenderingContext2d.cc
+++ b/src/CanvasRenderingContext2d.cc
@@ -2479,20 +2479,34 @@ inline double getBaselineAdjustment(PangoLayout* layout, short baseline) {
   PangoRectangle logical_rect;
   pango_layout_line_get_extents(pango_layout_get_line(layout, 0), NULL, &logical_rect);
 
-  double scale = 1.0 / PANGO_SCALE;
-  double ascent = scale * pango_layout_get_baseline(layout);
-  double descent = scale * logical_rect.height - ascent;
+  double ascent = pango_layout_get_baseline(layout);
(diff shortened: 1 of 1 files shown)


chearon:
We do support `textBaseline = "top"`. It's a no-op since the Pango layout is painted from the top-left.

The math for the existing alignments checks out. I can't explain why you got different results, as they look right when I try them (I used our own browser comparisons from `npm run test-server` to check). Does it still happen with different fonts? What OS did you try it on?

We can't do hanging or ideographic correctly because Pango doesn't expose that. I think it would be better to leave approximations up to the user for now.

semohr:
As noted in the issue: https://github.com/Automattic/node-canvas/issues/2343

I'm using
canvas@2.11.2 on Ubuntu 22.04.3 LTS (jammy)
node@v18.12.1
firefox@127.0.1
Arial font

Well as far as I can see the test might not be working correctly and the resolution is kinda low in these tests.
Please try the code snippet from the issue, the top alignment produces visible different results depending on if one uses the browser canvas implementation versus the node-canvas implementation.

semohr:
Just to be sure I tried it again with the newest version (master branch f138b3a667c50935d3e1ffa69ed0fb56f0f2877c) as I submitted this quite some time ago. I picked another font and made sure to load exactly the same font files both in the browser and in the node implementation.  The problem seems to persists.


image1101

semohr:
Noticed this PR was still open. While the issue remains in the current version, I’m closing it since I’ve migrated from node-canvas to skia-canvas and no longer have a personal stake in this.

PR #2366, opened 2024-03-25
[cleanup] feat(prebuild): support of Alpine binaries

Delagen:
- update dependencies
- rework binding.gyp
- rework msys bundle for Windows
- add bundle for MUSL (Alpine)
- rework bundle scripts and ci

Thanks for contributing!

- [ ] Have you updated CHANGELOG.md?

PR #2370, opened 2024-03-28
[prebuild] feat(prebuild): support of Alpine binaries

delagen:
- update dependencies
- rework binding.gyp
- rework msys bundle for Windows
- add bundle for MUSL (Alpine)
- rework bundle scripts and ci

moved to prebuild

Thanks for contributing!


Delagen:
Moved from #2366

Delagen:
@zbjornson  reworked. This is second.

ifurther:
Hi I had the suggest for your comment. You should split the comment for different system.

Delagen:
Reworked to support 22 version, remove error suppression in Windows test due to fix in 20.x Node

prebuild/Debian/preinstall.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env sh
+
+sudo apt-get update
+sudo apt-get install -y build-essential libcairo2-dev libjpeg-dev libpango1.0-dev libgif-dev librsvg2-dev libpixman-1-dev patchelf
```
zbjornson, line 4: @chearon did you build all of the Linux deps from source for a reason, or can we use precompiled packages as this PR does?

chearon: Yes, we need to build on an old OS for maximum compatibility. Programs built against a newer libc may have symbols that don't exist on older OSes (this happened to real users). But we want newer libraries, so have to build from source.

zbjornson: > But we want newer libraries

👍 forgot about this part.

chearon:
The `next` branch for v4 changes the build in a way that diverges totally from this. I did make sure we added binaries for musl. v4 can be tried by doing `npm install canvas@next`.

Issue #2300, opened 2023-11-02
Most RGB color functions are not interpreted correctly


haykam821:
- [x] If this is an issue with installation, I have read the [troubleshooting guide](https://github.com/Automattic/node-canvas/issues/1511).

From what I can tell, [the specification for `rgb()` and `rgba()`](https://drafts.csswg.org/css-color/#rgb-functions) allows for variations including the use of whitespace, commas, forward slash separators, percentages, and decimals, but the color parsing here seems to only allow for only a few specific formats. This issue creates an inconsistency where transparency is rendered as opaque compared to a browser environment.

## Steps to Reproduce

Compare the results of running the following snippet in a browser and in Node.js with `canvas`:

```js
const canvas = "document" in globalThis ? document.createElement("canvas") : require("canvas").createCanvas();
const context = canvas.getContext("2d");

const entries = [
	"rgb(0, 0, 0)",
	"rgba(0, 0, 0)",
	"rgb(0 0 0)",
	"rgba(0 0 0)",
	"rgb(0, 0, 0, 25%)",
	"rgba(0, 0, 0, 25%)",
	"rgb(0, 0, 0, 0.25)",
	"rgba(0, 0, 0, 0.25)",
	"rgb(0 0 0 / 25%)",
	"rgba(0 0 0 / 25%)",
	"rgb(0 0 0 / 0.25)",
	"rgba(0 0 0 / 0.25)",
];

for (const entry of entries) {
	context.fillStyle = entry;
	console.log(context.fillStyle);
}
```

Value | Browser (Expected) | Node.js (Actual)
--- | --- | ---
`rgb(0, 0, 0)` | `#000000` | `#000000`
`rgba(0, 0, 0)` | `#000000` | `#000000`
`rgb(0 0 0)` | `#000000` | `#000000`
`rgba(0 0 0)` | `#000000` | `#000000`
`rgb(0, 0, 0, 25%)` | `rgba(0, 0, 0, 0.25)` | `#000000`
`rgba(0, 0, 0, 25%)` | `rgba(0, 0, 0, 0.25)` | `#000000`
`rgb(0, 0, 0, 0.25)` | `rgba(0, 0, 0, 0.25)` | `#000000`
`rgba(0, 0, 0, 0.25)` | `rgba(0, 0, 0, 0.25)` | `rgba(0, 0, 0, 0.25)`
`rgb(0 0 0 / 25%)` | `rgba(0, 0, 0, 0.25)` | `#000000`
[...]

PR #2384, opened 2024-05-18
Fix rgba parsing (#2300)

pranav1344:
Fixes issues related for rgba parsing (#2300). Browser allows percentage signs in alpha, as well as / as separator. I've also added test cases for the same.


pranav1344:
The issue (#2300) should be fixed after these changes. Also added tests for the cases listed in #2300.

zbjornson (approved):
Excellent, thank you!

PR #2386, opened 2024-05-24
Allow alternate or properly escaped quotes *within* font-family names

virtulis:
This fixes the issue with font names that themselves contain quotes. This appears to be legal (or at least perfectly working) in the browser implementations, and after this fix Pango renders the correct font as well.

Importantly, font names are now split into an array, trimmed of quotes, unescaped, and joined back. This may have unexpected consequences beyond existing test cases, but I haven't come up with any.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@ This release notably changes to using N-API. 🎉
 * Fix usage of garbage value by filling the allocated memory entirely with zeros if it's not modified. (#2229)
 * Fix a potential memory leak. (#2229)
 * Fix the wrong type of setTransform
+* Fix the improper parsing of rgb functions issue. (#2300)
 
 2.11.2
 ==================


lib/parse-font.js
```diff
@@ -46,6 +49,12 @@ module.exports = str => {
   const sizeFamily = sizeFamilyRe.exec(str)
   if (!sizeFamily) return // invalid
 
+  const names = sizeFamily[3]
+    .match(familyRe)
+    // remove actual bounding quotes, if any, unescape any remaining quotes inside
+    .map(s => s.trim().replace(unquoteRe, '$1').replace(unescapeRe, '$1'))
+    .filter(s => !!s)
+
```
chearon, line 57: Isn't that the same as this? We already know it matches `familyRe`:

```suggestion
   // remove actual bounding quotes, if any, unescape any remaining quotes inside
  const names = sizeFamily[3].trim().replace(unquoteRe, '$1').replace(unescapeRe, '$1'))
```

chearon (commented):
Looks good, thanks. Just some small suggestions.

lib/parse-font.js
chearon: Oh my bad, now I see it. It's for comma-separated fonts.

chearon (approved):

chearon:
Oops... I squashed and moved the `CHANGELOG.md` item to underneath `3.0.0` since we haven't released it yet, but I accidentally reset your branch when I pushed it. Manually added the commit to `master`: ae1aacb72a292f20cacdbe324d04524da8e01a6c

PR #2394, opened 2024-07-09
fix TextMetrics type to include `alphabeticBaseline`, `emHeightAscent`, and `emHeightDescent` properties

musou1500:
Thanks for contributing!


chearon:
Looks good, thanks. I'm going to squash the commits and then move the changelog item to 3.0.0 since we haven't released it yet.

> `fontBoundingBoxAscent` and `fontBoundingBoxDescent` is not implemented but listed in TextMetrics interface.
Should I include fix in this PR?

I don't think Pango exposes what we'd need for those 2 properties. You'd need to know the first font that matched based on the `PangoFontDescription` and also that font's ascender and descender. I'm happy to review if you can figure that out, though.

chearon (approved):

chearon (approved):

Issue #2044, opened 2022-05-26
Feature: Insert Hyperlink in PDF


jeremywillans:
When rendering a PDF, is it possible to add the ability to insert a hyperlink when adding either an image or text?

Thanks!

zbjornson:
Cairo (the backend used by node-canvas) supports it, but there's no Canvas API that exposes it. We try to avoid non-standard APIs, but this seems like a highly useful and minimally invasive one to add.

PR #2405, opened 2024-07-17
Add link tags for pdfs

mcfedr:
Interested to see if people think this is a reasonable implementation.

I went with as simple as possible based off https://www.cairographics.org/manual/cairo-Tags-and-Links.html#link

I know its not standard api, but there isn't one for this in canvas, and as was mentioned in the linked issue the project should be open to such an addition.

Fixes https://github.com/Automattic/node-canvas/issues/2044

### DIFF
--- a/src/CanvasRenderingContext2d.cc
+++ b/src/CanvasRenderingContext2d.cc
@@ -19,6 +19,7 @@
 #include <string>
 #include "Util.h"
 #include <vector>
+#include <iostream>
 
 /*
  * Rectangle arg assertions.
@@ -134,6 +135,10 @@ Context2d::Initialize(Napi::Env& env, Napi::Object& exports) {
     InstanceMethod<&Context2d::CreatePattern>("createPattern", napi_default_method),
     InstanceMethod<&Context2d::CreateLinearGradient>("createLinearGradient", napi_default_method),
     InstanceMethod<&Context2d::CreateRadialGradient>("createRadialGradient", napi_default_method),
+    #if CAIRO_VERSION >= CAIRO_VERSION_ENCODE(1, 16, 0)
+    InstanceMethod<&Context2d::BeginTag>("beginTag", napi_default_method),
+    InstanceMethod<&Context2d::CloseTag>("closeTag", napi_default_method),
+    #endif
     InstanceAccessor<&Context2d::GetFormat>("pixelFormat", napi_default_jsproperty),
     InstanceAccessor<&Context2d::GetPatternQuality, &Context2d::SetPatternQuality>("patternQuality", napi_default_jsproperty),
     InstanceAccessor<&Context2d::GetImageSmoothingEnabled, &Context2d::SetImageSmoothingEnabled>("imageSmoothingEnabled", napi_default_jsproperty),
@@ -418,7 +423,7 @@ Context2d::fill(bool preserve) {
         width = cairo_image_surface_get_width(patternSurface);
         height = y2 - y1;
       }
-      
+
       cairo_new_path(_context);
       cairo_rectangle(_context, 0, 0, width, height);
       cairo_clip(_context);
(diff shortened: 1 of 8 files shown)


src/Canvas.cc
```diff
@@ -382,7 +382,15 @@ Canvas::ToBuffer(const Napi::CallbackInfo& info) {
       closure = static_cast<SvgBackend*>(backend())->closure();
     }
 
-    cairo_surface_finish(surface());
+    cairo_surface_t *surf = surface();
+    cairo_surface_finish(surf);
+
+    cairo_status_t status = cairo_surface_status(surf);
+    if (status != CAIRO_STATUS_SUCCESS) {
```
mcfedr, line 389: I know this isn't really part of the PR, but it was very helpful to be in testing this, and I thought be helpful for others

src/CanvasRenderingContext2d.cc
```diff
@@ -3354,3 +3358,29 @@ Context2d::Ellipse(const Napi::CallbackInfo& info) {
...
+ * Open and close a link tag
+ */
+
+void
+Context2d::BeginTag(const Napi::CallbackInfo& info) {
+  Napi::String strValue;
+  if (!info[0].ToString().UnwrapTo(&strValue)) return;
+
+  std::string str = strValue.Utf8Value();
+  std::string uri = "uri='" + str + "'";
```
zbjornson, line 3374: Wonder if we need some sort of escaping or validity testing to make sure `str` doesn't contain `'` or any other impermissible characters?

Cairo's docs say URI needs to be an ASCII string; what happens if it's not?

zbjornson (commented):
This is sort of like `addHitRegion()` (RIP) or the proposed `HitRegionSet` (https://github.com/whatwg/html/issues/3407).

If we add a non-standard API (generally avoided), we should make sure it covers all foreseeable use cases to avoid having to add another non-standard API in the future. In this case, we might want to (prepare to) expose the full Cairo API for tags (i.e. support specifying an array of rects as the hit region).

+0 from me.

src/CanvasRenderingContext2d.cc
mcfedr: yep, escaping `'` and `\` is needed.

I'd probably guess things are going to crash with a non-ascii string, its actually where this is likely very helpful, https://github.com/Automattic/node-canvas/pull/2410

mcfedr:
> This is sort of like addHitRegion() (RIP) or the proposed HitRegionSet (https://github.com/whatwg/html/issues/3407).

These proposals seem to mostly be dead or nearly dead, I think the best thing would just to avoid being close to them in any naming

> If we add a non-standard API (generally avoided), we should make sure it covers all foreseeable use cases to avoid having to add another non-standard API in the future

I agree, I think would be good to be able to support the different tag types that cairo can support.

[cairo_tag_begin](https://www.cairographics.org/manual/cairo-Tags-and-Links.html#cairo-tag-begin)

I see a couple of options

- api like in cairo, that takes two strings, tag_name and attributes - then this simple use case is a bit harder, as the user has to provide the tag name 'Link' and the uri=... attributes

- provide some sort of conversion of the attributes from an object into the format expected by cairo, this would be more work now

- maybe we assume that second param can be an object in the future, but for now take a string, and assume that its a uri
  
- if we want the use to provide the tag name, i.e. `"Link"` do we want to expose the const `CAIRO_TAG_LINK` as a const on `Context2d` - but that would be more non-standard API surface.

mcfedr:
@zbjornson I've made some changes to make it simple for this API to be extended in the future.

I've kept only support for Link with uri, and leave it as a task for someone who needs it to add support for more types, but with the object based config for tag, this should be simple to do.

I've also added escaping and check for ascii.

chearon (commented):
Yes, sorry for the long wait. I finally found time today to read the Cairo documentation and the PDF specs. I agree with @zbjornson that we should plan for all of the abilities that PDF structure gives us, so the current changes are in the right direction.

However, I think we should do a thin abstraction here. It's ever-so-slightly more performant not to have users put things into JS objects, only to serialize it and have Cairo deserialize it again. It will also make the implementation dead simple. It's less ergonomic when you're specifying rects or destinations, but those uses should both be pretty rare. We should be reporting `CAIRO_STATUS_TAG_ERROR` at some point, which helps.

So the API would look more like:

```js
ctx.beginTag("Link"); // or "Link", "rect=[0, 0, 100, 10] uri='https://cairographics.org'"
ctx.fillText("Cairo!", 0, 0);
ctx.endTag();
```

And if you want to define PDF structure:

```js
ctx.beginTag("H1");
ctx.fillText("Cairo!", 0, 0);
ctx.endTag();
```

I don't mind being so tied to Cairo because if we were to switch it out (which will probably never happen) there would be other breaking changes anyways.

Thoughts?

chearon:
I wanted to get this landed and I don't have permission to write to this branch, so I landed it as da33bbed88946188385af6dc10368410ffede365.

Changes I made:
- Refactored to remove object abstraction. This reduced code a ton. I verified that bad syntax will throw `Error: invalid tag name, attributes, or nesting` when you call `toBuffer`. The Cairo syntax isn't as easy as what you had here, but I updated the README to link to the documentation, and I think it's better that we leave any stringifying up to the user/other libraries.
- Combined examples into one
- Fixed output path of pdf to examples directory
- Removed unused `iostream` and `cairoVersion` imports
- Removed ASCII checks. This would be nice to report, but Cairo happily accepts non-ASCII and [it doesn't sound like](https://datatracker.ietf.org/doc/html/rfc2396#autoid-9) URLs are forbidden to use non-ASCII. Let's leave it up to the user (also faster this way). 
- Renamed `closeTag` to `endTag` (`end` goes with `begin` and matches the Cairo name)

Issue #2390, opened 2024-06-20
Problem with RC and jsdom


brettz9:
## Issue or Feature
- [x] If this is an issue with installation, I have read the [troubleshooting guide](https://github.com/Automattic/node-canvas/issues/1511).

Not sure whether you'll treat this as a jsdom issue (in case jsdom has to change its API or something), but with the RC release, jsdom no longer works.

## Steps to Reproduce

With jsdom version 24.1.0
```js
import jsdom from 'jsdom';

const {window: {document}} = new jsdom.JSDOM('');

const canvas = document.createElement('canvas');
const ctx = canvas.getContext('2d');
```

...gives:

```
/Users/brett/typeson-registry/node_modules/.pnpm/jsdom@24.1.0_canvas@3.0.0-rc2/node_modules/jsdom/lib/jsdom/living/nodes/HTMLCanvasElement-impl.js:116
  ctx[name] = function (image, ...rest) {
            ^

TypeError: Cannot assign to read only property 'createPattern' of object '#<CanvasRenderingContext2D>'
    at wrapNodeCanvasMethod (/Users/brett/typeson-registry/node_modules/.pnpm/jsdom@24.1.0_canvas@3.0.0-rc2/node_modules/jsdom/lib/jsdom/living/nodes/HTMLCanvasElement-impl.js:116:13)
    at HTMLCanvasElementImpl.getContext (/Users/brett/typeson-registry/node_modules/.pnpm/jsdom@24.1.0_canvas@3.0.0-rc2/node_modules/jsdom/lib/jsdom/living/nodes/HTMLCanvasElement-impl.js:35:11)
    at HTMLCanvasElement.getContext (/Users/brett/typeson-registry/node_modules/.pnpm/jsdom@24.1.0_canvas@3.0.0-rc2/node_modules/jsdom/lib/jsdom/living/generated/HTMLCanvasElement.js:131:58)
    at file:///Users/brett/typeson-registry/.idea/canvas.js:6:20
    at ModuleJob.run (node:internal/modules/esm/module_job:195:25)
    at async ModuleLoader.import (node:internal/modules/esm/loader:337:24)
    at async loadESM (node:internal/process/esm_loader:34:7)
    at async handleMainPromise (node:internal/modules/run_main:106:12)

Node.js v18.20.3
```

## Your Environment
* Version of node-canvas (output of `npm list canvas` or `yarn list canvas`): 3.0.0-rc2
[...]

PR #2409, opened 2024-07-18
Fix class properties should have defaults as standard js classes would have

mcfedr:
Fixes https://github.com/Automattic/node-canvas/issues/2390

jsdom tries to replace some functions

in a normal js canvas, all the props are like standard js classes

this change adds the `napi_property_attributes` for this

also some `InstanceValue` are replaced with `StaticValue` as this is how they should have been

This has changed in the 3.x RC with the move to napi


src/Canvas.cc
```diff
@@ -50,25 +50,25 @@ Canvas::Initialize(Napi::Env& env, Napi::Object& exports) {
...
-    InstanceValue("PNG_FILTER_AVG", Napi::Number::New(env, PNG_FILTER_AVG)),
-    InstanceValue("PNG_FILTER_PAETH", Napi::Number::New(env, PNG_FILTER_PAETH)),
-    InstanceValue("PNG_ALL_FILTERS", Napi::Number::New(env, PNG_ALL_FILTERS)),
-    StaticMethod<&Canvas::RegisterFont>("_registerFont"),
-    StaticMethod<&Canvas::DeregisterAllFonts>("_deregisterAllFonts")
+    InstanceAccessor<&Canvas::GetType>("type", napi_default_jsproperty),
+    InstanceAccessor<&Canvas::GetStride>("stride", napi_default_jsproperty),
+    InstanceAccessor<&Canvas::GetWidth, &Canvas::SetWidth>("width", napi_default_jsproperty),
+    InstanceAccessor<&Canvas::GetHeight, &Canvas::SetHeight>("height", napi_default_jsproperty),
+    StaticValue("PNG_NO_FILTERS", Napi::Number::New(env, PNG_NO_FILTERS), napi_default_jsproperty),
```
chearon, line 63: Changing these to a static value is a breaking change. This goes all the way back to 2013: 14d9f587b6106dd900cabe0d6bc31117acb596c6.

I think it's more correct to put on the class, but it definitely needs to be a documented change for 3.0.0, and the types need to be updated accordingly.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 ### Changed
 ### Added
 ### Fixed
+* Fix class properties should have defaults as standard js classes would have (#2390)
```
chearon, line 13: You don't need to document this since it hasn't been released yet. (I guess technically it has as an alpha version, but we don't have an entry in this for that, so...).

chearon (changes requested):
The descriptor changes look right 👍

CHANGELOG.md
mcfedr: So remove this? I added because of the PR template "Have you updated CHANGELOG.md?"

src/Canvas.cc
mcfedr: I've added the types for this and a test that uses these

CHANGELOG.md
zbjornson: Looks good where you moved it to now. 👍

chearon: I meant delete since it cancels itself out (broken in 3.0.0 and fixed in... 3.0.0) but nbd

src/Canvas.cc
```diff
@@ -382,7 +382,15 @@ Canvas::ToBuffer(const Napi::CallbackInfo& info) {
...
-    cairo_surface_finish(surface());
+    cairo_surface_t *surf = surface();
+    cairo_surface_finish(surf);
+
+    cairo_status_t status = cairo_surface_status(surf);
+    if (status != CAIRO_STATUS_SUCCESS) {
+      Napi::Error::New(env, cairo_status_to_string(status)).ThrowAsJavaScriptException();
+      return env.Undefined();
+    }
+
```
chearon, line 393: Whoops, this is in #2410 too, so if I merged this first, that PR wouldn't be mergeable.

chearon (commented):
All looking good except it has changes from #2410

chearon (approved):
Thank you!

PR #2410, opened 2024-07-18
Throw surface errors in toBuffer

mcfedr:
I noticed this when developing https://github.com/Automattic/node-canvas/pull/2405 

we can check cairo_surface_status for errors, which I've noticed happening when I was making tags incorrectly, but thought this is actually quite helpful, as overwise you just get en empty buffer, and have no idea why it happened.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ This release notably changes to using N-API. 🎉
 * Replaced `dtslint` with `tsd` (#2313)
 ### Added
 * Added string tags to support class detection
+* Throw Cairo errors in canvas.toBuffer()
 ### Fixed
 * Fix a case of use-after-free. (#2229)
 * Fix usage of garbage value by filling the allocated memory entirely with zeros if it's not modified. (#2229)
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 ==================
 ### Changed
 ### Added
+* Throw errors in canvas
```
chearon, line 12: Can you move this to 3.0.0 (sorry, everything's in kind of a weird state) and:

```suggestion
* Throw errors in canvas.toBuffer()
```

chearon (commented):
Not sure if this has any other real-world usages other than the development scenario you mentioned, but seems reasonable.

mcfedr:
> Not sure if this has any other real-world usages other than the development scenario you mentioned, but seems reasonable.

Yes, hopefully it isnt possible to do in normal usage, but getting an file with just a PDF header and nothing else was very confusing when it happened!

chearon (approved):

Issue #2197, opened 2023-02-16
Image class type in index.d.ts incompatible with HTMLImageElement


keean:
## Issue or Feature
- [x] If this is an issue with installation, I have read the [troubleshooting guide](https://github.com/Automattic/node-canvas/issues/1511).

Writing TypeScript code to run in both browser and backend has been broken with relatively recent changes. 

In types/index.d.ts, the following changes are needed:
```patch
--- a/node_modules/canvas/types/index.d.ts
+++ b/node_modules/canvas/types/index.d.ts
@@ -324,10 +327,10 @@ export class Image {
 	 * can embed JPEGs directly into the output, rather than re-encoding into
 	 * PNG. This can drastically reduce filesize and speed up rendering.
 	 */
-	dataMode: number
+	dataMode?: number
 
-	onload: (() => void) | null;
-	onerror: ((err: Error) => void) | null;
+	onload: ((this: GlobalEventHandlers, ev: Event) => any) | null;
+	onerror: ((err: string|Event) => void) | null;
 }
```
dataMode needs to be made optional, as it does not exist in the browser, and the types for `onload` and `onerror` are incorrect.

## Steps to Reproduce

```js
import { createCanvas } from 'canvas';
const canvas = Canvas.createCanvas(200, 200);
const ctx = canvas.getContext('2d');
const image = new Image();
cxt.drawImage(image, 0, 0, canvas.width, canvas.height);
```
In the browser, 'new Image()' creates an HTMLImageElement, which cannot be passed to 'drawImage' because the Image type from node-canvas is incompatible.

## Your Environment

canvas@2.11.0
npm@9.4.2
6.1.8-gentoo-x86_64 #1 SMP PREEMPT_DYNAMIC Fri Jan 27 17:36:19 GMT 2023 x86_64 AMD Ryzen 9 5950X 16-Core Processor AuthenticAMD GNU/Linux

chearon:
The parts in the green would be wrong, as we always set `dataMode`, we do not pass an argument to `onload`,  and we never pass a string to `onerror`.

If it used to build, that was deceptive. `node-canvas` types are not going to match browser types exactly. You can import the types and have your functions work with either one, though, and that will be less error prone:

```ts
/// <reference lib="dom" />
import type {CanvasRenderingContext2d as NodeCanvasRenderingContext2D} from 'canvas';
function drawToContext(ctx: CanvasRenderingContext2d | NodeCanvasRenderingContext2D) {
   // ...
}
[...]

PR #2427, opened 2024-08-31
fix(DOMMatrix/DOMPoint): align to spec

ShaMan123:
DOMMatrix/DOMPoint were missing a few neglible methods as per the spec/d.ts files

This PR aligns to it.

Note that running `npm run tsd` throws: 
```bash

  index.test-d.ts:7:0
  ✖  7:0  Parameter type { new (init?: string | number[] | undefined): DOMMatrix; prototype: DOMMatrix; fromFloat32Array(array32: Float32Array): DOMMatrix; fromFloat64Array(array64: Float64Array): DOMMatrix; fromMatrix(other?: DOMMatrixInit | undefined): DOMMatrix; } is not identical to argument type typeof DOMMatrix.  
  ✖  8:0  Parameter type DOMMatrix is not identical to argument type DOMMatrix.                                                                                                                                                                                                                                                  
  ✖  9:0  Parameter type { new (x?: number | undefined, y?: number | undefined, z?: number | undefined, w?: number | undefined): DOMPoint; prototype: DOMPoint; fromPoint(other?: DOMPointInit | undefined): DOMPoint; } is not identical to argument type typeof DOMPoint.                                                      

  3 errors

```

However running the following succeeds:

```bash
npx tsc ./index.test-d.ts --noEmit 
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 ### Changed
 ### Added
 ### Fixed
+* Align DOMMatrix/DOMPoint to spec by adding missing methods
 
 
 3.0.0
--- a/index.d.ts
+++ b/index.d.ts
@@ -400,10 +400,13 @@ export class DOMPoint {
 	x: number;
 	y: number;
 	z: number;
+    matrixTransform(matrix?: DOMMatrixInit): DOMPoint;
+    toJSON(): any;
+	static fromPoint(other?: DOMPointInit): DOMPoint
 }
 
 export class DOMMatrix {
-	constructor(init: string | number[]);
+	constructor(init?: string | number[]);
 	toString(): string;
 	multiply(other?: DOMMatrix): DOMMatrix;
 	multiplySelf(other?: DOMMatrix): DOMMatrix;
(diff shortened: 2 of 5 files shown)


ShaMan123 (commented):
Also I would like to suggest adding the following to the index.js file:
```
Object.assign(global, { DOMPoint, DOMMatrix })

```

This way projects using node-canvas and browser/offscreen native canvas with DOMPoint, DOMMatrix can be integrated seamlessly

chearon:
Changes look good. Going to force-push to fix the formatting and move the changelog entry to 3.0.

> Also I would like to suggest adding the following to the index.js file:

That's something that should probably be left up to users though.

chearon:
>> However running the following succeeds:
>
> `npx tsc ./index.test-d.ts --noEmit`

I'm not sure I understand. The `expectType` test _should_ fail, shouldn't it? We're probably never going to have exact types (or even assignable types: see #2197). I'm going to remove the tests in this PR since our `npm run tsd` script should be passing, but if you can get `expectAssignable` to work, I'd merge that.

ShaMan123:
Happy you merged it... don't remember anything about this... so no idea about types. Need to look into it, but since you merged I am guessing it is good enough.

PR #2447, opened 2024-10-27
docs: updated readme with command for installing packages from source using Yarn

sahanar11:
Added a command for installing **packages from source** using Yarn.

For npm, use: npm install --build-from-source=true
For yarn use: npm_config_build_from_source=true yarn install

Ref: Please find more details in the Yarn documentation here - https://classic.yarnpkg.com/en/docs/envvars/#toc-npm-config


chearon:
Thanks but this is documented in a different way now

PR #2450, opened 2024-11-01
Remove object files and other large not needed artifacts from prebuilds.

btsimonh:
Thanks for contributing!

- [ ] Have you updated CHANGELOG.md? - no there is none for prebuilds?

Removes LARGE artifacts in the prebuilds.

## Macos:
deletes object files (save a few mb)
## Linux:
deletes object files and strips libraries (save ~22mbytes)
## Windows
Deletes object files and debug files. (save 45mbytes)

### DIFF
--- a/.github/workflows/prebuild.yaml
+++ b/.github/workflows/prebuild.yaml
@@ -261,6 +261,8 @@ jobs:
       - name: Bundle pt 2
         shell: msys2 {0}
         run: |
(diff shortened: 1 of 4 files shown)


chearon (approved):
Thank you, I've been meaning to do this. I have a few other ideas on how to get the size down even more and hopefully fit it in the NPM tarball.

PR #2453, opened 2024-11-01
Master - add test prebuilds

btsimonh:
Thanks for contributing!

- [ ] Have you updated CHANGELOG.md? - no it's not a version update

see

https://github.com/Automattic/node-canvas/pull/2452

for details.  we need the file in master else we won't see the action :(.

### DIFF
--- a/.github/workflows/testprebuild.yaml
+++ b/.github/workflows/testprebuild.yaml
@@ -19,11 +19,16 @@ on:
         description: 'Target node-canvas version'
         required: true
         type: choice
+        # add releases here.  e.g.
+        #options:
+        #  - '["vdev"]'
+        #  - '["v3.0.0-rc2"]'
+        #  - '["vdev", "v3.0.0-rc2"]'
+        #default: '["vdev", "v3.0.0-rc2"]'
+        
         options:
-          - '["vdev"]'
           - '["v3.0.0-rc2"]'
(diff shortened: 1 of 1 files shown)


.github/workflows/testprebuild.yaml
```diff
@@ -0,0 +1,204 @@
...
+      node:
+        description: 'Target node version'
+        required: true
+        type: choice
+        options:
+          - '["18"]'
+          - '["20"]'
+          - '["21"]'
+          - '["22"]'
+          - '["18", "20", "22"]'
```
cclauss: 23, 24, 25?

.github/workflows/testprebuild.yaml
```diff
@@ -0,0 +1,204 @@
...
+      CANVAS_VERSION_TO_BUILD: ${{ matrix.canvas_tag }}
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          ref: ${{ matrix.canvas_tag }}
+          # Fetch all commits/all branches so we can checkout the prebuild
+          # branch's files
+          fetch-depth: 0
+
+      - uses: actions/setup-node@v1
```
cclauss:
```suggestion
      - uses: actions/setup-node@v6
```

chearon:
There is only one build now, making this unnecessary, but thanks for the effort anyways.

PR #2465, opened 2024-12-06
Modernize node version matrix

chearon:
These are the currently supported versions. Luckily the Windows file path issue was backported to node 20.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node: [18.12.0, 20.9.0]
+        node: [18.20.5, 20.18.1, 22.12.0, 23.3.0]
     steps:
       - uses: actions/setup-node@v4
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node: [18.12.0, 20.9.0]
+        node: [18.20.5, 20.18.1, 22.12.0, 23.3.0]
```
cclauss, line 16: Awesome work!! It would be better to have a 
canary in the coal mine to spot future breakage.
```suggestion
        node: [18.20.5, 20.18.1, 22.12.0, latest]
```

cclauss (approved):

.github/workflows/ci.yaml
chearon: Hmm only when we push though. I wish I'd pinned more versions in the prebuild scripts because it contributed to random failures.

I think new Node versions are something we have to explicitly do work for rather than something that should start potentially breaking builds. @LinusU any opinions?

Sorry about your other PR, I looked for existing ones but somehow still missed it.

cclauss: My sense with `latest` is that there is a chance that CI tells us we have a problem before end users complain and that the non-latest jobs assure us an only the bleed-edge end users are seeing those problems.

cclauss:
Please finalize the testing and create a new release.

On the `nodejs/node-gyp` repo there are [six pages of issues that mention `node-canvas`](https://github.com/search?q=repo%3Anodejs%2Fnode-gyp+node-canvas&type=issues) and there has not been a new release in two years.
* https://www.npmjs.com/package/canvas
* https://github.com/Automattic/node-canvas/releases

Issue #2392, opened 2024-07-03
Request for arm64 Linux Prebuild Package


narukeu:
## Issue or Feature
I am creating a Docker container for a Node.js project on an arm64 computer, which includes node-canvas. I believe there are many users with similar needs, so I hope you can provide an arm64 Linux prebuild package.

## Your Environment
Debian GNU/Linux arm64

giacomo-bananadevs:
Does anyone managed to build the package from source? I can't find a way to build it on a windows ARM64 that runs WSL2. 
I need to install the package on the ubuntu subsystem, running: 
```
gyp info using **node-gyp@10.2.0**
gyp info using **node@20.10.0 | linux | arm64**
gyp info find Python using **Python version 3.10.12** found at "/usr/bin/python3"
```
Even trying to hand-compiling the package results always in the same error: 
```
gyp info spawn args [ 'BUILDTYPE=Release', '-C', 'build' ]
node:internal/process/promises:289
            triggerUncaughtException(err, true /* fromPromise */);
            ^

[UnhandledPromiseRejection: This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason "undefined".] {
  code: 'ERR_UNHANDLED_REJECTION'
}

Node.js v20.10.0
gyp ERR! build error
gyp ERR! stack Error: `make` failed with exit code: 1
gyp ERR! stack at ChildProcess.<anonymous> ([path_top_home_dir]/.nvm/versions/node/v20.10.0/lib/node_modules/node-gyp/lib/build.js:216:23)
gyp ERR! **System Linux 5.15.153.1-microsoft-standard-WSL2**
gyp ERR! command "[path_top_home_dir]/.nvm/versions/node/v20.10.0/bin/node" "[path_top_home_dir]/.nvm/versions/node/v20.10.0/bin/node-gyp" "build"
gyp ERR! cwd [path_to_prj]/node_modules/canvas/build
gyp ERR! node -v v20.10.0
gyp ERR! node-gyp -v v10.2.0
gyp ERR! not ok
```

I there any workaround to installa the package on Ubuntu ARM64? Or am I missing something?

cjutstraMaxedy:
This would also be very helpful for us. We're running automated tests on ARM machines.

kadishmal:
From `.buildspec` for AWS CodeBuild:
```
version: 0.2
phases:
  install:
    commands:
[...]

PR #2494, opened 2025-03-11
feat: #2392: add arm based ubuntu runner

dsdeboer:
Thanks for contributing!

Due to the public availability of the new ubuntu-arm runner on GithubActions I think this is the appropriate time to also include the prebuild for this architecture.

### DIFF
--- a/.github/workflows/prebuild.yaml
+++ b/.github/workflows/prebuild.yaml
@@ -51,9 +51,10 @@ jobs:
       fail-fast: false
       matrix:
         node: [21]
+        os: ["ubuntu-latest", "ubuntu-latest-arm"]
         canvas_tag: ["v3.1.0"] # e.g. "v2.6.1"
     name: ${{ matrix.canvas_tag }}, Node.js ${{ matrix.node }}, Linux
-    runs-on: ubuntu-latest
+    runs-on: matrix.os
     container:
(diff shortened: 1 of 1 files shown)


mistic:
@chearon do you think we can move with this PR forward? Having this in will help stop building locally the native module on that arch type 😃

chearon:
I wish it was this easy, but all of the dependencies have to be built for ARM as well. This isn't going to work.

My long term plan is to use Zig's build system so that we can cross-compile easily, but that will require trimming dependencies first.

mistic:
@chearon I believe I got a working image with the changes I'm proposing in the Dockerfile at https://github.com/Automattic/node-canvas/pull/2516.

On top of that the image in use for the arm64 build context will have to be changed in this PR at https://github.com/Automattic/node-canvas/pull/2494/files#diff-c18dd11250e4e0f0d19035ce56dd87584b7d4f9803ff95437d4831a6d92e938aR59 .

Be free to produce your own images and publish those based out of the Dockerfile changes PR.

It's not optimal neither a long term solution as the one you have in mind by using Zig all around. But it should be enough to unblock this for now in the same build terms used for all the other current artifacts. 😃

chearon:
Thanks for kicking this off @dsdeboer and @mistic. I'll make the necessary changes to this PR separately. Hopefully we get an arm64 release today!

mistic:
Thanks for the cooperation @chearon , really much appreciated 🙏

Issue #2171, opened 2022-12-06
fillText with maxWidth 0 breaks subsequent fillText calls


JacobNorlin:
## Issue
Any call to fillText after a fillText with maxWidth=0 does not render anything.

## Steps to Reproduce

```js
const {createCanvas} = require('canvas');
const fs = require('fs');

const canvas = createCanvas(100, 100);
const context = canvas.getContext('2d');
context.fillStyle = 'white';
context.fillRect(0, 0, 100, 100);

context.font = '12px Arial';
context.fillStyle = 'black';
context.fillText('Test1', 0, 40);
context.fillText('Test2', 0, 60, 0);
context.fillText('Test3', 0, 80);

const buf = canvas.toBuffer();

fs.writeFileSync('test.png', buf);
```

I would expect 'Test3' to be rendered in this case, but there is nothing rendered.

## Your Environment
* Node canvas 2.10.1
* Node 16.13.2, Debian 10 (WSL)

PR #2496, opened 2025-03-17
Fixes related to fillText that caused subsequent calls of fillText to not appear

truman126:
Issue (#2171) fixed. Calling fillText with a maxWidth <= 0 caused subsequent calls of fillText to not appear. Now only the call with a maxWidth <= 0 does not appear.


chearon (approved):
Thanks! The divide-by zero -> NaN must put Cairo in an error state.

Issue #2448, opened 2024-10-31
Build Error: Incompatibility of `canvas@2.11.2` with Node.js v22.11.0


bahram-della-ai:
**Description:**

Hi there,

I'm experiencing build failures when trying to install `canvas@2.11.2` using Node.js v22.11.0. It appears there might be compatibility issues between this version of Node.js and the `canvas` module.

**Environment:**

- **Node.js Version:** v22.11.0
- **Canvas Version:** 2.11.2
- **Platform:** Linux (Docker base image `node:lts-slim` as of October 2024)
- **Dependency Manager:** npm

**Steps to Reproduce:**

1. **Dockerfile Setup:**

   ```dockerfile
   FROM node:lts-slim AS build
   RUN apt-get update && apt-get install -y --no-install-recommends \
       make \
       g++ \
       dumb-init \
       build-essential \
       libcairo2-dev \
       libpango1.0-dev \
       libjpeg-dev \
       libgif-dev \
       librsvg2-dev \
       python3 \
       && rm -rf /var/lib/apt/lists/*

   WORKDIR /usr
   COPY package*.json tsconfig.json ./
   RUN npm ci
   ```

2. **Install Dependencies:**

   - Run `npm ci` to install all dependencies, including `canvas@2.11.2`.

3. **Observe the Build Failure:**

   - The installation fails during the build process of `canvas`.

**Error Log Highlights:**

```
npm ERR! code 1
npm ERR! path /usr/node_modules/canvas
npm ERR! command failed
npm ERR! command sh -c node-pre-gyp install --fallback-to-build --update-binary
npm ERR! make: Entering directory '/usr/node_modules/canvas/build'
npm ERR!   SOLINK_MODULE(target) Release/obj.target/canvas-postbuild.node
npm ERR!   COPY Release/canvas-postbuild.node
npm ERR!   CXX(target) Release/obj.target/canvas/src/backend/Backend.o
npm ERR! In file included from ../src/backend/Backend.h:6,
npm ERR!                  from ../src/backend/Backend.cc:1:
npm ERR! ../../nan/nan.h: In function 'void Nan::SetAccessor(...)':
npm ERR! ../../nan/nan.h:2546:19: error: no matching function for call to 'v8::ObjectTemplate::SetAccessor(...)'
npm ERR! ...
npm ERR! gyp ERR! build error
npm ERR! gyp ERR! stack Error: `make` failed with exit code: 2
npm ERR! gyp ERR! node -v v22.11.0
[...]

PR #2497, opened 2025-03-29
fix(deps): update `prebuild-install` to work on Node 22.14+

agilgur5:
### Motivation

This installation error on Node 22.14+:
```
npm error prebuild-install warn This package does not support N-API version undefined
npm error prebuild-install warn install No prebuilt binaries found (target=undefined runtime=napi arch=x64 libc= platform=linux)
```

is resolved by updating to newer `prebuild-install` 7.1.3, which has a newer version of `napi-build-utils` (2.0.0) with a bugfix for newer N-API version comparison

### Further Reading

This fixes the issue I noted in https://github.com/Automattic/node-canvas/issues/2448#issuecomment-2759459548 with the suggestion in the following comment https://github.com/Automattic/node-canvas/issues/2448#issuecomment-2759866317. See also https://github.com/prebuild/prebuild-install/issues/203,  https://github.com/inspiredware/napi-build-utils/issues/6 for root cause.

### Notes to Reviewers

While consumers can update their version of `prebuild-install` in their lockfiles independently, the version bump here will _require_ a newer version that is backward- and forward-compatible.

### Verification

This dep bump fixed the error for me in https://github.com/agilgur5/react-signature-canvas/pull/132

### template

Thanks for contributing!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 * Fix a crash when SVGs without width or height are loaded (#2486)
+* Fix fetching prebuilds during installation on certain newer versions of Node (#2497)
 
 3.1.0
 ==================
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
   ],
   "dependencies": {
     "node-addon-api": "^7.0.0",
-    "prebuild-install": "^7.1.1"
+    "prebuild-install": "^7.1.3"
   },
   "devDependencies": {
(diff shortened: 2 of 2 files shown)


chearon (approved):
Thanks, I hadn't realized that the version after `^` is a minimum (i.e. I thought `7.1.1` satisfies `^7.1.3`). But a quick Google search shows that this should work.

agilgur5:
Indeed it is a minimum; for reference, here's the [official NPM `semver` docs on "caret ranges"](https://github.com/npm/node-semver/tree/v7.7.1?tab=readme-ov-file#caret-ranges-123-025-004)

Issue #2498, opened 2025-03-31
Scaled SVGs render incorrectly


mgilfillan:
When drawing an SVG after setting the width and height to a larger value the drawn SVG is scaled incorrectly.

I've put together a minimal reproduction:

**index.js**
```javascript
import {createCanvas, loadImage} from "canvas";
import fs from "fs";

function writeImage(image, size, fileName) {
    const canvas = createCanvas(size, size);
    const ctx = canvas.getContext('2d');

    image.width = size;
    image.height = size;

    ctx.drawImage(image, 0, 0, size, size);

    fs.writeFileSync(fileName, canvas.toBuffer());
}

writeImage(await loadImage('./image.svg'), 100, './no scaling.png');
writeImage(await loadImage('./image.svg'), 200, './2x scaling.png');
```

**image.svg**
```svg
<svg version="1.2" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 100 100" width="100" height="100">
    <circle fill="#ff0000" cx="50" cy="50" r="50"/>
</svg>
```

I would expect both output images to be visually the same, other than being scaled up, however the scaled one renders the SVG much larger than it should. This works correctly in node-canvas 2.x

| node-canvas version | No Scaling | 2x scaling |
|--------|--------|--------|
| 2.11.2| ![Image](https://github.com/user-attachments/assets/52fbe986-9527-4a4b-ab1d-b611beb47477) | ![Image](https://github.com/user-attachments/assets/3bc4fe8f-09d0-4f59-82d6-4cff101b5cc4) |
| 3.1.0| ![Image](https://github.com/user-attachments/assets/cb9fd860-c339-4cf4-975b-65189801a409) | ![Image](https://github.com/user-attachments/assets/84aff49a-10e1-419d-830a-6471bba27a17) |

PR #2500, opened 2025-04-03
Fix SVG scaling (#2498)

mgilfillan:
Thanks for contributing!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 * Fix a crash when SVGs without width or height are loaded (#2486)
 * Fix fetching prebuilds during installation on certain newer versions of Node (#2497)
 * Fixed issue with fillText that was breaking subsequent fillText calls (#2171)
+* Fix svg rendering when the image is resized (#2498)
 
 3.1.0
 ==================
--- a/src/Image.cc
+++ b/src/Image.cc
@@ -1506,9 +1506,6 @@ Image::renderSVGToSurface() {
   }
 
   cairo_t *cr = cairo_create(_surface);
-  cairo_scale(cr,
-    (double)width / (double)naturalWidth,
-    (double)height / (double)naturalHeight);
   status = cairo_status(cr);
   if (status != CAIRO_STATUS_SUCCESS) {
     g_object_unref(_rsvg);
--- a/test/public/tests.js
+++ b/test/public/tests.js
@@ -3,6 +3,8 @@ let Image
 let imageSrc
 const tests = {}
 
+/* global btoa */
+
 if (typeof module !== 'undefined' && module.exports) {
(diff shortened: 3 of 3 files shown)


mgilfillan:
Fixes #2498

chearon:
Thanks! This doesn't explain why it worked in 2.11.2 though. I'll dig into this soon.

chearon (changes requested):
Thanks for looking into this. I did not bisect, but it looks like yet another issue caused by f3184ba9da2737dbe25275631678a7ef5924fe6b. That changed from using [`rsvg_handle_render_cairo`](https://gnome.pages.gitlab.gnome.org/librsvg/Rsvg-2.0/method.Handle.render_cairo.html) to [`rsvg_handle_render_document`](https://gnome.pages.gitlab.gnome.org/librsvg/Rsvg-2.0/method.Handle.render_document.html).

I think the proper way to use that API actually _is_ to pass the `img.width` and `img.height`. The scaling we did for the old API is the problem. Can you see if the change below will work for you instead?

```diff
diff --git a/src/Image.cc b/src/Image.cc
index ce0b519..9737365 100644
--- a/src/Image.cc
+++ b/src/Image.cc
@@ -1506,9 +1506,6 @@ Image::renderSVGToSurface() {
   }
 
   cairo_t *cr = cairo_create(_surface);
-  cairo_scale(cr,
-    (double)width / (double)naturalWidth,
-    (double)height / (double)naturalHeight);
   status = cairo_status(cr);
   if (status != CAIRO_STATUS_SUCCESS) {
     g_object_unref(_rsvg);
```

mgilfillan:
I've tested the proposed change and it doesn't look like it works correctly. The image content remains the original size, rather than bing scaled to the new dimensions:


| No Scaling | 2x scaling | 0.5x scaling |
|--------|--------|--------|
| no scaling | 2x scaling | 0 5x scaling

chearon:
Are you sure you rebuilt correctly? These are the results I get:

| 0.5x | 2x | no scaling |
| -- | -- | -- |
| 0 5x scaling | 2x scaling | no scaling |

<details>
<summary>examples/bug.js</summary>

```js
const {createCanvas, loadImage} = require("..");
const fs = require("node:fs");

function writeImage(image, size, fileName) {
    const canvas = createCanvas(size, size);

    image.width = size;
    image.height = size;

    ctx.drawImage(image, 0, 0, size, size);

    fs.writeFileSync(fileName, canvas.toBuffer());
}

(async function () {
  writeImage(await loadImage('./image.svg'), 100, './no scaling.png');
  writeImage(await loadImage('./image.svg'), 200, './2x scaling.png');
  writeImage(await loadImage('./image.svg'), 50, './0.5x scaling.png');
})();
```

</details>

chearon:
I just amended it myself since it's a simple change, and added a test. Thanks!

PR #2510, opened 2025-04-30
fix(TextMetrics): rtl direction + start/end textAlign

### LINKED ISSUE #2508: [BUG]: `measureText` not respecting `ctx.direction`
I have encountered this bug when using `ctx.direction = 'rtl'` and `ctx.textAlign = 'start' | 'end'`

It is clear that the code fails to first resolve `start`/`end` `textAlign` and therefore executes as if direction is `ltr`.

https://github.com/Automattic/node-canvas/blob/1234a86f65b2318906334ce0790cd12401c67a25/src/CanvasRenderingContext2d.cc#L2819-L2832

I am willing to PR.

ShaMan123:
Thanks for contributing!

closes #2508

### DIFF
--- a/src/CanvasRenderingContext2d.cc
+++ b/src/CanvasRenderingContext2d.cc
@@ -2557,6 +2557,20 @@ inline double getBaselineAdjustment(PangoLayout* layout, short baseline) {
   }
 }
 
+text_align_t
+Context2d::resolveTextAlignment() {
+  text_align_t alignment = state->textAlignment;
+
+  // Convert start/end to left/right based on direction
+  if (alignment == TEXT_ALIGNMENT_START) {
+    return (state->direction == "rtl") ? TEXT_ALIGNMENT_RIGHT : TEXT_ALIGNMENT_LEFT;
+  } else if (alignment == TEXT_ALIGNMENT_END) {
+    return (state->direction == "rtl") ? TEXT_ALIGNMENT_LEFT : TEXT_ALIGNMENT_RIGHT;
+  }
+
+  return alignment;
+}
+
 /*
  * Set text path for the string in the layout at (x, y).
  * This function is called by paintText and won't behave correctly
@@ -2567,14 +2581,7 @@ inline double getBaselineAdjustment(PangoLayout* layout, short baseline) {
 void
 Context2d::setTextPath(double x, double y) {
   PangoRectangle logical_rect;
-  text_align_t alignment = state->textAlignment;
-
-  // Convert start/end to left/right based on direction
-  if (alignment == TEXT_ALIGNMENT_START) {
-    alignment = (state->direction == "rtl") ? TEXT_ALIGNMENT_RIGHT : TEXT_ALIGNMENT_LEFT;
-  } else if (alignment == TEXT_ALIGNMENT_END) {
-    alignment = (state->direction == "rtl") ? TEXT_ALIGNMENT_LEFT : TEXT_ALIGNMENT_RIGHT;
-  }
+  text_align_t alignment = resolveTextAlignment();
 
   switch (alignment) {
     case TEXT_ALIGNMENT_CENTER:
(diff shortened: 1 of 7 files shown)


ShaMan123 (commented):
Is there any other test/thing I should add/do?

test/wpt/generated/the-canvas-element.js
```diff
@@ -171,12 +171,6 @@ describe("WPT: the-canvas-element", function () {
 		assert.strictEqual(window.CanvasRenderingContext2D.prototype.fill, undefined, "window.CanvasRenderingContext2D.prototype.fill", "undefined")
 	});
 
-	it("2d.type class string", function () {
-		const canvas = createCanvas(100, 50);
-		const ctx = canvas.getContext("2d");
-		assert.strictEqual(Object.prototype.toString.call(ctx), '[object CanvasRenderingContext2D]')
```
chearon, line 177: Did this need to be removed?

src/CanvasRenderingContext2d.cc
```diff
@@ -2816,16 +2816,23 @@ Context2d::MeasureText(const Napi::CallbackInfo& info) {
...
   metrics = PANGO_LAYOUT_GET_METRICS(layout);
 
+  text_align_t alignment = state->textAlignment;
+
+  // Convert start/end to left/right based on direction
+  if (alignment == TEXT_ALIGNMENT_START) {
+    alignment = (state->direction == "rtl") ? TEXT_ALIGNMENT_RIGHT : TEXT_ALIGNMENT_LEFT;
+  } else if (alignment == TEXT_ALIGNMENT_END) {
+    alignment = (state->direction == "rtl") ? TEXT_ALIGNMENT_LEFT : TEXT_ALIGNMENT_RIGHT;
+  }
```
chearon, line 2826: Looks like this would be pretty easy to split out into a function since we do the exact same thing in `setTextPath`. What do you think?

test/wpt/text-styles.yaml
```diff
@@ -503,13 +503,34 @@
...
+    ltrEndMetrics = ctx.measureText('hello');
+    assert.deepStrictEqual(ltrEndMetrics, rightMetrics, "ltr end metrics should equal right metrics");
+    	
+    ctx.direction = "rtl";
+    rtlEndMetrics = ctx.measureText('hello');
+    assert.deepStrictEqual(rtlEndMetrics, leftMetrics, "rtl end metrics should equal left metrics");
 
-    ctx.textAlign = "left"
-    metrics = ctx.measureText('hello');
-    @assert metrics.actualBoundingBoxLeft < metrics.actualBoundingBoxRight;
```
chearon, line 512: I don't think this goes here. This test is imported from the [WPT tests](https://github.com/web-platform-tests/wpt/blob/ccaadde016187b16e607c6c6ae65e37089c452c5/html/canvas/tools/yaml/text.yaml#L522-L531), so it would get overwritten if they did any updates. You should be able to write this in `canvas.test.js` though.

chearon (changes requested):
Thanks and I'm glad to see new tests! Need a couple of changes I think.

src/CanvasRenderingContext2d.cc
ShaMan123: Will do, not well versed in c++ but will manage

test/wpt/text-styles.yaml
ShaMan123: I thought the yaml files generate the js test files. Regardless I saw nothing runs these tests on CI

src/CanvasRenderingContext2d.cc
```diff
@@ -2557,6 +2557,20 @@ inline double getBaselineAdjustment(PangoLayout* layout, short baseline) {
   }
 }
 
+text_align_t
```
ShaMan123, line 2560: Should I expose a strict text align enum?

```
enum text_align_strict_t : int8_t {
  STRICT_TEXT_ALIGNMENT_LEFT = TEXT_ALIGNMENT_LEFT,
  STRICT_TEXT_ALIGNMENT_CENTER = TEXT_ALIGNMENT_CENTER,
  STRICT_TEXT_ALIGNMENT_RIGHT = TEXT_ALIGNMENT_RIGHT
};

```

ShaMan123 (commented):
I reverted the pwt changes in favor of the jest test and exposed `resolveTextAlignment` as requested

src/CanvasRenderingContext2d.cc
ShaMan123: done

ShaMan123:
@chearon ?

chearon (approved):
Looks great!

PR #2516, opened 2025-06-16
Fix prebuilds docker image for arm64 build

mistic:
This PR is in the context of what was mentioned at https://github.com/Automattic/node-canvas/pull/2494#issuecomment-2977120532

With those changes I was able to successfully build both amd64 and arm64 images out of that Dockerfile. They can be found [here](https://hub.docker.com/r/tiagocostaelastic/canvas-prebuilt/tags).

Together with the changes at https://github.com/Automattic/node-canvas/pull/2494 plus an adjustment to use a correct image to build on in the context of arm64, it should be possible to release prebuilds support on that arch.

\cc @chearon


chearon:
Nice! I'm wrapping up a few more things but I should be doing another release soon. I'll give it a shot then.

mistic:
Thank you @chearon, much appreciated! 🙏

chearon (approved):

PR #2526, opened 2025-08-02
Set pango language through ctx.lang

mapmeld:
This would set pango's language through the Canvas context ( `ctx.lang`, docs: https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/lang )

My goal was to fix issues where the same Unicode codepoint should appear differently in Simplified Chinese, Traditional Chinese, and Japanese (as an example: lang=zh <span lang="zh">门</span> vs. lang=ja <span lang="ja">门</span>). This can be done with language-specific fonts, but modern fonts often support all options and use the lang tag.
Han unification: https://en.wikipedia.org/wiki/Han_unification
Related issue for Chrome client-side canvas: https://github.com/mapbox/tiny-sdf/pull/58

I'm submitting the code and test as a draft PR because **this is still failing** to change the output.
I did rebuild bindings, and tried with and without `pango_cairo_update_layout`.
Other ideas: this is failing to set the language, setting it at the wrong time, the fonts or something needs to be reset, or pango handles Han unification using something else? I'm a longtime node-canvas user but new to the internals.
I found this reference to `PANGO_LANGUAGE` and `%PANGO_SCRIPT_HAN` https://github.com/gtk-rs/gir-files/blob/main/Pango-1.0.gir#L11631

### DIFF
--- a/src/CanvasRenderingContext2d.cc
+++ b/src/CanvasRenderingContext2d.cc
@@ -792,7 +792,12 @@ Context2d::SetDirection(const Napi::CallbackInfo& info, const Napi::Value& value
  */
 Napi::Value
 Context2d::GetLanguage(const Napi::CallbackInfo& info) {
-  return Napi::String::New(env, state->lang);
+  PangoLanguage* lang = pango_context_get_language(pango_layout_get_context(_layout));
+  if (!lang) {
+    return Napi::String::New(env, "");
+  }
+  std::string langCode = pango_language_to_string(lang);
+  return Napi::String::New(env, langCode);
 }
 
 /*
@@ -803,10 +808,7 @@ Context2d::SetLanguage(const Napi::CallbackInfo& info, const Napi::Value& value)
   if (!value.IsString()) return;
 
   std::string lang = value.As<Napi::String>();
-  state->lang = lang;
-
   pango_context_set_language(pango_layout_get_context(_layout), pango_language_from_string(lang.c_str()));
-  pango_cairo_update_layout(context(), _layout);
 }
 
 /*
--- a/src/CanvasRenderingContext2d.cc
+++ b/src/CanvasRenderingContext2d.cc
@@ -806,7 +806,6 @@ Context2d::SetLanguage(const Napi::CallbackInfo& info, const Napi::Value& value)
   state->lang = lang;
 
   pango_context_set_language(pango_layout_get_context(_layout), pango_language_from_string(lang.c_str()));
-  pango_cairo_update_layout(context(), _layout);
 }
 
 /*
--- a/src/CanvasRenderingContext2d.cc
+++ b/src/CanvasRenderingContext2d.cc
@@ -792,12 +792,7 @@ Context2d::SetDirection(const Napi::CallbackInfo& info, const Napi::Value& value
  */
 Napi::Value
 Context2d::GetLanguage(const Napi::CallbackInfo& info) {
-  PangoLanguage* lang = pango_context_get_language(pango_layout_get_context(_layout));
-  if (!lang) {
-    return Napi::String::New(env, "");
-  }
-  std::string langCode = pango_language_to_string(lang);
-  return Napi::String::New(env, langCode);
+  return Napi::String::New(env, state->lang);
 }
 
 /*
@@ -808,7 +803,10 @@ Context2d::SetLanguage(const Napi::CallbackInfo& info, const Napi::Value& value)
   if (!value.IsString()) return;
 
   std::string lang = value.As<Napi::String>();
+  state->lang = lang;
+
   pango_context_set_language(pango_layout_get_context(_layout), pango_language_from_string(lang.c_str()));
+  pango_cairo_update_layout(context(), _layout);
 }
 
 /*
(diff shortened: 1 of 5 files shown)


src/CanvasRenderingContext2d.cc
```diff
@@ -786,6 +787,28 @@ Context2d::SetDirection(const Napi::CallbackInfo& info, const Napi::Value& value
   state->direction = dir;
 }
 
+/*
+ * Get language.
+ */
+Napi::Value
+Context2d::GetLanguage(const Napi::CallbackInfo& info) {
+  return Napi::String::New(env, state->lang);
```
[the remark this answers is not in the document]
mapmeld: makes sense, done ✅

chearon:
@mapmeld this does seem to work (using [Source Han Sans](https://github.com/adobe-fonts/source-han-sans/tree/release?tab=readme-ov-file), Super OTC):

font


<details>
<summary>source (change examples/font.js with this)</summary>

```js
const fs = require('fs')
const path = require('path')
const Canvas = require('..')

const canvas = Canvas.createCanvas(320, 320)

ctx.lang = 'zh'
ctx.font = '50px Source Hans Sans'
ctx.fillText('\u8FD4', 0, 70, 100)

ctx.lang = 'zh-HK'
ctx.font = '50px Source Han Sans'
ctx.fillText('\u8FD4', 0, 140)

ctx.lang = 'tc'
ctx.font = '50px Source Han Sans'
ctx.fillText('\u8FD4', 0, 210)

canvas.createPNGStream().pipe(fs.createWriteStream(path.join(__dirname, 'font.png')))
```

</details>

My impression has been that usually different fonts are shipped for different languages, but this should definitely be used more.

src/CanvasRenderingContext2d.cc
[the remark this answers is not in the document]
1ec5: In that case, should `PANGO_LAYOUT_GET_METRICS` use the `lang` field instead of `pango_context_get_language()`?

src/CanvasRenderingContext2d.cc
```diff
@@ -786,6 +787,28 @@ Context2d::SetDirection(const Napi::CallbackInfo& info, const Napi::Value& value
...
+ */
+void
+Context2d::SetLanguage(const Napi::CallbackInfo& info, const Napi::Value& value) {
+  if (!value.IsString()) return;
+
+  std::string lang = value.As<Napi::String>();
+  state->lang = lang;
+
+  pango_context_set_language(pango_layout_get_context(_layout), pango_language_from_string(lang.c_str()));
+  pango_cairo_update_layout(context(), _layout);
```
chearon, line 809: This already happens lazily in the text paths: `MeasureText` and `paintText`


```suggestion
```

src/CanvasRenderingContext2d.cc
```diff
@@ -786,6 +787,28 @@ Context2d::SetDirection(const Napi::CallbackInfo& info, const Napi::Value& value
...
+ * Set language.
+ */
+void
+Context2d::SetLanguage(const Napi::CallbackInfo& info, const Napi::Value& value) {
+  if (!value.IsString()) return;
+
+  std::string lang = value.As<Napi::String>();
+  state->lang = lang;
+
+  pango_context_set_language(pango_layout_get_context(_layout), pango_language_from_string(lang.c_str()));
```
chearon, line 808: This should be done lazily too. Can you put it in `MeasureText` and `paintText`, right before `pango_cairo_update_layout`?

test/canvas.test.js
```diff
@@ -486,6 +486,24 @@ describe('Canvas', function () {
...
+    ctx1.fillText('门', 10, 10)
+
+    const canvas2 = createCanvas(200, 200)
+    const ctx2 = canvas2.getContext('2d')
+    ctx2.font = '20px sans-serif';
+    ctx2.lang = 'zh'
+    ctx2.fillText('门', 10, 10)
+
+    assert.notDeepEqual(ctx1.getImageData(10, 10, 40, 40), ctx2.getImageData(10, 10, 40, 40))
+  })
```
chearon, line 505: This test won't work on an OS that doesn't come preinstalled with fonts that support the different languages. I'd be happy with just an update to the `describe('Context2d#save()/restore()'` test. But we could also add a font to the repo if it's small enough.

src/CanvasRenderingContext2d.cc
chearon, line 1: Can you also update the `CHANGELOG.md`?

chearon (commented):
Need to also update the `CHANGELOG.md`,

test/canvas.test.js
mapmeld: @chearon oh cool, it is working with Source Han Sans! Thanks for finding that. As long as it's documented / searchable here I'm OK with moving forward with the simpler test

src/CanvasRenderingContext2d.cc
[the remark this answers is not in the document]
mapmeld: maybe the better way to get the language before it is lazily committed?
```
if (state->lang == "") { // unchanged
  pango_context_get_language()
} else {
  pango_language_from_string(state->lang)
}
```

chearon:
Thanks!

Issue #1391, opened 2019-03-22
Support for color values with percentage representation ?


kyasu1:
## Issue or Feature

The current implementation looks like ignoring color value with percentage representations `rgba(100%,100%,100%,1.0)`. It is valid css color value according to [this](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value) document. Is there any plan for supporting this or am I overlooking something ?

A library I am using outputs with this format and I am stuck here :S

## Your Environment
* Version of node-canvas (output of `npm list canvas` or `yarn list canvas`): canvas@2.4.1
* Environment (e.g. node 4.2.0 on Mac OS X 10.8): Mac OS X 10.4.3

PR #2540, opened 2025-10-29
Add support for % in rgb and fix parsing for % in alpha.

pranav1344:

The parsing for % values in alpha wasn't working properly for single digit integers and for all decimal percentages starting with 0. Also added support for % parsing in rgb and rgba. This was mentioned in issue #1391 too.

### DIFF
--- a/test/canvas.test.js
+++ b/test/canvas.test.js
@@ -232,6 +232,27 @@ describe('Canvas', function () {
 
     ctx.fillStyle = 'rgb( 255, 300.09, 90, 40%)'
     assert.equal('rgba(255, 255, 90, 0.40)', ctx.fillStyle)
+
+    ctx.fillStyle = 'rgb( 255, 100%, 50%, 60%)'
+    assert.equal('rgba(255, 255, 128, 0.60)', ctx.fillStyle)
+
+    ctx.fillStyle = 'rgb( 255, 20%, 90, 70%)'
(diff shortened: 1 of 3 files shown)


test/canvas.test.js
```diff
@@ -232,6 +232,27 @@ describe('Canvas', function () {
...
+    ctx.fillStyle = 'rgb( 100%, 300%, 90, 50%)'
+    assert.equal('rgba(255, 255, 90, 0.50)', ctx.fillStyle)
+
+    ctx.fillStyle = 'rgb( 100%, 300.09, 90%, 0)'
+    assert.equal('rgba(255, 255, 230, 0.00)', ctx.fillStyle)
+
+    ctx.fillStyle = 'rgb( 100%, 99%, 80%     50%)'
+    assert.equal('rgba(255, 253, 204, 0.50)', ctx.fillStyle)
+
+    ctx.fillStyle = 'rgb( 100%, 99.99%, 40% / 50%)'
```
yisibl: To my knowledge, this syntax is not supported in CSS or browsers and requires wrapping with `calc()`.

yisibl: Try~
image

chearon: It would be better not to, but it's probably not a big deal to parse the commas with the slash, since users will almost for sure be using one of the two syntaxes. Good to point it out though.

chearon (approved):

PR #2543, opened 2025-11-16
Tests occasionally timeout on macOS 26

cclauss:
Should we increase the `timeout` value or do something different?

>    291 passing (3s)
  6 pending
  1 failing

>  1) Canvas
       Canvas#getContext("2d"):
     Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/Users/runner/work/node-canvas/node-canvas/test/canvas.test.js)
      at process.processImmediate (node:internal/timers:491:21)

Also, add Node.js 25.2.0 and DRY the CI strategy so it can be configured once and reused.
* https://github.blog/changelog/2025-09-18-actions-yaml-anchors-and-non-public-workflow-templates

Thanks for contributing!

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -11,9 +11,10 @@ jobs:
   Linux:
     name: Test on Linux
     runs-on: ubuntu-latest
-    strategy:
+    strategy: &job_strategy  # Define a YAML anchor on first use
+      fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


chearon:
Thanks but I don't really mind the duplication in `ci.yaml`. I haven't seen anything time out on later versions of macOS either. If I'm wrong about that, we can reopen this or make a new PR, but I'm trying to cut down on the PR list.

Issue #2545, opened 2025-12-08
Canvas install failed on docker.


fahad-6sensehq:
In my NestJS app, I used canvas for exporting graphs from my system. Canvas work properly on my local machine (Windows). But when I try to dockerize it for AWS ECR, its failed. Facing with error form last week before that everything was fine. Is it node:22-apline issue or canvas issue?

How can I resolve this issue?

 > [deps 6/6] RUN pnpm install:
64.81 .../canvas@3.2.0/node_modules/canvas install: gyp ERR! stack Error: `make` failed with exit code: 2
64.81 .../canvas@3.2.0/node_modules/canvas install: gyp ERR! stack at ChildProcess.<anonymous> (/usr/local/lib/node_modules/pnpm/dist/node_modules/node-gyp/lib/build.js:216:23)
64.81 .../canvas@3.2.0/node_modules/canvas install: gyp ERR! System Linux 6.11.0-1018-azure
64.81 .../canvas@3.2.0/node_modules/canvas install: gyp ERR! command "/usr/local/bin/node" "/usr/local/lib/node_modules/pnpm/dist/node_modules/node-gyp/bin/node-gyp.js" "rebuild"
64.81 .../canvas@3.2.0/node_modules/canvas install: gyp ERR! cwd /app/node_modules/.pnpm/canvas@3.2.0/node_modules/canvas
64.81 .../canvas@3.2.0/node_modules/canvas install: gyp ERR! node -v v22.21.1
64.81 .../canvas@3.2.0/node_modules/canvas install: gyp ERR! node-gyp -v v10.3.1
64.81 .../canvas@3.2.0/node_modules/canvas install: gyp ERR! not ok 
64.81 .../canvas@3.2.0/node_modules/canvas install: Failed
64.83  ELIFECYCLE  Command failed with exit code 1.

GitHub Actions Error:

<img width="1275" height="338" alt="Image" src="https://github.com/user-attachments/assets/be5af6e5-f598-4786-aae2-c53c9b62f612" />

<img width="1306" height="743" alt="Image" src="https://github.com/user-attachments/assets/4ed313e3-45d5-4d2e-9a91-25fc7773c327" />

My docker file:

FROM node:22-alpine AS deps
WORKDIR /app

RUN apk add --no-cache \
    python3 \
    make \
    g++ \
    pkgconfig \
    cairo-dev \
    pango-dev \
    jpeg-dev \
    giflib-dev \
    fontconfig \
    ttf-dejavu

[...]

PR #2546, opened 2025-12-16
Fix compilation on GCC 15 by including <cstdint>

skooch:
As per the GCC 15 porting guide, section "Header dependency changes": https://gcc.gnu.org/gcc-15/porting_to.html#header-dep-changes

> Some C++ Standard Library headers have been changed to no longer include other headers that were being used internally by the library. As such, C++ programs that used standard library components without including the right headers will no longer compile.
>
> In particular, the following headers are used less widely within libstdc++ and may need to be included explicitly when compiling with GCC 15:
>
> `<stdint.h>` (for `int8_t`, `int32_t` etc.) and `<cstdint>` (for `std::int8_t`, `std::int32_t` etc.)
> `<ostream>` (for `std::endl`, `std::flush` etc.)

`<cstdint>` must now be explicitly included to use these types and was missing from `CharData.h`.

This resolves issue #2545 

Thanks for contributing!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 * Fix error message HTTP response status code in image src setter
 * `roundRect()` shape incorrect when radii were large relative to rectangle size (#2400)
 * Reject loadImage when src is null or invalid (#2304)
+* Fix compilation on GCC 15 by including <cstdint> (#2545)
 
 3.2.0
 ==================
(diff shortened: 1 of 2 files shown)


skooch:
@chearon No rush, but when would we expect a release to be cut after merging this? Just want to know when I should check back, since I have compiler flags littered in builds at work.

chearon:
I plan to be working on node-canvas next week, and I can do a patch release then. No release is quick to do until we get our prebuilds overhauled 😕

skooch:
@chearon Thank you!

PR #2556, opened 2026-03-13
fix: arm64 prebuild missing libexpat.so.1

martonperei:
On arm64, /etc/ld.so.conf.d/aarch64-linux-gnu.conf sorts before libc.conf, causing ldconfig to prefer /lib/aarch64-linux-gnu over /usr/local/lib. This means lddtree resolves libexpat.so.1 to the system copy in /lib/, which bundle.sh then filters out.

On x86, libc.conf (containing /usr/local/lib) sorts first by coincidence, so the locally-built libexpat is preferred and bundled.

Fix by adding 00-local.conf to ensure /usr/local/lib always takes priority regardless of architecture.


chearon:
Thanks but we've migrated to zig build instead of the prebuilds branch. The old build system was a huge pain. The latest binaries are built statically, no more `libexpat.so.1`, so it should work for you:

```sh
npm install canvas@next
```

PR #2562, opened 2026-04-06
Bump `node-addon-api` to fix memory leak #2436

### LINKED ISSUE #2436: Memory leak-like behavior caused by node-addon-api not doing GC synchronously
In the node-canvas@3.0.0-rc2 there is a big memory leak on the loadImage function.
The image loaded is never released.

reverting to 2.11.2 fixed my issue

iurisilvio:
## Fix memory leak from deferred N-API weak reference callbacks

Fixes #2436

### Problem

Since the NAN -> N-API migration (v3.0.0), `ObjectWrap` destructors (which free cairo surfaces) are deferred to the next `SetImmediate` instead of running during GC. In server environments where the event loop is always busy, these destructors never fire and memory grows indefinitely.

This was reported upstream in nodejs/node-addon-api#1140 and fixed in nodejs/node#42651 by introducing `node_api_nogc_finalize`, a finalizer that runs directly during GC for native-only cleanup.

### Fix

- Upgrade `node-addon-api` from 7.x to 8.x (which uses `node_api_nogc_finalize` for `ObjectWrap::FinalizeCallback`)
- Add `NAPI_EXPERIMENTAL` to `binding.gyp` defines (required to enable `NODE_API_EXPERIMENTAL_HAS_POST_FINALIZER` in [Node's headers](https://github.com/nodejs/node/blob/4f08c6478d6ecd073c03536b8a6c473232e16c37/src/js_native_api.h#L540))

No code changes, `node-addon-api` 8.x automatically routes `ObjectWrap` destructor callbacks through the nogc finalizer path when the flag is set.

I don't fully understand the impact of `node-addon-api` upgrade or enabling `NAPI_EXPERIMENTAL`.

### Before / After

200 requests decoding 3000x2000 JPEG images into thumbnails:

```javascript
// node --expose-gc stress.js
const { createCanvas, loadImage } = require("canvas");

let s = 1;
function rand() {
    s = (s + 0x6d2b79f5) | 0;
    let t = Math.imul(s ^ (s >>> 15), 1 | s);
    t = (t + Math.imul(t ^ (t >>> 7), 61 | t)) ^ t;
    return ((t ^ (t >>> 14)) >>> 0) / 4294967296;
}

function genImage(w, h) {
    const c = createCanvas(w, h);
    const ctx = c.getContext("2d");
    for (let i = 0; i < 10; i++) {
        ctx.fillStyle = `rgb(${(rand() * 255) | 0},${(rand() * 255) | 0},${(rand() * 255) | 0})`;
        ctx.fillRect((rand() * w) | 0, (rand() * h) | 0, (rand() * w) / 2, (rand() * h) / 2);
    }
    return c.toBuffer("image/jpeg");
}

async function run() {
    const imgs = [genImage(2000, 1500), genImage(3000, 2000)];
    let peak = 0;
    for (let i = 0; i < 50; i++) {
        const promises = [];
        for (let j = 0; j < 4; j++) {
            promises.push(
                loadImage(imgs[j % 2]).then((img) => {
                    const c = createCanvas(200, Math.round((img.height / img.width) * 200));
                    c.getContext("2d").drawImage(img, 0, 0, c.width, c.height);
                    return c.toBuffer("image/jpeg");
                })
            );
        }
        await Promise.all(promises);
        const rss = process.memoryUsage().rss;
        if (rss > peak) peak = rss;
        if ((i + 1) % 10 === 0)
            console.log(`  batch ${i + 1}: rss=${Math.round(rss / 1024 / 1024)}MB`);
    }
    console.log(`Peak RSS: ${Math.round(peak / 1024 / 1024)}MB`);
}

run();
```

```
// Before (node-addon-api 8, no NAPI_EXPERIMENTAL):
  batch 10: rss=784MB
  batch 20: rss=1477MB
  batch 30: rss=2169MB
  batch 40: rss=2862MB
  batch 50: rss=3555MB
Peak RSS: 3555MB

// After (node-addon-api 8 + NAPI_EXPERIMENTAL):
  batch 10: rss=196MB
  batch 20: rss=195MB
  batch 30: rss=196MB
  batch 40: rss=196MB
  batch 50: rss=196MB
Peak RSS: 196MB
```

### DIFF
--- a/src/Canvas.cc
+++ b/src/Canvas.cc
@@ -141,12 +141,6 @@ Canvas::~Canvas() {
   destroySurface();
 }
 
-void Canvas::Finalize(Napi::Env env) {
-  if (_surface && type == CANVAS_TYPE_IMAGE) {
-    Napi::MemoryManagement::AdjustExternalMemory(env, -(int64_t)approxBytesPerPixel() * width * height);
-  }
-}
-
 /*
  * Get type string.
  */
@@ -997,6 +991,9 @@ Canvas::ensureSurface() {
 void
 Canvas::destroySurface() {
   if (_surface) {
+    if (type == CANVAS_TYPE_IMAGE) {
+      Napi::MemoryManagement::AdjustExternalMemory(env, -(int64_t)approxBytesPerPixel() * width * height);
+    }
     // flush any operations that may use the closure that is freed below
     cairo_surface_finish(_surface);
     cairo_surface_destroy(_surface);
--- a/src/Canvas.cc
+++ b/src/Canvas.cc
@@ -141,6 +141,12 @@ Canvas::~Canvas() {
   destroySurface();
 }
 
+void Canvas::Finalize(Napi::Env env) {
+  if (_surface && type == CANVAS_TYPE_IMAGE) {
+    Napi::MemoryManagement::AdjustExternalMemory(env, -(int64_t)approxBytesPerPixel() * width * height);
+  }
+}
+
 /*
  * Get type string.
  */
@@ -993,9 +999,6 @@ Canvas::destroySurface() {
   if (_surface) {
     // flush any operations that may use the closure that is freed below
     cairo_surface_finish(_surface);
-    if (type == CANVAS_TYPE_IMAGE) {
-      Napi::MemoryManagement::AdjustExternalMemory(env, -(int64_t)approxBytesPerPixel() * width * height);
-    }
     cairo_surface_destroy(_surface);
     _surface = nullptr;
   }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,10 @@ project adheres to [Semantic Versioning](http://semver.org/).
 (Unreleased)
 ==================
 ### Changed
+* Upgrade node-addon-api from 7.x to 8.x
 ### Added
 ### Fixed
+* Fix memory leak caused by N-API weak reference callbacks being deferred to SetImmediate instead of running during GC. Enabled `NAPI_EXPERIMENTAL` to use `node_api_nogc_finalize` for ObjectWrap destructor. (#2436)
 
 3.2.3
 ==================
--- a/binding.gyp
+++ b/binding.gyp
@@ -58,7 +58,7 @@
     {
       'target_name': 'canvas',
       'include_dirs': ["<!(node -p \"require('node-addon-api').include_dir\")"],
-      'defines': [ 'NAPI_DISABLE_CPP_EXCEPTIONS', 'NODE_ADDON_API_ENABLE_MAYBE' ],
+      'defines': [ 'NAPI_DISABLE_CPP_EXCEPTIONS', 'NODE_ADDON_API_ENABLE_MAYBE', 'NAPI_EXPERIMENTAL' ],
       'sources': [
         'src/bmp/BMPParser.cc',
         'src/Canvas.cc',
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "util/"
   ],
   "dependencies": {
-    "node-addon-api": "^7.0.0",
+    "node-addon-api": "^8.7.0",
     "prebuild-install": "^7.1.3"
   },
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "prebenchmark": "node-gyp build",
     "benchmark": "node benchmarks/run.js",
     "lint": "standard examples/*.js test/server.js test/public/*.js benchmarks/run.js lib/context2d.js util/has_lib.js browser.js index.js",
-    "test": "mocha test/*.test.js",
+    "test": "mocha --node-option expose-gc test/*.test.js",
     "pretest-server": "node-gyp build",
     "test-server": "node test/server.js",
     "generate-wpt": "node ./test/wpt/generate.js",
--- a/src/Canvas.h
+++ b/src/Canvas.h
@@ -65,6 +65,7 @@ class Canvas : public Napi::ObjectWrap<Canvas> {
   public:
     Canvas(const Napi::CallbackInfo& info);
     ~Canvas();
+    void Finalize(Napi::Env env);
     static void Initialize(Napi::Env& env, Napi::Object& target);
 
     Napi::Value ToBuffer(const Napi::CallbackInfo& info);
--- a/src/Canvas.h
+++ b/src/Canvas.h
@@ -65,7 +65,6 @@ class Canvas : public Napi::ObjectWrap<Canvas> {
   public:
     Canvas(const Napi::CallbackInfo& info);
     ~Canvas();
-    void Finalize(Napi::Env env);
     static void Initialize(Napi::Env& env, Napi::Object& target);
 
     Napi::Value ToBuffer(const Napi::CallbackInfo& info);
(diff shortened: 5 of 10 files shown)


chearon:
Great find! That tracks perfectly with what we discussed in the issue, and it looks like we are good to upgrade since we don't support node 16.

The tests are failing, though.

iurisilvio:
Thank you for running the tests, I didn't run them local before! Now I ran it on Mac and Linux, reproduced the failures local, I hope my last commit fix all of them. 🙏🏻

iurisilvio:
Nice, all green now! 💚 Looking forward to have it reviewed/merged/released!

src/Canvas.cc
```diff
@@ -141,6 +141,12 @@ Canvas::~Canvas() {
   destroySurface();
 }
 
+void Canvas::Finalize(Napi::Env env) {
+  if (_surface && type == CANVAS_TYPE_IMAGE) {
+    Napi::MemoryManagement::AdjustExternalMemory(env, -(int64_t)approxBytesPerPixel() * width * height);
+  }
+}
+
```
chearon: From my reading of node-addon-api, having this member function forces the destructor to get called after gc, though before JS goes back to executing. It makes sense since this couldn't get called if the instance were already destroyed. Here's the relevant bit, cleaned up ([source](https://github.com/nodejs/node-addon-api/blob/0add1306f60b81432da94d13683aa0b06aa52925/napi-inl.h#L5331-L5365)):

```cpp
// If class overrides the basic finalizer, execute it.
if constexpr (details::HasBasicFinalizer<T>::value) {
  HandleScope scope(env);
  instance->Finalize(Napi::BasicEnv(env));
}

// If class overrides the (extended) finalizer, schedule it.
if constexpr (details::HasExtendedFinalizer<T>::value) {
  // Attach via node_api_post_finalizer. `PostFinalizeCallback` is responsible
  // for deleting the `T* instance`, after calling the user-provided finalizer.
  napi_status status =
      node_api_post_finalizer(env, PostFinalizeCallback, data, nullptr);
  NAPI_FATAL_IF_FAILED(status,
                       "ObjectWrap<T>::FinalizeCallback",
                       "node_api_post_finalizer failed");
}
// If the instance does _not_ override the (extended) finalizer, delete the
// `T* instance` immediately.
else {
  delete instance;
}

```

And the `PostFinalizeCallback` looks like this:

```cpp
T* instance = static_cast<T*>(data);
instance->Finalize(Napi::Env(env));
delete instance;
```

But it would be better to destroy _within_ gc, which is the whole point of `BasicEnv`. All day I've been trying to figure out why that causes node to crash (your initial test failures). I know it crashes in `AdjustExternalMemory`, but it seems to be failing because of something internal to node or node-addon-api.

`AdjustExternalMemory` takes a `BasicEnv`, and that env should still be valid, so I'm kind of at a loss for now. I'd like to take some time to understand it, which either leads to a bug fix in node or a corrected understanding on my part. I'm a little more hesitant to use the experimental API now. If we wanted to use it now, just having this method be empty and leaving the `AdjustExternalMemory` below would be better, since we want that to get called if we destroy from the C++ side.

I also think we should probably refactor the classes to only store `BasicEnv` to avoid accidentally trying to invoke JS at the wrong times. It's pretty easy to grab the full `Napi::Env` from `info` or other context.

iurisilvio:
I added a test to make sure memory leak doesn't come back later.

I don't know how to proceed from here.

flohall:
Works like a charm for our server-side rendering. Memory issues we had are gone. We already use a merged version internally at the company. Looking forward for the official merge.

iurisilvio:
I built binaries with this fix for me [here](https://github.com/iurisilvio/node-canvas/releases/tag/v3.2.3-memfix.0).

With all the recent supply chain attacks, you really should not trust me. 😆 The build is based on [this PR](https://github.com/iurisilvio/node-canvas/pull/1), you can easily fork it, inspect all the changes and build it for yourself using the github action workflow.

iurisilvio:
I deployed it today and looks like it didn't fix my memory leak. 😢 

Happy that it worked for others.

iurisilvio:
For completeness, it doesn't mean the fix is not real, others confirmed it works, but probably I have other leaks. 😆

flohall:
Sorry it did not fix your issues @iurisilvio

Still it fixed our memory issues. Here is the exact code version of canvas we used for testing in case someone is interested:
https://github.com/Automattic/node-canvas/commit/6a84247870b797635ef4172a7df4265f0ee6d0f9

iurisilvio:
No worries, I'm tracking other memory leaks in this lib and I think my leak is fixed after them.

* #2572 
* #2573 
* #2576 
* #2577 
* #2578 
* #2579 
* #2584 
* #2586

iurisilvio:
Memory super stable with all these pull requests applied. 👆🏻 I don't know which ones are more relevant.

src/Canvas.cc
chearon: I made progress figuring out most of the answers to my questions: https://github.com/nodejs/node-addon-api/issues/1724#issuecomment-4583914521

chearon:
As-is this PR is still indirectly moving the cleanup to a post finalizer (after gc, before js). Right now I think there's a bug in nodejs that doesn't let us do `napi_delete_reference` during gc (https://github.com/nodejs/node-addon-api/issues/1724#issuecomment-4583914521) so we could maybe use that as a stopgap, but all these things are behind `NAPI_EXPERIMENTAL`, including the bug, so I'm more inclined to wait until the API is mature.

Keep in mind the node-canvas bug is easily worked around by adding `await new Promise(resolve => setTimeout(resolve))` before your canvas is created. 

Also some of the other changes in this PR I find hard to justify as well - a `CanvasPattern` doesn't need to hold a persistent reference to the `Canvas`. The underlying Cairo pattern already has a reference-counted reference to the Cairo instance.

iurisilvio:
To be fair, I don't know if this change contributed to my memory leak fix. I applied all my open PRs and it is stable. I'll try to run without the NAPI_EXPERIMENTAL changes later.

flohall:
An update from my side. This PR did not help us notably in the end. The testing we did in the past was done wrong. We only noticed that there is no big effect after running the change on production for some while. Instead of trying out more potential fixes, we decided to run the code that references node-canvas in a subprocess. We never had such a clean memory - w are using node canvas for some years on production now. Obviously this comes at the price, that starting the subprocess takes 500-800 ms (on AWS Lambda - a faster CPU would delay less.)

Still stability does count more for us than the additional time. So we go with this.

chearon:
After 6400e7a17da48d6c1af8a141e48aa59f99ee0550 which I landed yesterday, the patch to fix the memory "leak" is simply:

```diff
diff --git a/binding.gyp b/binding.gyp
index 495f6e1..2e1e1c1 100644
--- a/binding.gyp
+++ b/binding.gyp
@@ -59,6 +59,7 @@
       'target_name': 'canvas',
       'include_dirs': ["<!(node -p \"require('node-addon-api').include_dir\")"],
       'defines': [
+        'NAPI_EXPERIMENTAL',
         'NAPI_DISABLE_CPP_EXCEPTIONS',
         'NODE_ADDON_API_ENABLE_MAYBE',
         'NODE_ADDON_API_REQUIRE_BASIC_FINALIZERS',
```

However, I don't know if experimental APIs can guarantee forward compatibility. We saw the crashes it caused, very likely due to a bug with the new API, and we haven't heard anything from the node-addon-api team. So the safest thing to do is wait for the experimental APIs to mature, and in the meantime, people can just pepper in `await new Promise(resolve => setTimeout(resolve))`. Keep following #2436 for updates.

iurisilvio:
I agree, it also didn't really fix the memory leak in 2 cases. It probably improves something, but I don't want to push for this merge. Thank you for working in the other issues, 🙏🏻 happy to test it soon!

PR #2582, opened 2026-05-18
Take ownership of cairo_pattern_t in canvas_state_t with refcounts

### LINKED ISSUE #2578: canvas_state_t leaks cairo_pattern_t members; fillStyle/strokeStyle are weak refs
(This issue makes sense only after https://github.com/Automattic/node-canvas/pull/2562)

## Bug
`canvas_state_t` ([CanvasRenderingContext2d.h:19-72](https://github.com/Automattic/node-canvas/blob/master/src/CanvasRenderingContext2d.h#L19-L72)) holds four `cairo_pattern_t*` members (`fillPattern`, `strokePattern`, `fillGradient`, `strokeGradient`) but the destructor only frees `fontDescription`. Every time a state is destroyed with any of those four pointers non-null, the cairo pattern leaks.
Additionally, the copy constructor copies the raw pointers without `cairo_pattern_reference()`. Multiple states (from `save()`/`restore()` chains) end up sharing the same `cairo_pattern_t*` with no shared ownership — so freeing in any state's destructor would dangle the others.
A related correctness issue: `Context2d::_fillStyle` / `Context2d::_strokeStyle` are weak references (`.Reset(obj)`, refcount 0). Under prompt finalization (e.g. when `NAPI_EXPERIMENTAL` is enabled and `node_api_post_finalizer` runs the finalizer at a safe point), V8 can garbage-collect the JS `CanvasGradient` / `CanvasPattern` object while the Context2d's state still references its native `cairo_pattern_t`. Subsequent reads of `ctx.fillStyle` return a stale handle.
## Reproduction
```js
'use strict';
if (typeof gc !== 'function') {
  console.error('Run with --expose-gc');
  process.exit(1);
}
const canvas = require('canvas');
const ITER = parseInt(process.argv[2] || '20000', 10);
const MODE = process.argv[3] || 'gradient';   // 'gradient' or 'color' (control)
const mb = (n) => +(n / 1024 / 1024).toFixed(1);
(async () => {
  gc(); gc();
  const baseline = mb(process.memoryUsage().rss);
  let last = baseline;
  console.log(`mode=${MODE} iter=${ITER} baseline rss=${baseline} MiB`);
  for (let i = 1; i <= ITER; i++) {
    const c = canvas.createCanvas(200, 200);
    const ctx = c.getContext('2d');
    if (MODE === 'gradient') {
      const grad = ctx.createLinearGradient(0, 0, 200, 200);
      grad.addColorStop(0, 'red');
      grad.addColorStop(1, 'blue');
      ctx.fillStyle = grad;
    } else {
      ctx.fillStyle = 'red';
    }
    ctx.fillRect(0, 0, 200, 200);
    c.width = 0; c.height = 0;
    if (i % 2000 === 0) {
      gc(); gc();
      const rss = mb(process.memoryUsage().rss);
      console.log(`iter=${i} rss=${rss}M slope=${(((rss - last) * 1024) / 2000).toFixed(2)} KiB/iter`);
      last = rss;
    }
  }
})();
```
## Observed on `canvas@3.2.3` from npm (Linux arm64, Node 20)
```
mode=gradient iter=10000 baseline rss=46 MiB
iter=2000  rss=51.7M  slope=2.92 KiB/iter
iter=4000  rss=60.5M  slope=4.51 KiB/iter
iter=6000  rss=70.4M  slope=5.07 KiB/iter
iter=8000  rss=80.0M  slope=4.92 KiB/iter
iter=10000 rss=89.6M  slope=4.92 KiB/iter
per-iter=4.97 KiB
```
Slope is consistent ~5 KiB/iter — each iteration leaks roughly one `cairo_pattern_t` worth (a few hundred bytes for solid gradients plus cairo's internal metadata). Slope on `mode=color` is ~0 since the four pattern fields stay null.
## Fix
Take refcounted ownership in `canvas_state_t`:
1. Setters refcount the new pattern (`cairo_pattern_reference`) and release the previous one (`cairo_pattern_destroy`).
2. Copy constructor and `operator=` reference each pattern instead of copying raw pointers.
3. Destructor releases all four patterns.
4. New helpers `setFillPattern` / `setStrokePattern` / `setFillGradient` / `setStrokeGradient` / `clearFillPattern` / `clearStrokePattern` enforce the discipline at the call sites in `Context2d::SetFillStyle` / `Context2d::SetStrokeStyle` / `_setFillColor` / `_setStrokeColor`.
For the style retention bug, change `_fillStyle.Reset(obj)` and `_strokeStyle.Reset(obj)` to `Reset(obj, 1)` so the JS gradient/pattern object stays alive as long as the Context2d references its native `cairo_pattern_t`.
The same change should also clear the state stack in `~Context2d` before destroying `_layout`/`_context` so state destructors run with a valid context, and replace `resetState`'s single `states.pop()` with a `while (!states.empty()) states.pop()` followed by `states.emplace()` + `state = &states.top()` (the original `pop()` underflows if `states` is empty and leaves `state` dangling).
## Affects
Any application that uses gradients or patterns in a canvas pipeline (data viz, charting, complex image processing). The bug is in core `Context2d` state ownership.

iurisilvio:
## Take refcounted ownership of `cairo_pattern_t` in `canvas_state_t`

`canvas_state_t` holds four `cairo_pattern_t*` members (`fillPattern`, `strokePattern`, `fillGradient`, `strokeGradient`), but the destructor only freed `fontDescription` and the copy constructor copied the raw pointers. So `save()`/`restore()` chains shared the same `cairo_pattern_t*` across states with no shared ownership, and every state destroyed with a non-null pattern leaked it.

This PR takes proper refcounted ownership:

- the copy ctor and a new `operator=` `cairo_pattern_reference` each pattern;
- `~canvas_state_t` `cairo_pattern_destroy`s all four;
- `set*` / `clear*` helpers handle replacement at the call sites (`SetFillStyle`, `SetStrokeStyle`, `_setFillColor`, `_setStrokeColor`).

The state now holds its own reference, so a gradient/pattern stays valid for the state's lifetime regardless of when the JS wrapper is GC'd.

Fixes #2578.

---

**Scope:** this PR is now just the leak fix. The `fillStyle`/`strokeStyle` strong-ref and the `Context2d` state-lifetime correctness changes that were originally bundled here have moved to #2598 — they're independent of this ownership change (and not tied to the experimental finalizer path that the original description mentioned), so they're easier to review on their own.

### DIFF
--- a/src/CanvasRenderingContext2d.cc
+++ b/src/CanvasRenderingContext2d.cc
@@ -90,7 +90,6 @@ inline static bool checkArgs(const Napi::CallbackInfo&info, double *args, int ar
 
 void
 Context2d::Initialize(Napi::Env& env, Napi::Object& exports) {
-  Napi::HandleScope scope(env);
   InstanceData* data = env.GetInstanceData<InstanceData>();
 
   Napi::Function ctor = DefineClass(env, "CanvasRenderingContext2D", {
@@ -1968,11 +1967,11 @@ Context2d::SetFillStyle(const Napi::CallbackInfo& info, const Napi::Value& value
     if (obj.InstanceOf(data->CanvasGradientCtor.Value()).UnwrapOr(false)) {
       _fillStyle.Reset(obj);
       Gradient *grad = Gradient::Unwrap(obj);
-      state->fillGradient = grad->pattern();
+      state->setFillGradient(grad->pattern());
     } else if (obj.InstanceOf(data->CanvasPatternCtor.Value()).UnwrapOr(false)) {
       _fillStyle.Reset(obj);
       Pattern *pattern = Pattern::Unwrap(obj);
-      state->fillPattern = pattern->pattern();
+      state->setFillPattern(pattern->pattern());
     }
   }
 }
@@ -2008,11 +2007,11 @@ Context2d::SetStrokeStyle(const Napi::CallbackInfo& info, const Napi::Value& val
     if (obj.InstanceOf(data->CanvasGradientCtor.Value()).UnwrapOr(false)) {
       _strokeStyle.Reset(obj);
       Gradient *grad = Gradient::Unwrap(obj);
-      state->strokeGradient = grad->pattern();
+      state->setStrokeGradient(grad->pattern());
     } else if (obj.InstanceOf(data->CanvasPatternCtor.Value()).UnwrapOr(false)) {
       _strokeStyle.Reset(value);
       Pattern *pattern = Pattern::Unwrap(obj);
-      state->strokePattern = pattern->pattern();
+      state->setStrokePattern(pattern->pattern());
     }
   }
 }
@@ -2193,7 +2192,7 @@ Context2d::_setFillColor(Napi::Value arg) {
     if (status != napi_ok) return;
     uint32_t rgba = rgba_from_string(buf, &ok);
     if (!ok) return;
-    state->fillPattern = state->fillGradient = NULL;
+    state->clearFillPattern();
     state->fill = rgba_create(rgba);
   }
 }
@@ -2219,7 +2218,7 @@ Context2d::_setStrokeColor(Napi::Value arg) {
   std::string str = arg.As<Napi::String>();
   uint32_t rgba = rgba_from_string(str.c_str(), &ok);
   if (!ok) return;
-  state->strokePattern = state->strokeGradient = NULL;
+  state->clearStrokePattern();
   state->stroke = rgba_create(rgba);
 }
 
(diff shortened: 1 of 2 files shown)


iurisilvio:
Closing: real ownership gap by inspection (canvas_state_t didn't free its cairo_pattern_t members), but I couldn't reproduce a measurable leak — it sits below RSS noise on the deferred (node-addon-api 7) build; the numbers in #2578 were measured on the experimental nogc build. Scoping my fork to leaks reproduced with before/after RSS (EXIF #2574, SVG #2585).

PR #2585, opened 2026-05-18
Release RsvgHandle and partial surface on SVG decode error paths

### LINKED ISSUE #2584: loadSVGFromBuffer leaks RsvgHandle when SVG has no intrinsic size
## Bug

`Image::loadSVGFromBuffer()` calls `rsvg_handle_new_from_data()` to build `_rsvg`, then calls `rsvg_handle_get_intrinsic_size_in_pixels()` to read `d_width` / `d_height`. If the SVG parses as valid XML but has no intrinsic size (no `width`/`height` attribute on `<svg>` and no element that implies one), `rsvg_handle_get_intrinsic_size_in_pixels()` returns `FALSE` and leaves the out-params untouched. The current code:

```cpp
double d_width;
double d_height;

rsvg_handle_get_intrinsic_size_in_pixels(_rsvg, &d_width, &d_height);

width  = naturalWidth  = d_width;
height = naturalHeight = d_height;

if (width <= 0 || height <= 0) {
  this->errorInfo.set("Width and height must be set on the svg element");
  return CAIRO_STATUS_READ_ERROR;
}
```

- `d_width` / `d_height` are read uninitialized when `rsvg_handle_get_intrinsic_size_in_pixels` returns `FALSE`.
- The error return leaves `_rsvg` attached to the `Image`. The `RsvgHandle` (and the parsed SVG DOM it owns) is only released when the JS `Image` wrapper is eventually finalized.

A related path in `Image::renderSVGToSurface()` has the same shape: some error branches `g_object_unref(_rsvg)` but don't clear `_rsvg`, and don't destroy `_surface` / the cairo context that have already been created. `Image::surface()` then runs a redundant `g_object_unref(_rsvg)` if `renderSVGToSurface` already unreffed it on its error path — which is what masks the leak from showing as a crash in callers that retry.

## Reproduction

```js
'use strict';
if (typeof gc !== 'function') {
  console.error('Run with --expose-gc');
  process.exit(1);
}
const canvas = require('canvas');
const ITER = parseInt(process.argv[2] || '10000', 10);
const mb = (n) => +(n / 1024 / 1024).toFixed(1);

// Valid XML, parses cleanly. No width/height, no element that implies a size.
const svg = Buffer.from(
  '<svg xmlns="http://www.w3.org/2000/svg"><path d="M1,1"/></svg>'
);

(async () => {
  gc(); gc();
  const baseline = mb(process.memoryUsage().rss);
  let last = baseline;
  console.log(`baseline rss=${baseline} MiB`);
  for (let i = 1; i <= ITER; i++) {
    try { await canvas.loadImage(svg); } catch (_) { /* expected */ }
    if (i % 2000 === 0) {
      gc(); gc();
      const rss = mb(process.memoryUsage().rss);
      console.log(`iter=${i} rss=${rss}M slope=${(((rss - last) * 1024) / 2000).toFixed(3)} KiB/iter`);
      last = rss;
    }
  }
  gc(); gc();
  const end = mb(process.memoryUsage().rss);
  console.log(`baseline=${baseline}M end=${end}M delta=${(end - baseline).toFixed(1)}M per-iter=${(((end - baseline) * 1024) / ITER).toFixed(3)} KiB`);
})();
```

## Observed on `canvas@3.2.3` from npm (Linux x86_64 in node:20-bookworm)

```
baseline rss=45.9 MiB
iter=2000  rss=65.7M  slope=10.138 KiB/iter
iter=4000  rss=73.4M  slope=3.942  KiB/iter
iter=6000  rss=80.7M  slope=3.738  KiB/iter
iter=8000  rss=87.7M  slope=3.584  KiB/iter
iter=10000 rss=94.7M  slope=3.584  KiB/iter
baseline=45.9M end=94.7M delta=48.8M per-iter=4.997 KiB
```

Steady ~4 KiB/iter slope — one orphaned `RsvgHandle` (with its parsed DOM) per failed `loadImage`. The slope is also visible without `--expose-gc`, just with more noise from delayed finalization.

## Cause

1. `Image::loadSVGFromBuffer()` doesn't check the return value of `rsvg_handle_get_intrinsic_size_in_pixels()` and uses uninitialized `d_width` / `d_height` when it returns `FALSE`.
2. On the missing-dimensions error return, `_rsvg` is not released and `_is_svg` is left set, so the orphaned `RsvgHandle` lives until the `Image` wrapper is collected.
3. `Image::renderSVGToSurface()` has the inverse problem on its own error branches: `_rsvg` is unreffed but not nulled, and `_surface` / the cairo context aren't torn down, so a later cleanup pass (or a redundant `g_object_unref` in the `surface()` caller) can either leak or double-unref depending on which branch was hit.

## Fix

In `loadSVGFromBuffer`:

- initialize `d_width = 0` / `d_height = 0`
- use the return value of `rsvg_handle_get_intrinsic_size_in_pixels()` to decide whether to trust the values
- on the missing-dimensions error path: `g_object_unref(_rsvg)`, set `_rsvg = NULL`, and clear `_is_svg` before returning the error

In `renderSVGToSurface`:

- on each error branch, destroy the cairo context if created, destroy `_surface` and null it, unref and null `_rsvg`, and clear `_is_svg`
- callers (`Image::surface()`) should not unref `_rsvg` again after `renderSVGToSurface()` has already cleaned up on its error path

After applying both fixes, the same reproduction runs flat (slope is in the same range as N-API finalizer / cairo allocator noise).

iurisilvio:
Fixes #2584.

`Image::loadSVGFromBuffer()` ignored the return value of `rsvg_handle_get_intrinsic_size_in_pixels()`, so when the SVG parsed cleanly but had no intrinsic size, `d_width`/`d_height` were read uninitialized and the function returned `CAIRO_STATUS_READ_ERROR` with `_rsvg` still attached and `_is_svg` still set. The orphaned `RsvgHandle` survived until the JS `Image` wrapper was finalized — visible as ~4 KiB/iter of steady RSS growth across repeated `loadImage()` of size-less SVG input.

`Image::renderSVGToSurface()` had the inverse problem on its own error branches: `_rsvg` was unreffed but not nulled, and `_surface` (and the cairo context on later branches) were left live. `Image::surface()` then ran a redundant `g_object_unref(_rsvg)` on the error path, double-touching state that may or may not have already been released.

## Changes

- `loadSVGFromBuffer`: initialize `d_width`/`d_height`, gate the assignment on the `gboolean` return value of `rsvg_handle_get_intrinsic_size_in_pixels()`, and unref/null `_rsvg` plus clear `_is_svg` before returning the missing-dimensions error.
- `renderSVGToSurface`: on each error branch, destroy the cairo context if created, destroy and null `_surface`, unref and null `_rsvg`, clear `_is_svg`.
- `Image::surface`: drop the now-redundant `g_object_unref(_rsvg)` on the `renderSVGToSurface` failure path — the inner function already cleaned up.

## Verification

Repro from #2584, `canvas@3.2.3` on Linux x86_64 / Node 20:

```
baseline rss=45.9 MiB
iter=10000 rss=94.7M slope=3.584 KiB/iter
baseline=45.9M end=94.7M delta=48.8M per-iter=4.997 KiB
```

After this patch the same loop is flat (slope in the range of N-API finalizer / cairo allocator noise).

### DIFF
--- a/src/Image.cc
+++ b/src/Image.cc
@@ -405,7 +405,6 @@ cairo_surface_t *Image::surface() {
 
     cairo_status_t status = renderSVGToSurface();
     if (status != CAIRO_STATUS_SUCCESS) {
-      g_object_unref(_rsvg);
       Napi::Error::New(env, cairo_status_to_string(status)).ThrowAsJavaScriptException();
 
       return NULL;
@@ -1473,16 +1472,19 @@ Image::loadSVGFromBuffer(uint8_t *buf, unsigned len) {
     return CAIRO_STATUS_READ_ERROR;
   }
 
-  double d_width;
-  double d_height;
+  double d_width = 0;
+  double d_height = 0;
 
-  rsvg_handle_get_intrinsic_size_in_pixels(_rsvg, &d_width, &d_height);
+  gboolean has_size = rsvg_handle_get_intrinsic_size_in_pixels(_rsvg, &d_width, &d_height);
 
-  width = naturalWidth = d_width;
-  height = naturalHeight = d_height;
+  width = naturalWidth = has_size ? d_width : 0;
+  height = naturalHeight = has_size ? d_height : 0;
 
   if (width <= 0 || height <= 0) {
     this->errorInfo.set("Width and height must be set on the svg element");
+    g_object_unref(_rsvg);
+    _rsvg = NULL;
+    _is_svg = false;
     return CAIRO_STATUS_READ_ERROR;
   }
 
@@ -1501,6 +1503,10 @@ Image::renderSVGToSurface() {
   status = cairo_surface_status(_surface);
   if (status != CAIRO_STATUS_SUCCESS) {
     g_object_unref(_rsvg);
+    _rsvg = NULL;
+    cairo_surface_destroy(_surface);
+    _surface = NULL;
+    _is_svg = false;
     return status;
   }
 
@@ -1508,6 +1514,11 @@ Image::renderSVGToSurface() {
   status = cairo_status(cr);
   if (status != CAIRO_STATUS_SUCCESS) {
     g_object_unref(_rsvg);
+    _rsvg = NULL;
+    cairo_destroy(cr);
+    cairo_surface_destroy(_surface);
+    _surface = NULL;
+    _is_svg = false;
     return status;
   }
 
@@ -1520,7 +1531,11 @@ Image::renderSVGToSurface() {
   gboolean render_ok = rsvg_handle_render_document(_rsvg, cr, &viewport, nullptr);
   if (!render_ok) {
     g_object_unref(_rsvg);
+    _rsvg = NULL;
     cairo_destroy(cr);
+    cairo_surface_destroy(_surface);
+    _surface = NULL;
+    _is_svg = false;
     return CAIRO_STATUS_READ_ERROR; // or WRITE?
   }
 


chearon:
RSVG is now gone, but I wouldn't turn down any improvements to the new SVG implementation. node-canvas now uses [LunaSVG](https://github.com/sammycage/lunasvg).

iurisilvio:
Looks like lunasvg has similar issues, I'll open another PR soon.

iurisilvio:
I opened https://github.com/Automattic/node-canvas/pull/2613 and https://github.com/chearon/lunasvg/pull/1 to handle some SVG issues, both with repro.

PR #2601, opened 2026-07-13
Support Node.js object URLs as image sources

### LINKED ISSUE #2525: `Image#src=` does not work with object URL (`onload` never called, cannot be drawn on 2d context)
```js
import { Image } from 'canvas';
import { readFileSync } from 'fs';

const file = readFileSync('test.png');
const objectUrl = URL.createObjectURL(new Blob([file], { type: 'image/png' }));
const image = new Image();
image.onload = () => {
	URL.revokeObjectURL(objectUrl);
	console.log('in onload: complete =', image.complete); // never called
};
image.src = objectUrl;
console.log('immediately after setting src: complete =', image.complete); // true
```

If you try to draw an image with src being an object URL on a 2d context, you get `Error: Image given has not completed loading`.

benitoalba:
## Summary

- resolve Node.js `blob:nodedata:` URLs through `resolveObjectURL`
- pass the resolved Blob bytes through the existing Buffer image loader and preserve the original source URL
- test successful loading and drawing, plus rejection after URL revocation
- document the fix in the changelog

## Why

`Image#src` only recognized data and HTTP(S) URLs. Node.js object URLs therefore fell through to local path handling and were passed to `fopen`, so `onload` never fired and the image could not be drawn.

Fixes #2525.

## Testing

- `npx mocha test/image.test.js`: 37 passing, 3 pending
- `npm run lint`
- `npm run tsd`
- `npm test`: 301 passing, 7 pending; two pre-existing `CanvasPattern`/`CanvasGradient` save/restore failures on Windows also reproduce on unmodified `master`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 ### Changed
 ### Added
 ### Fixed
+* Load images from Node.js object URLs (#2525)
 
 3.2.3
 ==================
--- a/lib/image.js
+++ b/lib/image.js
@@ -13,6 +13,7 @@
 const bindings = require('./bindings')
 const Image = module.exports = bindings.Image
 const util = require('util')
+const { resolveObjectURL } = require('node:buffer')
 
 const { GetSource, SetSource } = bindings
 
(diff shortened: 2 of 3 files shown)


chearon (approved):
Looks great, thank you!

Issue #2595, opened 2026-06-01
npm RFC #868 will block this package's install script by default


JamieMagee:
Hi there,

I'm reaching out because canvas is one of the most depended-on npm packages that runs an install script, and an accepted npm RFC is going to change how those scripts behave.

The RFC blocks dependency install scripts by default during `npm install`. That's the same thing pnpm, Yarn Berry, Bun, and Deno already do. Users opt back in per package through a new `allowScripts` field in `package.json` (or the `npm approve-scripts` command). The motivation is the run of supply-chain attacks over the last couple of years that used `postinstall` hooks to run code the moment a package landed in the tree.

For canvas, the script that's affected is:

    "install": "prebuild-install -r napi || node-gyp rebuild"

That script compiles a native addon, so under the new default it won't run unless the user adds `canvas` to their `allowScripts`. The change that helps your users most is shipping prebuilt binaries as `optionalDependencies`, the pattern esbuild, sharp, and @swc/core moved to. Most installs then need no compile step and no allowlist entry. If prebuilds aren't practical for canvas, a short note in your README showing the `allowScripts` line people should add will save a lot of confused issues.

I wanted to flag this early so you've got time to plan rather than find out when the install warnings start showing up. The full RFC is here: https://github.com/npm/rfcs/pull/868. Happy to answer any questions.

chearon:
Fixed via #2602

PR #2602, opened 2026-07-14
Version 4

### LINKED ISSUE #1394: make module context-aware

chearon:
Once and for all, this fixes the top issues users have experienced, which fall under these categories:

1.  Font issues. Pango doesn't support registering fonts from buffers or indeed, registering fonts at all. Our font selection has always been a hack: we have to guess what family name and attributes _should_ select a font that a user registered. Emojis didn't work unless everything is compiled right, and our current binaries are not.
    
    I wrote our own font stack so we can have full control over font lookup and the registration API. Fonts are the whole reason I got involved in node-canvas over [10 years ago](https://github.com/Automattic/node-canvas/pull/715), and I've used most of those 10 years to learn how to write a font layout engine! Instead of Pango, we use [zg](https://codeberg.org/atman/zg) and [SheenBidi](https://github.com/Tehreer/SheenBidi) to segment the text into shaping boundaries, then [HarfBuzz](https://github.com/harfbuzz/harfbuzz) in between shaping boundaries to get glyphs for Cairo. This a relatively easy layout engine - we don't have to do line wrapping in canvas! Getting fonts efficiently from the OS is harder, but Firefox's source was a great guide.
2. Native builds. Modern node addons are distributed as a wide array of binaries; even users on more rare platforms are not expected to compile. Our existing solution of building for the host machine doesn't support this. We need to be able to cross-compile. Zig is an excellent fit, and I found it easier to use than CMake or Meson. This PR adds 6 more platforms and it'll be easy to add more. Bonus: size is halved in the _worst_ case: some tarballs go from 20MiB to 3MiB.
3. Native _distribution_. Post-install scripts are being phased out: #2595. NPM packages are the modern way to distribute binaries because it's safer and more reliable. We know the user can access NPM servers, but not necessarily GitHub releases, since corporate internet and proxies can interfere.

Why couldn't I do this in small pieces? (2) and (3) are tied together because it wouldn't make sense to update the old build system. (1) is tied to those because updating the old build before replacing the new (low dependency) font stack would be tons of wasted work. Adding the font stack first would've meant we couldn't use zg. Unicode libraries are hard to package, bug Zig makes it easy. So the PR removes fonts, adds static build, then adds fonts back. Diff accordingly.

As a bonus I fixed #1394 since it became easy on this branch (last commit).

I don't expect anyone to seriously review all of this code, but review at any level is welcome, as well as testing. I'll release some betas off of this PR soon. @LinusU and @zbjornson - it's been a while, but I would love to hear from you!

### The Ugly

It isn't possible to keep using librsvg and have a static build, it is just too huge of a library and I don't have the time for it. [lunasvg](https://github.com/sammycage/lunasvg) is able to support a surprising number of SVG 1.1 files while still being nimble. Sadly, no `<filter>` or SVG2 support, and that is a breaking change.

---

There are too many issues to list here. I'll hunt them down and close them after this gets merged.


lvcabral:
I'm happy to test the Beta for #1394 fix

chearon:
Yes that would be much appreciated! So close to getting a beta out, but one of the NPM packages triggered fraud detection. Waiting on NPM support, then I'll post back here.

chearon:
4.0.0-rc1 released!

```
npm install canvas@next
```

lvcabral:
@chearon in my preliminary tests, noticed that using the new fonts system, trying to render empty string is crashing. How/where do you want me to report issues on 4.0-rc1 ?

chearon:
You can just open issues and put 4.0.0-rc1 as the version number. Thanks for the testing.

lvcabral:
> You can just open issues and put 4.0.0-rc1 as the version number. Thanks for the testing.

Created a couple of issues related to font rendering

lvcabral:
@chearon I can't thank you enough for this release, you unlocked something for my project that I was waiting for years! 

Now I combined `node-canvas` with `terminal-image` in my BrightScript Simulation Engine and I can run a Roku app fully on the terminal! 

https://github.com/user-attachments/assets/824d610e-cfa6-4775-b95e-20980e52e33c


https://github.com/user-attachments/assets/8233ddcb-ebe7-4a11-abe4-9d98bc53ec11

src/Canvas.cc
```diff
@@ -390,10 +373,6 @@ Canvas::ToBuffer(const Napi::CallbackInfo& info) {
   if (info[0].StrictEquals(Napi::String::New(env, "raw"))) {
     cairo_surface_t *surface = ensureSurface();
     cairo_surface_flush(surface);
-    if (nBytes() > node::Buffer::kMaxLength) {
-      Napi::Error::New(env, "Data exceeds maximum buffer length.").ThrowAsJavaScriptException();
-      return env.Undefined();
-    }
```
zbjornson, line 409: Is this condition no longer possible?

src/closure.h
```diff
@@ -7,20 +7,16 @@ class Canvas;
...
 
 #include <napi.h>
 #include <png.h>
 #include <stdint.h> // node < 7 uses libstdc++ on macOS which lacks complete c++11
 #include <vector>
 
-#ifndef PAGE_SIZE
-  #define PAGE_SIZE 4096
-#endif
+#define PAGE_SIZE 4096
```
zbjornson: Might want to rename this so it doesn't look confusing/sloppy, and make it a constexpr.

zbjornson (commented):
Really awesome!

The old, migrated code looks good. I don't know enough about fonts to help review the new font code.

src/Canvas.cc
chearon: Good eye!

chearon: Actually, I think this was intentional... node headers aren't used anymore, and this was wrong if loaded by another JS engine. Untested, but the buffer ctor should throw an exception if you try to create one beyond the limit, so it wasn't ever necessary.

src/FontLayout.h
chearon, line 1: fillText, measureText, and strokeText are significantly slower right now (although the order of magnitude is in the microseconds). They should be, and were at one point, significantly faster. I did a little debugging and narrowed it down to font selection. This is highly cacheable and just needs to be reworked a bit.

chearon:
Published `4.0.0-rc2`

src/Canvas.cc
zbjornson: You're right, Node does check that.

chearon:
`4.0.0-rc3` released with the performance fix. `ctx.font = ` is now twice as fast as `master` and fill/stroke/measureText are about 3x as fast with `FontLayout.cc` replacing Pango 🎉 

Thanks for the help @lvcabral and @zbjornson, I'm going to merge this and add `npm install canvas@next` to the README. That way I can start closing issues and hopefully we can attract some more testers.

PR #2612, opened 2026-08-20
Faster fill/strokeText

chearon:
I found a performance regression in v4 with `fillText` once I looked at real-world numbers. Added a benchmark to `master` that shows it. `FT_Face` stores caching info and `cairo_set_font_size` can destroy caching info as well.


948764321:
Is it related to https://github.com/Automattic/node-canvas/issues/2615?

chearon:
Possibly, but it sounds different. I just commented on that issue.

PR #2613, opened 2026-08-20
Fix segfault, use-after-free and per-parse leak in the SVG paths

iurisilvio:
Three failures in the lunasvg SVG paths, each with a test that fails on master and passes here. Reproduced first against the published `canvas@4.0.0-rc2`, then against a source build of master.

## The failures

**An SVG with an embedded image in a format we cannot decode segfaults.** The decoder callback in `loadSVGFromBuffer()` ignored the status of `loadFromBuffer()` and returned whatever surface the failed decode left behind, which for an unsupported format is null. lunasvg passed that straight to `cairo_surface_set_user_data()`. Loading a document with `<image href="data:image/webp;base64,...">` — or with corrupt image data — takes the process down.

**A failed SVG re-render leaves a dangling surface.** `renderSVGToSurface()` destroyed `_surface` without nulling it before its error returns, so the next render and `clearData()` both destroy it again:

```js
const img = new Image()
img.src = Buffer.from('<svg xmlns="http://www.w3.org/2000/svg" width="16" height="16"><rect width="16" height="16"/></svg>')
ctx.drawImage(img, 0, 0)
img.width = img.height = 40000   // > the largest image surface cairo will make
ctx.drawImage(img, 0, 0)         // throws, as it should
ctx.drawImage(img, 0, 0)         // aborts
```

```
Assertion failed: (CAIRO_REFERENCE_COUNT_HAS_REFERENCE (&surface->ref_count)),
function cairo_surface_destroy, file cairo-surface.c, line 963
```

Only an assertion-enabled cairo turns it into an abort, so it shows up on the debug build `zig build` produces and is a silent use-after-free in a `ReleaseFast` one.

**Every parse of an SVG with an embedded JPEG, GIF or BMP leaks one decoded frame.** Those decoders back their surface with a `new uint8_t[]` buffer through `cairo_image_surface_create_for_data()`, which cairo does not own. `transferSurface()` hands the surface to lunasvg and drops `_data` without freeing it or attaching a destroy callback, so nothing frees the pixels — ever. An embedded PNG is unaffected, since cairo owns the pixels on that path, which makes it a clean control:

```
1200x1200 raster, RSS slope over 100 parses after a 50-parse warmup
master     embedded PNG      1 KiB/parse     embedded JPEG  6238 KiB/parse
this PR    embedded PNG      2 KiB/parse     embedded JPEG     5 KiB/parse
                                             (one decoded frame is 5625 KiB)
```

## Changes

- `transferSurface()` attaches `_data` to the surface with a destroy callback, so the buffer dies with the surface that points at it. If cairo cannot store it, the surface goes too rather than leaving a buffer nothing can free.
- `renderSVGToSurface()` nulls `_surface` after destroying it, and treats a null `Bitmap` — what `renderToBitmap()` returns for a document with no intrinsic size, or when the allocation fails — as an error instead of reading a status through a null pointer.
- `Context2d::DrawImage()` stops when `surface()` returns null. It always could, once a re-render fails; nothing checked because the abort came first.
- The decoder callback returns null on a failed decode instead of a half-built surface.
- `loadSVGFromBuffer()` releases the parsed document on both of its error paths, where it is only reachable again through an `Image` that never completed loading.
- `pkg/lunasvg` picks up chearon/lunasvg#1, which null-checks the decoder result. **This currently points at my fork so the tests pass here — happy to re-point it at chearon/lunasvg once that lands, or drop the bump and let it come in separately.**

## Tests

`test/svg.test.js`. Both crashes take the process down, so each case runs in a child process and the test reports the signal. The leak test measures an RSS slope and calibrates it against the same image embedded as a PNG in the same process, rather than against a fixed threshold — debug and release builds differ by two orders of magnitude in allocator noise.

On master: 3 failing. On this branch: the full suite passes, 308 tests, on both `zig build` and `zig build -Doptimize=ReleaseFast`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 ### Fixed
 * Load images from Node.js object URLs (#2525)
+* Fix crash loading an SVG with an embedded image in an undecodable format (#2613)
+* Fix use-after-free when re-rendering an SVG image at a size cairo rejects (#2613)
+* Fix leak of one decoded frame per parse of an SVG with an embedded JPEG, GIF or BMP (#2613)
 
 3.2.3
 ==================
--- a/pkg/lunasvg/build.zig.zon
+++ b/pkg/lunasvg/build.zig.zon
@@ -5,8 +5,8 @@
     .dependencies = .{
         .cairo = .{ .path = "../cairo" },
         .lunasvg = .{
-            .url = "https://github.com/iurisilvio/lunasvg/archive/c8fbe875e4c241a8b45c8e274647297db12bda22.tar.gz",
-            .hash = "N-V-__8AAPD5FADyrh8KcLt8S2MOxVcQuxCvX9DkrI5zKsUZ",
+            .url = "https://github.com/iurisilvio/lunasvg/archive/b03b0dd7ba9967bd507c4a13ab12ff37e1d4128a.tar.gz",
+            .hash = "N-V-__8AAGcLFQB2sVL8Tvu9IeCPuaczmK9v1-aj16-Uasup",
         }
     },
 }
--- a/pkg/lunasvg/build.zig.zon
+++ b/pkg/lunasvg/build.zig.zon
@@ -5,8 +5,8 @@
     .dependencies = .{
         .cairo = .{ .path = "../cairo" },
         .lunasvg = .{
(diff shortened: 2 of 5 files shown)


chearon:
Thanks for finding these. Do you mind if I use this like a GitHub issue and implement the fixes myself? I have several thoughts/comments on how to fix, but I don't want to go back and forth with an AI (your other PR I will merge after the one change though).

iurisilvio:
No problem, you can do it yourself, I'm happy just providing the failure cases! Thank you!

